PR #56, opened 2018-11-13
CI Concurrent Jobs

KtorZ:
This should speed up CI execution time even further. The downside: we might need multiple caches, one for each job type (caches depends on available ENV vars). 

We'll see how it goes.


Anviking (approved):
lgtm

PR #57, opened 2018-11-13
BIP-44 A. Review & Extend Existing API & Handlers

denisshevchenko:
Review & Extend Existing API & Handlers (3 points)

1. The API must support 'Create' operation on E.O.S Wallets.
2. Address Pool Gap may be configurable per E.O.S Wallets at the API-level.

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -39,20 +39,18 @@ module Cardano.Wallet.API.V1.Types (
   , WalletType (..)
   , WalletOperation (..)
   , SpendingPassword
-  , ExternalWallet (..)
+  , EosWallet (..)
+  , EosWalletId (..)
   , PublicKeyAsBase58
   , mkPublicKeyAsBase58
   , mkPublicKeyFromBase58
   , Base58PublicKeyError (..)
-  , NewExternalWallet (..)
+  , NewEosWallet (..)
   , WalletAndTxHistory (..)
   -- * Addresses
   , AddressOwnership (..)
   , AddressIndex
   , AddressValidity (..)
-  , AddressAsBase58
-  , mkAddressAsBase58
-  , mkAddressFromBase58
   -- * Accounts
   , Account (..)
   , accountsHaveSameId
@@ -150,9 +148,12 @@ import           Data.ByteString.Base58 (bitcoinAlphabet, decodeBase58,
                      encodeBase58)
 import qualified Data.Char as C
 import           Data.Default (Default (def))
+import           Data.List ((!!))
 import           Data.Semigroup (Semigroup)
 import           Data.Swagger hiding (Example, example)
 import           Data.Text (Text, dropEnd, toLower)
+import           Data.UUID (UUID)
+import qualified Data.UUID as Uuid
 import           Formatting (bprint, build, fconst, int, sformat, stext, (%))
 import qualified Formatting.Buildable
 import           Generics.SOP.TH (deriveGeneric)
@@ -508,7 +509,7 @@ instance BuildableSafeGen NewWallet where
 -- 1. As a root PK to identify external wallet.
 -- 2. As a derived PK, please see 'AddressWithProof' type.
 newtype PublicKeyAsBase58 = PublicKeyAsBase58Unsafe
-    { ewalPublicKeyAsBase58 :: Text
+    { eoswalPublicKeyAsBase58 :: Text
     } deriving (Eq, Generic, Ord, Show)
 
 deriveJSON Aeson.defaultOptions ''PublicKeyAsBase58
@@ -518,7 +519,7 @@ instance Arbitrary PublicKeyAsBase58 where
 
 instance ToSchema PublicKeyAsBase58 where
     declareNamedSchema =
-        genericSchemaDroppingPrefix "ewal" (\(--^) props -> props
+        genericSchemaDroppingPrefix "eoswal" (\(--^) props -> props
             & ("publicKeyAsBase58" --^ "Extended public key in Base58-format.")
         )
 
@@ -527,7 +528,7 @@ instance BuildableSafeGen PublicKeyAsBase58 where
     buildSafeGen sl PublicKeyAsBase58Unsafe{..} = bprint ("{"
         %" publicKeyAsBase58="%buildSafe sl
         %" }")
-        ewalPublicKeyAsBase58
+        eoswalPublicKeyAsBase58
 
 instance FromHttpApiData PublicKeyAsBase58 where
     parseQueryParam someText =
@@ -537,6 +538,10 @@ instance FromHttpApiData PublicKeyAsBase58 where
 
 instance ToHttpApiData PublicKeyAsBase58 where
     toQueryParam (PublicKeyAsBase58Unsafe pk) = pk
+
+-- | To be able to work with the list of encoded public keys.
+instance Buildable [PublicKeyAsBase58] where
+    build = bprint listJson
 
 -- | Smart constructor for 'PublicKeyAsBase58'.
 mkPublicKeyAsBase58 :: PublicKey -> PublicKeyAsBase58
@@ -565,43 +570,11 @@ mkPublicKeyFromBase58 (PublicKeyAsBase58Unsafe encodedXPub) = do
             Left problem -> Left $ NotAPublicKey (toText problem)
             Right xPub   -> Right $ PublicKey xPub
 
--- | Type for representation address in Base58-format.
--- We use it for external wallets.
-newtype AddressAsBase58 = AddressAsBase58Unsafe
-    { ewalAddressAsBase58 :: Text
-    } deriving (Eq, Generic, Ord, Show)
-
-deriveJSON Aeson.defaultOptions ''AddressAsBase58
-instance Arbitrary AddressAsBase58 where
-    arbitrary = AddressAsBase58Unsafe <$> pure
-        "DdzFFzCqrhsxqFTw2ENzvwisYmS2DcUTujXDoMXtMrCvMFAa3DikLj8YYTWNZaEjthKZpMNWKo9RUoq3gP797yP8MP4g9qiEegvGEY9w"
-
-instance ToSchema AddressAsBase58 where
-    declareNamedSchema =
-        genericSchemaDroppingPrefix "ewal" (\(--^) props -> props
-            & ("addressAsBase58" --^ "Address in Base58-format.")
-        )
-
-deriveSafeBuildable ''AddressAsBase58
-instance BuildableSafeGen AddressAsBase58 where
-    buildSafeGen sl AddressAsBase58Unsafe{..} = bprint ("{"
-        %" addressAsBase58="%buildSafe sl
-        %" }")
-        ewalAddressAsBase58
-
--- | Smart constructor for 'AddressAsBase58'.
-mkAddressAsBase58 :: Core.Address -> AddressAsBase58
-mkAddressAsBase58 = AddressAsBase58Unsafe . decodeUtf8 . Core.addrToBase58
-
--- | Decode Base58-address to real Address.
-mkAddressFromBase58 :: AddressAsBase58 -> Either Text Core.Address
-mkAddressFromBase58 (AddressAsBase58Unsafe txtAddr) = Core.decodeTextAddress txtAddr
-
 -- | Type for representation of serialized transaction in Base16-format.
 -- We use it for external wallets (to send/receive raw transaction during
 -- external signing).
 newtype TransactionAsBase16 = TransactionAsBase16Unsafe
-    { ewalTransactionAsBase16 :: Text
+    { eoswalTransactionAsBase16 :: Text
     } deriving (Eq, Generic, Ord, Show)
 
 deriveJSON Aeson.defaultOptions ''TransactionAsBase16
@@ -611,7 +584,7 @@ instance Arbitrary TransactionAsBase16 where
 
 instance ToSchema TransactionAsBase16 where
     declareNamedSchema =
-        genericSchemaDroppingPrefix "ewal" (\(--^) props -> props
+        genericSchemaDroppingPrefix "eoswal" (\(--^) props -> props
             & ("transactionAsBase16" --^ "Serialized transaction in Base16-format.")
         )
 
@@ -620,7 +593,7 @@ instance BuildableSafeGen TransactionAsBase16 where
     buildSafeGen sl TransactionAsBase16Unsafe{..} = bprint ("{"
         %" transactionAsBase16="%buildSafe sl
         %" }")
-        ewalTransactionAsBase16
+        eoswalTransactionAsBase16
 
 -- | Type for representation of transaction signature in Base16-format.
 -- We use it for external wallet. Please note that technically there's no
@@ -652,38 +625,46 @@ mkTransactionSignatureAsBase16 :: Core.Signature Txp.TxSigData -> TransactionSig
 mkTransactionSignatureAsBase16 (Core.Signature txSig) =
     TransactionSignatureAsBase16Unsafe . Base16.encode . CC.unXSignature $ txSig
 
--- | A type modelling the request for a new 'ExternalWallet',
+instance Buildable [PublicKey] where
+    build = bprint listJson
+
+-- | A type modelling the request for a new 'EosWallet',
 -- on the mobile client or hardware wallet.
-data NewExternalWallet = NewExternalWallet
-    { newewalRootPK         :: !PublicKeyAsBase58
-    , newewalAssuranceLevel :: !AssuranceLevel
-    , newewalName           :: !WalletName
+data NewEosWallet = NewEosWallet
+    { neweoswalAccountsPublicKeys :: ![PublicKey]
+    , neweoswalAddressPoolGap     :: !(Maybe Word)
+    , neweoswalAssuranceLevel     :: !AssuranceLevel
+    , neweoswalName               :: !WalletName
     } deriving (Eq, Show, Generic)
 
-deriveJSON Aeson.defaultOptions ''NewExternalWallet
-instance Arbitrary NewExternalWallet where
-    arbitrary = NewExternalWallet <$> arbitrary
-                                  <*> arbitrary
-                                  <*> pure "My external Wallet"
+deriveJSON Aeson.defaultOptions ''NewEosWallet
+instance Arbitrary NewEosWallet where
+    arbitrary = NewEosWallet <$> arbitrary
+                             <*> arbitrary
+                             <*> arbitrary
+                             <*> pure "My EOS-wallet"
 
-instance ToSchema NewExternalWallet where
+instance ToSchema NewEosWallet where
     declareNamedSchema =
-        genericSchemaDroppingPrefix "newewal" (\(--^) props -> props
-            & ("rootPK"         --^ "Root public key to identify external wallet.")
-            & ("assuranceLevel" --^ "Desired assurance level based on the number of confirmations counter of each transaction.")
-            & ("name"           --^ "External wallet's name.")
+        genericSchemaDroppingPrefix "neweoswal" (\(--^) props -> props
+            & ("accountsPublicKeys" --^ "External wallet's accounts public keys.")
+            & ("addressPoolGap"     --^ "Address pool gap for this wallet.")
+            & ("assuranceLevel"     --^ "Desired assurance level based on the number of confirmations counter of each transaction.")
+            & ("name"               --^ "External wallet's name.")
         )
 
-deriveSafeBuildable ''NewExternalWallet
-instance BuildableSafeGen NewExternalWallet where
-    buildSafeGen sl NewExternalWallet{..} = bprint ("{"
-        %" rootPK="%buildSafe sl
+deriveSafeBuildable ''NewEosWallet
+instance BuildableSafeGen NewEosWallet where
+    buildSafeGen sl NewEosWallet{..} = bprint ("{"
+        %" accountsPublicKeys="%build
+        %" addressPoolGap="%build
         %" assuranceLevel="%buildSafe sl
         %" name="%buildSafe sl
         %" }")
-        newewalRootPK
-        newewalAssuranceLevel
-        newewalName
+        neweoswalAccountsPublicKeys
+        neweoswalAddressPoolGap
+        neweoswalAssuranceLevel
+        neweoswalName
 
 -- | A type modelling the update of an existing wallet.
 data WalletUpdate = WalletUpdate {
@@ -997,40 +978,77 @@ instance BuildableSafeGen Wallet where
 instance Buildable [Wallet] where
     build = bprint listJson
 
--- | An external wallet (mobile client or hardware wallet).
-data ExternalWallet = ExternalWallet
-    { ewalId             :: !WalletId
-    , ewalName           :: !WalletName
-    , ewalBalance        :: !(V1 Core.Coin)
-    , ewalAssuranceLevel :: !AssuranceLevel
+instance ToSchema PublicKey where
+    declareNamedSchema _ =
+        pure $ NamedSchema (Just "PublicKey") $ mempty
+            & type_ .~ SwaggerString
+            & format ?~ "base58"
+
+newtype EosWalletId = EosWalletId UUID
+    deriving (Show, Eq, Ord, Generic)
+
+deriveJSON Aeson.defaultOptions ''EosWalletId
+
+instance ToSchema EosWalletId where
+  declareNamedSchema = genericDeclareNamedSchema defaultSchemaOptions
+
+instance ToJSONKey EosWalletId
+
+instance Arbitrary EosWalletId where
+    arbitrary = EosWalletId <$> Gen.chooseAny
+
+deriveSafeBuildable ''EosWalletId
+instance BuildableSafeGen EosWalletId where
+    buildSafeGen sl (EosWalletId uuid) =
+        bprint (plainOrSecureF sl build (fconst "<eos wallet id>")) $ Uuid.toText uuid
+
+instance FromHttpApiData EosWalletId where
+    parseQueryParam rawUuid = case Uuid.fromText rawUuid of
+        Nothing   -> Left "Invalid EOS-wallet id (not a UUID)."
+        Just uuid -> Right . EosWalletId $ uuid
+
+instance ToHttpApiData EosWalletId where
+    toQueryParam (EosWalletId uuid) = Uuid.toText uuid
+
+-- | Externally-owned sequential (EOS) wallet (mobile client or hardware wallet).
+data EosWallet = EosWallet
+    { eoswalId             :: !EosWalletId
+    , eoswalName           :: !WalletName
+    , eoswalAddressPoolGap :: !Word
+    , eoswalBalance        :: !(V1 Core.Coin)
+    , eoswalAssuranceLevel :: !AssuranceLevel
     } deriving (Eq, Ord, Show, Generic)
 
-deriveJSON Aeson.defaultOptions ''ExternalWallet
-instance ToSchema ExternalWallet where
+deriveJSON Aeson.defaultOptions ''EosWallet
+instance ToSchema EosWallet where
     declareNamedSchema =
-        genericSchemaDroppingPrefix "ewal" (\(--^) props -> props
-            & ("id"             --^ "Unique wallet identifier.")
-            & ("name"           --^ "Wallet's name.")
-            & ("balance"        --^ "Current balance, in Lovelaces.")
-            & ("assuranceLevel" --^ "The assurance level of the wallet.")
+        genericSchemaDroppingPrefix "eoswal" (\(--^) props -> props
+            & ("id"                 --^ "Unique wallet's identifier.")
+            & ("name"               --^ "Wallet's name.")
+            & ("addressPoolGap"     --^ "Address pool gap for this wallet.")
+            & ("balance"            --^ "Current balance, in Lovelaces.")
+            & ("assuranceLevel"     --^ "The assurance level of the wallet.")
         )
 
-instance Arbitrary ExternalWallet where
-    arbitrary = ExternalWallet <$> arbitrary
-                               <*> pure "My external wallet"
-                               <*> arbitrary
-                               <*> arbitrary
+instance Arbitrary EosWallet where
+    arbitrary = EosWallet <$> arbitrary
+                          <*> pure "My EOS-wallet"
+                          <*> arbitrary
+                          <*> arbitrary
+                          <*> arbitrary
 
-deriveSafeBuildable ''ExternalWallet
-instance BuildableSafeGen ExternalWallet where
-  buildSafeGen sl ExternalWallet{..} = bprint ("{"
+deriveSafeBuildable ''EosWallet
+instance BuildableSafeGen EosWallet where
+  buildSafeGen sl EosWallet{..} = bprint ("{"
     %" id="%buildSafe sl
     %" name="%buildSafe sl
+    %" addressPoolGap="%build
     %" balance="%buildSafe sl
     %" }")
-    ewalId
-    ewalName
-    ewalBalance
+    eoswalId
+    eoswalName
+    eoswalAddressPoolGap
+    eoswalBalance
 
 --------------------------------------------------------------------------------
 -- Addresses
@@ -1930,7 +1948,7 @@ instance Buildable [AddressLevel] where
 
 -- | Source address and corresponding derivation path, for external wallet.
 data AddressAndPath = AddressAndPath
-    { aapSrcAddress     :: !AddressAsBase58
+    { aapSrcAddress     :: !(V1 Core.Address) -- AsBase58
     -- ^ Source address in Base58-format.
     , aapDerivationPath :: ![AddressLevel]
     -- ^ Derivation path used during generation of this address.
@@ -1995,8 +2013,8 @@ instance BuildableSafeGen UnsignedTransaction where
 -- it returns the source address, the signature and the derived PK of
 -- the transaction input.
 data AddressWithProof = AddressWithProof
-    { awpSrcAddress  :: !AddressAsBase58
-    -- ^ Base58-encoded source address.
+    { awpSrcAddress  :: !(V1 Core.Address)
+    -- ^ Source address.
     , awpTxSignature :: !TransactionSignatureAsBase16
     -- ^ Base16-encoded signature of transaction (made by derived SK).
     , awpDerivedPK   :: !PublicKeyAsBase58
@@ -2259,8 +2277,8 @@ type family New (original :: *) :: * where
         NewAccount
     New WalletAddress =
         NewAddress
-    New ExternalWallet =
-        NewExternalWallet
+    New EosWallet =
+        NewEosWallet
 
 type CaptureWalletId = Capture "walletId" WalletId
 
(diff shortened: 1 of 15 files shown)


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -997,10 +1006,17 @@ instance BuildableSafeGen Wallet where
 instance Buildable [Wallet] where
     build = bprint listJson
 
+instance ToSchema PublicKey where
+    declareNamedSchema _ =
+        pure $ NamedSchema (Just "PublicKey") $ mempty
+            & type_ .~ SwaggerString
```
KtorZ, line 984: NOTE:  Swagger also supports a `format` type which is handy to specify encoding formats like `base58` :+1:

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -543,33 +552,33 @@ createHdWallet newRoot defaultHdAccountId defaultHdAddress utxoByAccount =
...
-                       -- ^ The default HdAccountId to go with this HdRoot. This
-                       -- function will take responsibility of creating the associated
-                       -- 'HdAccount'.
-                       -> Map HdAccountId (Utxo, [AddrWithId])
-                       -> Update DB (Either HD.CreateHdRootError ())
-createHdExternalWallet newRoot defaultHdAccountId utxoByAccount =
-    runUpdateDiscardSnapshot . zoom dbHdWallets $ do
-      HD.createHdRoot newRoot
-      updateAccounts_ $ map mkUpdateCreateHdWallet (Map.toList (insertDefault utxoByAccount))
+createHdExternalWallet :: EosHdRoot
```
KtorZ: Should we also rename to "createEosHdWallet" ?

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -543,33 +552,33 @@ createHdWallet newRoot defaultHdAccountId defaultHdAddress utxoByAccount =
...
-        , accountUpdateNew   = AccountUpdateNewUpToDate utxo
-        , accountUpdateAddrs = addrs
-        , accountUpdate      = return () -- just need to create it, no more
+    mkAccountIxs :: [Word32]
+    mkAccountIxs =
+        if numOfAccounts == 1
+            then [firstHardened]
+            else [firstHardened, firstHardened + 1 .. firstHardened + (numOfAccounts - 1)]
+      where
+        numOfAccounts = fromIntegral $ length accountsPKs
```
KtorZ: This isn't true. In the case of EOS wallets, we just _don't know_ the account indexes. And in theory, we don't need to know them since we won't be deriving new accounts.

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -649,6 +658,24 @@ data AccountUpdate e a = AccountUpdate {
...
 
+-- | All the information we need to update an account in EOS-wallet
+--
+-- See 'updateEosAccount' or 'updateEosAccounts'.
+data EosAccountUpdate e a = EosAccountUpdate {
+      -- | Account to update
+      eosAccountUpdateId    :: !HdAccountId
+
+      -- | Information needed when we need to create the account from scratch
+    , eosAccountUpdateNew   :: !AccountUpdateNew
```
KtorZ: "when we need to create the account from scratch"

I suppose this won't happen with EOS accounts ?

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -696,6 +723,24 @@ accountUpdateCreate accId (AccountUpdateNewIncomplete curUtxo genUtxo ctx) =
         , _hdIncompleteHistorical = Checkpoints $ one $ initCheckpoint        genUtxo
         }
 
+-- | Brand new account in EOS-wallet (if one needs to be created)
+eosAccountUpdateCreate :: HdAccountId -> AccountUpdateNew -> PublicKey -> EosHdAccount
```
KtorZ: So yeah, EOS wallets' accounts are really _just_ public keys. There's no index or id whatsoever.

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -721,6 +766,20 @@ updateAccounts = fmap Map.fromList . mapM updateAccount
...
+updateEosAccount :: EosAccountUpdate e a -> Update' e EosHdWallets (HdAccountId, a)
+updateEosAccount EosAccountUpdate{..} = do
+    res <- zoomOrCreateEosHdAccount
+             assumeEosHdRootExists
+             (eosAccountUpdateCreate eosAccountUpdateId
+                                     eosAccountUpdateNew
+                                     eosAccountUpdatePK)
+             eosAccountUpdateId
+             ((eosAccountUpdateId, ) <$> eosAccountUpdate)
+    return res
```
KtorZ: hlint not happy here :upside_down_face: 

```hs 
do
  a <- io 
  return a

-- vs

do
  io

```

src/Cardano/Wallet/Kernel/DB/EosHdWallet.hs
```diff
@@ -0,0 +1,250 @@
...
+data EosHdRoot = EosHdRoot {
+      -- | Wallet ID
+      _eosHdRootId             :: !HdRootId
+      -- | Wallet name
+    , _eosHdRootName           :: !WalletName
+      -- | Assurance level
+    , _eosHdRootAssurance      :: !AssuranceLevel
+      -- | Address pool gap for this wallet
+    , _eosHdRootAddressPoolGap :: !Word
+    } deriving (Eq, Show)
```
KtorZ, line 85: :+1:

src/Cardano/Wallet/Kernel/DB/EosHdWallet.hs
```diff
@@ -0,0 +1,250 @@
...
+instance Arbitrary EosHdRoot where
+    arbitrary = EosHdRoot <$> arbitrary
+                          <*> arbitrary
+                          <*> arbitrary
+                          <*> pure (20 :: Word)
+
+-- | Account in externally-owned sequential HD wallet
+data EosHdAccount = EosHdAccount {
+      -- | Account index
+      _eosHdAccountId            :: !HdAccountId
```
KtorZ: To be removed, cf comments above.

src/Cardano/Wallet/Kernel/DB/EosHdWallet.hs
```diff
@@ -0,0 +1,250 @@
...
+      -- | Account state
+      --
+      -- When the account is up to date with the blockchain, the account state
+      -- coincides with the state of a " wallet " as mandated by the formal
+      -- spec.
+    , _eosHdAccountState         :: !HdAccountState
+
+      -- | A local counter used to generate new 'AutoIncrementKey' for
+      -- addresses.
+    , _eosHdAccountAutoPkCounter :: !AutoIncrementKey
```
KtorZ: Not sure what this is :thinking: ?

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -230,37 +233,37 @@ createHdWallet pw mnemonic spendingPassword assuranceLevel walletName = do
 
 -- | Creates a new external HD 'Wallet'.
 createExternalHdWallet :: PassiveWallet
-                       -> PublicKey
-                       -- ^ Extended public key that corresponds to the root secret key of
-                       -- this external wallet (assumed that root secret key is storing
-                       -- externally, for example in the memory of Ledger device).
+                       -> [PublicKey]
```
KtorZ: Same as previous comment. we may want to rename this function as well :+1:

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -230,37 +233,37 @@ createHdWallet pw mnemonic spendingPassword assuranceLevel walletName = do
...
+                       -> IO (Either CreateExternalWalletError EosHdRoot)
+createExternalHdWallet pw accountsPKs addressPoolGap assuranceLevel walletName = do
+    -- Here, we review the definition of a wallet down to a list of account public keys with
+    -- no relationship whatsoever from the wallet's point of view. New addresses can be derived
+    -- for each account at will and discovered using the address pool discovery algorithm
+    -- described in BIP-44. Public keys are managed and provided from an external sources.
+    let nm = makeNetworkMagic (pw ^. walletProtocolMagic)
+    -- By default we need a root key for creating 'HdRoot'. But we don't have a root key
+    -- for externally-owned sequential wallet, so just take the first account's public
+    -- key as a base for wallet's 'HdRoot'.
```
KtorZ: Hopla! This should come from the API or be derived from it. A `RootId` is merely an `Address` derived from the wallet root public key. I think that's a fair thing to ask from the API (as it was already the case I believe?)

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -161,27 +163,47 @@ createExternalWallet :: MonadIO m
...
-                                      (HD.WalletName name)
-    return (mkExternalWallet name assuranceLevel root)
+    if null encodedAccountsPKs
+        then throwError CreateExternalWalletNoAccountsPKs
+        else do
+            let eitherPKs = map V1.mkPublicKeyFromBase58 encodedAccountsPKs
+                problems  = lefts eitherPKs
+                validPKs  = rights eitherPKs
+            if not . null $ problems
+                then throwError $ CreateExternalWalletInvalidAccountPK $ problems !! 0
```
KtorZ: Let's not throw here but leverage the `ExceptT` monad and  return a "pure" failure.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -161,27 +163,47 @@ createExternalWallet :: MonadIO m
...
+                     -> EosHD.EosHdRoot
+                     -> V1.ExternalWallet
+    mkExternalWallet v1WalletName accountsPKs addressPoolGap v1AssuranceLevel eosHdRoot =
+        V1.ExternalWallet { ewalId             = walletId
+                          , ewalName           = v1WalletName
+                          , ewalAccountsPKs    = accountsPKs
+                          , ewalAddressPoolGap = addressPoolGap
+                          , ewalBalance        = V1 (mkCoin 0)
+                          , ewalAssuranceLevel = v1AssuranceLevel
+                          }
```
KtorZ: Just  a minor styling advice: prefer the example below than the above. Variable-length indentation is a pain and creates unnecessary conflicts with git; try to avoid it at most. I know it's in a lot of place in the current codebase, but let's not introduce new ones :) 

```hs
mkExternalWallet 
  :: Text
  -> [PublicKey]
  -> Word
  -> V1.AssuranceLevel
  -> EosHD.EosHdRoot
  -> V1.ExternalWallet
mkExternalWallet v1WalletName accountsPKs addressPoolGap v1AssuranceLevel eosHdRoot =
    V1.ExternalWallet 
      { ewalId             = walletId
      , ewalName           = v1WalletName
      , ewalAccountsPKs    = accountsPKs
      , ewalAddressPoolGap = addressPoolGap
      , ewalBalance        = V1 (mkCoin 0)
      , ewalAssuranceLevel = v1AssuranceLevel
      }
```

KtorZ (changes requested):

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -518,7 +519,7 @@ instance Arbitrary PublicKeyAsBase58 where
 
 instance ToSchema PublicKeyAsBase58 where
     declareNamedSchema =
-        genericSchemaDroppingPrefix "ewal" (\(--^) props -> props
+        genericSchemaDroppingPrefix "eoswal" (\(--^) props -> props
             & ("publicKeyAsBase58" --^ "Extended public key in Base58-format.")
         )
```
KtorZ, line 524: I believe this schema is wrong. This is parsed and serialized a JSON `String` and not an object, right ?

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -578,7 +583,7 @@ instance Arbitrary AddressAsBase58 where
 
 instance ToSchema AddressAsBase58 where
     declareNamedSchema =
-        genericSchemaDroppingPrefix "ewal" (\(--^) props -> props
+        genericSchemaDroppingPrefix "eoswal" (\(--^) props -> props
             & ("addressAsBase58" --^ "Address in Base58-format.")
         )
```
KtorZ: Same as above.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -568,7 +573,7 @@ mkPublicKeyFromBase58 (PublicKeyAsBase58Unsafe encodedXPub) = do
 -- | Type for representation address in Base58-format.
 -- We use it for external wallets.
 newtype AddressAsBase58 = AddressAsBase58Unsafe
-    { ewalAddressAsBase58 :: Text
+    { eoswalAddressAsBase58 :: Text
     } deriving (Eq, Generic, Ord, Show)
```
KtorZ: Wondering: what's the need for a `newtype` here? Current `Address` are already serialized from and to Base58 JSON Strings :thinking:

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -652,38 +657,43 @@ mkTransactionSignatureAsBase16 :: Core.Signature Txp.TxSigData -> TransactionSig
...
--- | A type modelling the request for a new 'ExternalWallet',
+-- | A type modelling the request for a new 'EosWallet',
 -- on the mobile client or hardware wallet.
-data NewExternalWallet = NewExternalWallet
-    { newewalRootPK         :: !PublicKeyAsBase58
-    , newewalAssuranceLevel :: !AssuranceLevel
-    , newewalName           :: !WalletName
+data NewEosWallet = NewEosWallet
+    { neweoswalAccountsPKs    :: ![PublicKeyAsBase58]
+    , neweoswalAddressPoolGap :: !(Maybe Word)
```
KtorZ: NOTE for later: We probably want something much smaller than `word` here.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -652,38 +657,43 @@ mkTransactionSignatureAsBase16 :: Core.Signature Txp.TxSigData -> TransactionSig
...
                                   <*> arbitrary
                                   <*> pure "My external Wallet"
 
-instance ToSchema NewExternalWallet where
+instance ToSchema NewEosWallet where
     declareNamedSchema =
-        genericSchemaDroppingPrefix "newewal" (\(--^) props -> props
-            & ("rootPK"         --^ "Root public key to identify external wallet.")
+        genericSchemaDroppingPrefix "neweoswal" (\(--^) props -> props
+            & ("accountsPKs"    --^ "External wallet's accounts public keys.")
```
KtorZ: let's go for `publicKeys`, don't you think? I think it's clearer at the api-level, or at the very least: `accountPublicKeys` to avoid acronyms.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -997,40 +1007,81 @@ instance BuildableSafeGen Wallet where
...
+    declareNamedSchema _ =
+        pure $ NamedSchema (Just "PublicKey") $ mempty
+            & type_ .~ SwaggerString
+            & format ?~ "base58"
+
+newtype EosWalletId = EosWalletId Text
+    deriving (Show, Eq, Ord, Generic)
+
+exampleEosWalletId :: EosWalletId
+exampleEosWalletId = EosWalletId "c2cc10e1-57d6-4b6f-9899-38d972112d8c"
```
KtorZ: Let's use the `Example` class for that :upside_down_face:

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -649,6 +646,21 @@ data AccountUpdate e a = AccountUpdate {
...
+      -- | Account's public key (we obtained it from the user during wallet
+      -- creation).
+      eosAccountUpdatePK     :: !PublicKey
+
+      -- Root id of EOS-wallet this EOS-account belongs to.
+    , eosAccountUpdateRootId :: !EosHdRootId
+
+      -- | The update to run
+    , eosAccountUpdate       :: !(Update' e EosHdAccount a)
+    }
```
KtorZ, line 662: Out-of-scope for this PR / story. But, okay (don't remove it know. We'll refine if needed when time comes) :sweat_smile:

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -721,6 +733,14 @@ updateAccounts = fmap Map.fromList . mapM updateAccount
 updateAccounts_ :: [AccountUpdate e ()] -> Update' e HdWallets ()
 updateAccounts_ = mapM_ updateAccount
 
+updateEosAccount :: EosAccountUpdate e a -> Update' e EosHdWallets (PublicKey, a)
+updateEosAccount EosAccountUpdate{..} =
+    zoomOrCreateEosHdAccount
+        assumeEosHdRootExists
+        (EosHD.initEosHdAccount eosAccountUpdatePK eosAccountUpdateRootId)
+        eosAccountUpdatePK
+        ((eosAccountUpdatePK, ) <$> eosAccountUpdate)
```
KtorZ, line 742: Out-of-scope for this PR / story. But, okay (don't remove it know. We'll refine if needed when time comes) :sweat_smile:

src/Cardano/Wallet/Kernel/DB/EosHdWallet.hs
```diff
@@ -0,0 +1,217 @@
...
+-- | Currently, during EOS-wallet creation we do not receive
+-- root public key of this wallet, but only the list of accounts'
+-- public keys. So we cannot create 'HdRootId'. Since we need some
+-- unique identifier for EOS-wallet, we use UUID.
+-- Please note that UUID-based solution may be changed in the future.
+newtype EosHdRootId = EosHdRootId { getEosHdRootId :: UUID }
+    deriving (Eq, Ord, Show)
+
+instance Arbitrary EosHdRootId where
+    arbitrary = EosHdRootId <$> pure (fromJust $ Uuid.fromText "550e8400-e29b-41d4-a716-446655440000")
```
KtorZ: `UUID` implements the `Random` type-class, we can derive `arbitrary` from it using [`chooseAny`](https://hackage.haskell.org/package/QuickCheck-2.12.6.1/docs/Test-QuickCheck-Gen.html#v:chooseAny)

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -997,40 +1007,81 @@ instance BuildableSafeGen Wallet where
...
-    , ewalBalance        :: !(V1 Core.Coin)
-    , ewalAssuranceLevel :: !AssuranceLevel
+instance ToSchema PublicKey where
+    declareNamedSchema _ =
+        pure $ NamedSchema (Just "PublicKey") $ mempty
+            & type_ .~ SwaggerString
+            & format ?~ "base58"
+
+newtype EosWalletId = EosWalletId Text
+    deriving (Show, Eq, Ord, Generic)
```
KtorZ, line 988: Why isn't this a `UUID` instead of `Text` :thinking:  ?

src/Cardano/Wallet/Kernel/DB/EosHdWallet/Create.hs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+-- | New account in externally-owned sequential (EOS) wallet.
+initEosHdAccount :: PublicKey
+                 -> EosHdRootId
+                 -> EosHdAccount
+initEosHdAccount accPK eosRootId = EosHdAccount {
+      _eosHdAccountPK     = accPK
+    , _eosHdAccountRootId = eosRootId
+    }
```
KtorZ: Not sure to get the use for those functions :thinking: ? They are just `EosHdRoot` and `EosHdAccount` ...

src/Cardano/Wallet/Kernel/DB/EosHdWallet.hs
```diff
@@ -0,0 +1,217 @@
...
+import           Cardano.Wallet.Kernel.DB.Util.IxSet
+import qualified Cardano.Wallet.Kernel.DB.Util.IxSet as IxSet hiding (Indexable)
+
+-- | Currently, during EOS-wallet creation we do not receive
+-- root public key of this wallet, but only the list of accounts'
+-- public keys. So we cannot create 'HdRootId'. Since we need some
+-- unique identifier for EOS-wallet, we use UUID.
+-- Please note that UUID-based solution may be changed in the future.
+newtype EosHdRootId = EosHdRootId { getEosHdRootId :: UUID }
+    deriving (Eq, Ord, Show)
```
KtorZ, line 64: _[should fix]_

I'd see value in a `newEosHdRootId = EosHdRootId <$> randomIO :: IO EosHdRootId` to avoid bubbling the `uuid` and `random` package everywhere outside here.

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -228,39 +232,40 @@ createHdWallet pw mnemonic spendingPassword assuranceLevel walletName = do
...
+    let newEosRoot = EosHD.initEosHdRoot (EosHD.EosHdRootId newUUID)
+                                         walletName
+                                         assuranceLevel
+                                         addressPoolGap
+    res <- update' (pw ^. wallets) $ CreateEosHdWallet newEosRoot accountsPKs
+    case either Left (const (Right newEosRoot)) res of
+        Left e@(EosHD.CreateEosHdRootExists _) ->
+            return . Left $ CreateEosWalletFailed e
+        Right eosHdRoot ->
+            return (Right eosHdRoot)
```
KtorZ: What about:

```hs
return $ case res of
    Left e@(EosHD.CreateEosHdRootExists _) ->
        Left $ CreateEosWalletFailed e
    Right _ -> 
      Right newEosRoot
```

:sweat_smile: ?

src/Cardano/Wallet/WalletLayer.hs
```diff
@@ -95,25 +95,28 @@ instance Buildable CreateWalletError where
     build (CreateWalletError kernelError) =
         bprint ("CreateWalletError " % build) kernelError
 
-data CreateExternalWalletError =
-      CreateExternalWalletInvalidRootPK Base58PublicKeyError
-    | CreateExternalWalletError Kernel.CreateExternalWalletError
+data CreateEosWalletError =
+    CreateEosWalletNoAccountsPKs
```
KtorZ: Not sure what this is for, but I don't think it's a requirement to have non-empty list of accounts :thinking: ? Should work with an empty list.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -156,32 +159,56 @@ createWallet wallet newWalletRequest = liftIO $ do
...
+                -> V1.NewEosWallet
+                -> m (Either CreateEosWalletError V1.EosWallet)
+createEosWallet wallet newEosWalletRequest = runExceptT $ do
+    let encodedAccountsPKs = V1.neweoswalAccountsPKs newEosWalletRequest
+        addressPoolGap = maybe defaultAddressPoolGap id $
+            V1.neweoswalAddressPoolGap newEosWalletRequest
+        name = V1.neweoswalName newEosWalletRequest
+        assuranceLevel = V1.neweoswalAssuranceLevel newEosWalletRequest
+    if null encodedAccountsPKs
+        then ExceptT $ return $ Left CreateEosWalletNoAccountsPKs
```
KtorZ: cf comment above :+1:

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -156,32 +159,56 @@ createWallet wallet newWalletRequest = liftIO $ do
...
+        else do
+            let eitherPKs = map V1.mkPublicKeyFromBase58 encodedAccountsPKs
+                problems  = lefts eitherPKs
+                validPKs  = rights eitherPKs
+            if not . null $ problems
+                then ExceptT $ return $ Left $ CreateEosWalletInvalidAccountPK $ problems !! 0
```
KtorZ: All this should happen at the API-level I believe; It would suffice to have more constrained API types and have this parsing done in the JSON serializers / deserializers.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -156,32 +159,56 @@ createWallet wallet newWalletRequest = liftIO $ do
...
+                -> V1.EosWallet
+    mkEosWallet v1WalletName accountsPKs addressPoolGap v1AssuranceLevel eosHdRoot =
+        V1.EosWallet
+            { eoswalId             = mkEosWalletId eosHdRoot
+            , eoswalName           = v1WalletName
+            , eoswalAccountsPKs    = accountsPKs
+            , eoswalAddressPoolGap = addressPoolGap
+            , eoswalBalance        = V1 (mkCoin 0)
+            , eoswalAssuranceLevel = v1AssuranceLevel
+            }
```
KtorZ: No need for this indirection :+1:

KtorZ (changes requested):

src/Cardano/Wallet/API/V1/Types.hs
denisshevchenko: `Word8`?

KtorZ: or "newtype AddressPoolGap" :sweat_smile: ... even a `Word8` is too big I believe.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -997,40 +975,80 @@ instance BuildableSafeGen Wallet where
...
+
+deriveJSON Aeson.defaultOptions ''EosWalletId
+
+instance ToSchema EosWalletId where
+  declareNamedSchema = genericDeclareNamedSchema defaultSchemaOptions
+
+instance ToJSONKey EosWalletId
+
+instance Arbitrary EosWalletId where
+    arbitrary = example
```
KtorZ: _[must fix]_

Nope! That's, at best, the other way around. 

- Arbitrary = Generators we use for testing.
- Example = Human-friendly values we display in the documentation

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -652,38 +625,43 @@ mkTransactionSignatureAsBase16 :: Core.Signature Txp.TxSigData -> TransactionSig
...
 
--- | A type modelling the request for a new 'ExternalWallet',
+-- | A type modelling the request for a new 'EosWallet',
 -- on the mobile client or hardware wallet.
-data NewExternalWallet = NewExternalWallet
-    { newewalRootPK         :: !PublicKeyAsBase58
-    , newewalAssuranceLevel :: !AssuranceLevel
-    , newewalName           :: !WalletName
+data NewEosWallet = NewEosWallet
+    { neweoswalAccountsPublicKeys :: ![PublicKeyAsBase58]
```
KtorZ: _[should fix]_

Can we shove `PublicKey` directly here? Why do we need this intermediate type ? (Or at least, have a `PublicKey` inside a `PublicKeyAsBase58` instead of a raw `Text`.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -997,40 +975,80 @@ instance BuildableSafeGen Wallet where
...
+        Just uuid -> Right . EosWalletId $ uuid
+
+instance ToHttpApiData EosWalletId where
+    toQueryParam (EosWalletId uuid) = Uuid.toText uuid
+
+-- | Externally-owned sequential (EOS) wallet (mobile client or hardware wallet).
+data EosWallet = EosWallet
+    { eoswalId                 :: !EosWalletId
+    , eoswalName               :: !WalletName
+    , eoswalAccountsPublicKeys :: ![PublicKey]
```
KtorZ: _[may fix]_

I am thinking out-loud here but, do we actually need this at all :thinking: ? I mean.. Having an endpoint `/wallets/{id}/accounts`, we already have a way to get this list of public keys :) ... So in the end, to make the API representation a bit more orthogonal, we could just remove public keys completely from the representation and only keep those in our internal wallet layer type. 

What do you think ?

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -2297,6 +2315,9 @@ instance Example ShieldedRedemptionCode
 instance Example (V1 Core.PassPhrase)
 instance Example (V1 Core.Coin)
 
+instance Example EosWalletId where
+    example = EosWalletId <$> pure (fromJust . Uuid.fromText $ "c2cc10e1-57d6-4b6f-9899-38d972112d8c")
```
KtorZ: _[may fix]_

Side-note: that's one of those cases where I think `read "c2cc10e1-57d6-4b6f-9899-38d972112d8c"` is totally fine :)

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -228,39 +232,40 @@ createHdWallet pw mnemonic spendingPassword assuranceLevel walletName = do
...
+    -- For FOS-wallets we need a root key for creating 'HdRoot'. But we don't have a root key
+    -- for EOS-wallet, so use UUID as a unique identifier of the wallet. Please note that
+    -- UUID-based solution may be changed in the future.
+    newUUID <- nextRandom
+    let newEosRoot = EosHdRoot (EosHD.EosHdRootId newUUID)
+                               walletName
+                               assuranceLevel
+                               addressPoolGap
+    res <- update' (pw ^. wallets) $ CreateEosHdWallet newEosRoot accountsPKs
+    return $ case either Left (const (Right newEosRoot)) res of
```
KtorZ: _[may fix]_ 

:sweat_smile:  my point wasn't much about the `return $` (I just noticed ...) but more about the unecessary `either Left (const (Right ...))`, this is unnecessary complicated as we are truly just doing `case res of`

src/Cardano/Wallet/WalletLayer.hs
```diff
@@ -95,25 +95,25 @@ instance Buildable CreateWalletError where
...
-      CreateExternalWalletInvalidRootPK Base58PublicKeyError
-    | CreateExternalWalletError Kernel.CreateExternalWalletError
+data CreateEosWalletError =
+    CreateEosWalletInvalidAccountPublicKey Base58PublicKeyError  
+  | CreateEosWalletError Kernel.CreateEosWalletError
 
 -- | Unsound show instance needed for the 'Exception' instance.
-instance Show CreateExternalWalletError where
+instance Show CreateEosWalletError where
     show = formatToString build
```
KtorZ, line 104: _[must fix]_ 

Arg! No! Didn't catch that before but, please no custom `Show` instances :(, those should be derived only and always respect `show . read = id`

If there's any other type doing so, may you please open a ticket to remove them all :)

src/Cardano/Wallet/WalletLayer.hs
```diff
@@ -191,21 +191,21 @@ instance Buildable DeleteWalletError where
...
-      DeleteExternalWalletInvalidRootPK Base58PublicKeyError
-    | DeleteExternalWalletError Kernel.UnknownHdRoot
+data DeleteEosWalletError =
+      DeleteEosWalletInvalidRootPK Base58PublicKeyError
+    | DeleteEosWalletError Kernel.UnknownHdRoot
 
 -- | Unsound show instance needed for the 'Exception' instance.
-instance Show DeleteExternalWalletError where
+instance Show DeleteEosWalletError where
     show = formatToString build
```
KtorZ, line 200: _[must fix]_

Cf above.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -156,32 +158,44 @@ createWallet wallet newWalletRequest = liftIO $ do
...
+        addressPoolGap = maybe defaultAddressPoolGap id $
+            V1.neweoswalAddressPoolGap newEosWalletRequest
+        name = V1.neweoswalName newEosWalletRequest
+        assuranceLevel = V1.neweoswalAssuranceLevel newEosWalletRequest
+        eitherPublicKeys = map V1.mkPublicKeyFromBase58 encodedAccountsPublicKeys
+        problems = lefts eitherPublicKeys
+        validPublicKeys = rights eitherPublicKeys
+    if not . null $ problems
+        then ExceptT $ return $ Left $ CreateEosWalletInvalidAccountPublicKey $ problems !! 0
+        else do
```
KtorZ: _[should fix]_ 

Reflecting on the comment above about having a more precise type for public keys in `NewEosWallet`. This should be caught earlier. And as a rule of thumb, we can eliminate a whole class of errors by having better types. There's no need to handle conversions from text to public keys if we directly manipulate public keys :smile:

KtorZ (changes requested):
Getting there :stuck_out_tongue: 
Still some points that must be changed, and some that I think can be tackled in a new PR!

src/Cardano/Wallet/WalletLayer.hs
denisshevchenko: > If there's any other type doing so, may you please open a ticket to remove them all :)

Yes, there's a LOT of other types doing so. Ticket: https://github.com/input-output-hk/cardano-wallet/issues/71

src/Cardano/Wallet/API/V1/Types.hs
denisshevchenko: Type `PublicKeyAsBase58` will be removed in the next PRs. We don't need it actually. :-)

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -79,18 +83,18 @@ instance Buildable CreateWalletError where
...
-data CreateExternalWalletError =
-      CreateExternalWalletFailed HD.CreateHdRootError
-      -- ^ When trying to create the 'ExternalWallet', the DB operation failed.
+data CreateEosWalletError =
+      CreateEosWalletFailed EosHD.CreateEosHdRootError
+      -- ^ When trying to create the 'EosWallet', the DB operation failed.
 
-instance Arbitrary CreateExternalWalletError where
+instance Arbitrary CreateEosWalletError where
     arbitrary = oneof []
```
KtorZ, line 91: _[should fix]_

Just noticing now but not sure that this will play with Quickcheck when generating stuff :sweat_smile:

denisshevchenko: I just took it from:

```
instance Arbitrary CreateWalletError where
    arbitrary = oneof []

```

:-)

KtorZ (approved):
LGTM :+1:

As a reminder:

_should fix_

- [ ] arbitrary instance of `CreateEosWalletError` -> `oneof []` (this throws as runtime :upside_down_face: )
- [ ] provide a smart constructor for generating new `EosRootId`
- [ ] define a tailored type for `addressPoolGap`
- [ ] clean-up unused types and constructors (related to `PublicKeyAsBase58` mostly)

src/Cardano/Wallet/Kernel/Wallets.hs
KtorZ: I know :upside_down_face:  .... 

image

Guess we have never used those `Arbitrary` instances ...

PR #66, opened 2018-11-19
Setup Travis to work with Ubuntu EC2 fast-booting containers (faster and more reliable than OSX full-blown VMs ones)

KtorZ:
# Linked Issue

<p align="right">#None</p>

### Acceptance criteria:

¯\\\_(ツ)\_/¯

# Comments

We went for OSX VMs in a first time to get to work with the CI. However, they are a bit slower than the Ubuntu one and especially, they are waaaay longer to be scheduled by the provider. Using Ubuntu / non-sudo VMs relies on AWS EC2 containers are are booted in a second after a push. 
The only quirks was to install and cache `librocksdb-dev` (without sudo!). Turns out we can do that in a few steps by just pulling in the library and pre-compiled binaries easily obtained on a debian package (`.deb`). 
Then the trick is to point the `PATH` and `LIBRARY_PATH` to the right location, inside our `.local` cache folder to work fine. 

# Checklist

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 # Use new container infrastructure to enable caching
 sudo: false
 
-# Use OSX to by-pass issue with CouchDB PGP Key on Linux
-os: osx
-
 # Generic language, we provide our custom builds
 language: generic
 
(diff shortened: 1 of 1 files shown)


Anviking (dismissed):
🎉

.travis.yml
```diff
@@ -57,9 +52,13 @@ before_install:
...
   if: type != cron AND commit_message =~ /ci@update lts/
   script:
+  # Prepare directories
+  - mkdir -p $HOME/.local/bin $HOME/.local/lib $HOME/.local/include
   # Install Stack
-  - mkdir -p $HOME/.local/bin $HOME/.local/.stack-work
-  - travis_retry curl --insecure -L https://www.stackage.org/stack/osx-x86_64 | tar xz --strip-components=1 --include '*/stack' -C ~/.local/bin
+  - travis_retry curl -L https://get.haskellstack.org/stable/linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
+  # Install rocksdb
+  - git clone --depth 1 --branch=librocksdb-dev $(git remote get-url origin) $HOME/.local/lib && rm -rf $HOME/.local/lib/.git
```
Anviking: Seemed weird to me at first to use a custom branch in this repo, but I guess it's rather nice

KtorZ: Well, I didn't want to pollute the `develop` branch with some `.ci` folder or something like this. And having that in a gist is quite ad-hoc and out-of-scope. So, the nicest compromise I found was to have that in an orphan branch that we can update if necessary.

Anviking (approved):

PR #70, opened 2018-11-21
Review Fee Calculation (Backport from Cardano-SL)

KtorZ:
# Linked Issue

<p align="right">input-output-hk/cardano-sl/pull/3861</p>

### Acceptance criteria:

¯\\\_(ツ)\_/¯

# Comments

This is just a backport of the fix done in `cardano-sl/develop` and `cardano-sl/release-2.0.0` until we get this submodules setup correctly done....

# Checklist

---

### DIFF
--- a/src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
+++ b/src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
@@ -1,10 +1,11 @@
 {-# LANGUAGE BangPatterns #-}
 
-module Cardano.Wallet.Kernel.CoinSelection.Generic.Fees (
-    ExpenseRegulation(..)
-  , FeeOptions(..)
-  , adjustForFees
-  ) where
+module Cardano.Wallet.Kernel.CoinSelection.Generic.Fees
+    ( ExpenseRegulation(..)
+    , FeeOptions(..)
+    , CoinSelFinalResult(..)
+    , adjustForFees
+    ) where
 
 import           Universum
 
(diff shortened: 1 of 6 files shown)


src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
```diff
@@ -27,130 +28,236 @@ data ExpenseRegulation =
...
+        case foExpenseRegulation feeOptions of
+          ReceiverPaysFee ->
+            coinSelLiftExcept $ receiverPaysFee feeOptions inps outs chgs
+
+          SenderPaysFee ->
+            senderPaysFee pickUtxo feeOptions inps outs chgs
+
+    let neInps = case inps' of
+            []   -> error "adjustForFees: empty list of inputs"
+            i:is -> i :| is
```
parsonsmatt, line 102: `CoinSelT` has an error branch. We should be using that instead of `error`.

src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
```diff
@@ -27,130 +28,236 @@ data ExpenseRegulation =
...
+
+    let neOuts = case outs' of
+            []   -> error "adjustForFees: empty list of outputs"
+            o:os -> o :| os
```
parsonsmatt, line 106: same as above

src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
```diff
@@ -167,13 +274,19 @@ reduceChangeOutputs totalFee cs =
...
+    -> [UtxoEntry dom]
+    -> [Output dom]
+    -> [Value dom]
+    -> Fee dom
+feeUpperBound FeeOptions{..} inps outs chgs =
     foEstimate numInputs outputs
   where
-    numInputs = fromIntegral $ sum (map (sizeToWord . coinSelInputSize) css)
-    outputs   = concatMap coinSelOutputs css
+    numInputs = fromIntegral $ sizeToWord $ selectedSize $ foldr' select emptySelection inps
```
parsonsmatt: If `foldr'` is a strict `foldr` then we should use `foldl' (flip select)` instead -- a strict `foldr'` must walk the entire list and accumulate the spine in memory before it can begin accumulating the result, while `foldl'` is tail recursive and can GC the list as it works over it.

parsonsmatt (dismissed):
Unsafe usage of `error` should be removed and replaced with the error mechanism in the `CoinSelT` stack.

src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
KtorZ: Note that here, those aren't error case we want to handle but actual invariant violation.
The kerbel uses this error "pattern" in multiple places to put light on invariant. This is not something we want to shove in our error type.

src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
KtorZ: Indeed, though inputs are usually quite small here (we're talking about a handful of coins in general, not millions).
Should favor foldl' anyway :+1:

denisshevchenko (approved):

Issue #33, opened 2018-11-07
Address Pool Module


KtorZ:
# Context

<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

Here, we are looking for a module which can keep track of address pools and answer the question:

```hs
isOurs: Address -> Maybe (WalletId, AccountId) 
```

This can be achieve in various way but a few pointers:

- We can have an opaque `AddressPool` types which is implemented behind the scene as some specific 
  data-structure (like a `Map`); managing this structure is what the module does.

- How to generate new items should be provided to the module (ideally, the module doesn't know about addresses).

- We should have a way to initialize the module from a list of items to track.

# Decision

<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

FIXME: To be completed with design decision (@uroboros)

# Acceptance Criteria

<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

1. - [x] `cardano-wallet` _should_ provide a new module to host Address Discovery utilities as explained in BIP-44.
1. - [x] The module from #33.1 _should_ provide an abstract and self-contained API.

---

# PR

| Number    | Base      |
| ---       | ---       |
| [118][PR-1] | `develop` |
| [129][PR-2] | `develop` |

[PR-1]: https://github.com/input-output-hk/cardano-wallet/pull/118
[PR-2]: https://github.com/input-output-hk/cardano-wallet/pull/129

# QA 

<WHEN IN PROGRESS> 
<How are we covering acceptance criteria. Give here manual steps or tests that
are covering the above criteria.>

| Criteria | Coverage |
| --- | --- | 
| Creation & validation of address gap | See [Spec/AddressPoolGap](https://github.com/input-output-hk/cardano-wallet/blob/develop/test/unit/Test/Spec/AddressPoolGap.hs) |
| Creation, lookups and extension of address pool | See [Spec/AddressPool](https://github.com/input-output-hk/cardano-wallet/blob/develop/test/unit/Test/Spec/AddressPool.hs) |

```
$ stack test cardano-wallet:unit --test-arguments "--match Pool"
AddressPool
  lookupAddressPool
    +++ OK, passed 100 tests:
    62% hit outside pool
    38% hit within pool
[...]

PR #73, opened 2018-11-22
[BIP44-33] AddressPool - core types and algorithm to support address …

uroboros:
…discovery for EOS/FOS wallets

# Linked Issue

https://github.com/input-output-hk/cardano-wallet/issues/33

# Comments

First draft of AddressPool module, sans any tests.

### DIFF
--- a/src/Cardano/Wallet/Kernel/HdSeq.hs
+++ b/src/Cardano/Wallet/Kernel/HdSeq.hs
@@ -1,0 +1,385 @@
+{-# LANGUAGE AllowAmbiguousTypes        #-}
+
+module Cardano.Wallet.Kernel.HdSeq
+       (
+         addrInPool
+       ,
+         firstAddressRange
+       ,
+         initEOSAddressPool
+       ,
+         initFOSAddressPool
+       ,
+         newAccountPool
+       ,
+         selectOurs
+       ,
+         AccountPoolStatus (..)
+       ,
+         AccountPool (..)
+       ,
+         AddressPool (..)
+       ,
+         AccountPoolAmendment (..)
+       ) where
+
+import           Universum
+
+import qualified Data.Map as Map
+import qualified Data.List.NonEmpty as NE
+
+import           Cardano.Wallet.Kernel.DB.HdWallet(HdAddressIx (..))
+
+{-------------------------------------------------------------------------------
+  The AddressPool module enables BIP44-style address and account discovery
+  for 2 different styles of wallets:
+
+  * Externally Owned Sequential (EOS) wallets *
+
+  For EOS wallets, we can't add new accounts and therefor can't discover any
+  accounts that were not provided upfront.
+
+  In this style of address discovery we can make no assumptions about the sequence
+  of accounts, we can only discover addresses in the given wallets (and indeed
+  add new addresses to be discovered)
+
+  * Fully Owned Sequential (FOS) wallets *
+
+  For FOS wallets, we can automatically create (and thereby discover) new accounts,
+  since we have the necessary wallet credentials.
+
+  In this style of address/account discovery, the accounts are considered to be
+  in sequence, and we can discover all accounts for the wallet with this algorithm.
+-------------------------------------------------------------------------------}
+
+{-------------------------------------------------------------------------------
+  AddressPool Types
+-------------------------------------------------------------------------------}
+
+-- | When an address is discovered in a sequence of addresses, if that
+--   address is "close enough" to the end of the sequence, we want to generate
+--   more addresses in the sequence (to increase the range of addresses that
+--   may be discovered). The AddressPoolGap expresses this notion of "close enough".
+type AddressPoolGap = Word32
+
+-- | An AccountPool is considered _live_ if any address in the pool has already been
+--   discovered, otherwise the pool is _pending_
+data AccountPoolStatus = LivePool | PendingPool
+    deriving (Eq, Ord, Show)
+
+-- | Constructor for creating a new sequential Hd account, takes as argument
+--   the previous Hd account in the sequence (this is only relevant to FOS wallets)
+type NewAccount account = account -> account
+
+-- | Constructor for creating a new sequential Hd address, takes as argument
+--   the address index
+type NewAddress address = HdAddressIx -> address
+
+-- | The pool of "our" addresses for an account
+
+-- Note: the account and address types are polymorphic in order to support
+-- varying wallet types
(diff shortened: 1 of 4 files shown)


src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+         selectOurs
+       ,
+         AccountPoolStatus (..)
+       ,
+         AccountPool (..)
+       ,
+         AddressPool (..)
+       ,
+         AccountPoolAmendment (..)
+       ) where
```
KtorZ, line 24: so
many 
new
lines

:smile:

src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+
+{-------------------------------------------------------------------------------
+  AddressPool Types
+-------------------------------------------------------------------------------}
+
+-- | When an address is discovered in a sequence of addresses, if that
+--   address is "close enough" to the end of the sequence, we want to generate
+--   more addresses in the sequence (to increase the range of addresses that
+--   may be discovered). The AddressPoolGap expresses this notion of "close enough".
+type AddressPoolGap = Word32
```
KtorZ, line 63: _may fix_

`Word32` sounds like a lot. This has some overlap with #85; so I'd suggest to align with @denisshevchenko and either conflate the two modules into one, or have this one import the AddressPoolGap type.

src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+       ,
+         AccountPoolAmendment (..)
+       ) where
+
+import           Universum
+
+import qualified Data.Map as Map
+import qualified Data.List.NonEmpty as NE
+
+import           Cardano.Wallet.Kernel.DB.HdWallet(HdAddressIx (..))
```
KtorZ, line 31: _must fix_

This one doesn't belong here I believe.

src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+-------------------------------------------------------------------------------}
+
+-- | Construct a sequenced range of addresses, starting at index 'fromIx'
+-- TODO @uroboros PRECONDITION - range >= 2
+newAddressRange :: AddressPoolGap
+                -> HdAddressIx
+                -> NewAddress address
+                -> NonEmpty (address,HdAddressIx)
+newAddressRange range (HdAddressIx fromIx) newAddress
+    = (f fromIx) NE.:| (map f [fromIx + 1 .. fromIx + range - 1])
```
KtorZ, line 143: _may fix_ 

No need for qualified import here, `:|` is common enough to be used as a operator.

src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+      ,
+        _addrPoolGap        = gap
+      ,
+        -- the constructor to extend the address pool range for an account
+        _addrPoolNewAddress = newAddress
+      ,
+        -- the constructor to create the next account in a sequence (given the
+        -- previous account)
+        _addrPoolNewAccount = Just newAccount
+    }
```
KtorZ, line 253: _must fix_

Let's not bother about FOS wallet now and proceed in steps. For now, we only consider case where we ignore relationship accounts have to each others and we only build from a set of account public keys.

src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+--
+--   In other words, the address pools we match against may change as we
+--   discover new addresses.
+addrsInPools :: (Eq account, Ord address)
+             => AddressPoolGap
+             -> Maybe (NewAccount account)
+             -> NewAddress address
+             -> [AccountPool account address]
+             -> [address]
+             -> [(account, HdAddressIx)]
```
KtorZ, line 358: _must fix_

This looks a lot complicated for an exposed function. One of the goal we want to achieve by making an isolated module is to provide some easy handles on an opaque type such that, it's easy for users of this module to achieve what they want. 

```hs
addrInPool 
  :: AddressPool account address
  -> address
  -> (AddressPool account address, Maybe (account, AddressIdx))
addrInPool ... =
```

That's all we should need, right ?

src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+       ,
+         selectOurs
+       ,
+         AccountPoolStatus (..)
+       ,
+         AccountPool (..)
+       ,
+         AddressPool (..)
+       ,
+         AccountPoolAmendment (..)
```
KtorZ, line 23: _must fix_

We expose too many things from this module. The point of making this an isolated module is to encapsulate the complexity and implementation details inside the module and make sure they don't get out! 
Underneath it all, the API boils down to this:

```hs
module 
  ( AddressPool
  , initAddressPool
  , addrInPool -- cf comment below
  ) where
```

We may provide some extra "debugging" / testing helpers like "totalPoolSize" to help us write properties about those. But, conceptually, I believe we don't need to expose more than that, don't we?

src/Cardano/Wallet/Kernel/HdSeq.hs
```diff
@@ -0,0 +1,385 @@
...
+
+-- | Constructor for creating a new sequential Hd address, takes as argument
+--   the address index
+type NewAddress address = HdAddressIx -> address
+
+-- | The pool of "our" addresses for an account
+
+-- Note: the account and address types are polymorphic in order to support
+-- varying wallet types
+data AccountPool account address = AccountPool {
```
KtorZ, line 82: _may fix_ 

What above shoving in there the `Ord` and `Eq` instances for `account` and `address` ?

KtorZ (changes requested):
- Let's not even mention FOS wallets at this stage
- We ought to provide a "dead simple" API for this module and have the complexity lives inside it

Haven't much reviewed the algorithm so-to-speak as I think the various comment will provoke some important refactoring :/ 
Also, make sure to align with @denisshevchenko on the `AddressPoolGap` type.

denisshevchenko:
Closed, because simplified version is here: https://github.com/input-output-hk/cardano-wallet/pull/118

Issue #89, opened 2018-11-22
Move some (long) unit tests into nightly tests


KtorZ:
# Context

<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

Some unit tests are actually stress tests which try to trigger race conditions or concurrency issues. Because of their very nature, those tests need to generate a lot of data and run for a while. 
This can be quite painful for our CI and iteration time.

# Decision

<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

Create a nightly job on CI and move some long-running tests as nightly tests.

# Acceptance Criteria

<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

1. - [x] `unit` tests _must_ only contains fast executed tests.
2. - [x] `nightly` tests _must_ be ran once a day and loudly fail in case of failure.

---

# PR

| Number    | Base      |
| ---       | ---       |
| [#90][PR-1] | `develop` |

[PR-1]: https://github.com/input-output-hk/cardano-wallet/pull/90

# Consequences

<To be completed in retrospective. What becomes easier or more difficult to do
because of this change.>

"PR CI" is now back to its usual time around 15 minutes. We may still reduce this time by making some optimization in our testing, some of them still take an unreasonable amount of time for what they actually do. 

![image](https://user-images.githubusercontent.com/5680256/49205956-d51c5e80-f3b0-11e8-8a3b-c5a9214d3e1a.png)

PR #90, opened 2018-11-22
Create 'nightly' tests and move sync tests from TxMetaStorage in there

KtorZ:
# Linked Issue

<p align="right">#89</p>

### Acceptance criteria:

1. - [x] `unit` tests _must_ only contains fast executed tests.
2. - [x] `nightly` tests _must_ be ran once a day and loudly fail in case of failure.

# Comments

Some tiny duplication in the nightly, but all for a greater good. Ideally, we would like to have those `Util` in a library...

On Travis:

image

# Checklist

---

### DIFF
--- a/cardano-wallet.cabal
+++ b/cardano-wallet.cabal
@@ -49,7 +49,6 @@ library
     , aeson-pretty
     , async
     , base58-bytestring
-    , basement
     , beam-core
     , beam-migrate
     , beam-sqlite
@@ -462,7 +461,6 @@ test-suite unit
       base
     , acid-state
     , aeson
-    , async
     , bytestring
     , cardano-crypto
     , cardano-sl
@@ -486,7 +484,6 @@ test-suite unit
     , cryptonite
     , data-default
     , directory
-    , filepath
     , formatting
     , hedgehog
     , hspec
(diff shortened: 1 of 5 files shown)


cardano-wallet.cabal
```diff
@@ -551,12 +547,6 @@ test-suite unit
       Util.Buildable
       Util.Buildable.Hspec
       Util.Buildable.QuickCheck
-      Util.Distr
-      Util.GenHash
-      Util.Histogram
-      Util.MultiSet
-      Util.QuickCheck
-      Util.Range
```
KtorZ, line 559: `weeder` told me :no_mouth:

denisshevchenko (approved):

cardano-wallet.cabal
Anviking: Are these not used at all now? `git grep "Util.Distr"` returns imports from other files, but no match from a .cabal

Doesn't this mean we have dead files?

KtorZ: They're used from InputSelection stuff, which isn't so-to-speak unit tests. Located there for historical reason. I am thinking about moving those into their own dedicated executable.

Issue #117, opened 2018-11-27
(CI) Review structure and implementation of integration tests 


KtorZ:
# Context

<WHEN PROPOSED>
<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

This is roughly a "backport" of CO-256, split across multiple tasks. For a long time now, we have been willing to refactor our current integration tests such that they will be easier to maintain, but also and in particular, easier to comprehend for non-haskellers. 
Right now, a lot of code in there suffers from heavy duplication and is overly verbose to the point where it is not obvious nor not even simple to understand what a particular spec does. The reasons behind this "chaos" are plural but mostly due to the fact that, integration tests were written in a hurry and "maintained" by many persons without clear design consensus or dialogue. 

# Decision

<WHEN PROPOSED>
<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

Therefore, we will review entirely the design and try to stick with something that is simple to use. The overall goal we're trying to achieve here is to generate very readable specs, in particular for the non-haskellers. It should be obvious at first glance what a given test is actually testing.

# Acceptance Criteria

<WHEN PROPOSED>
<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

1. - [x] Integration tests _must_ be as much as possible readable for non-haskellers
1. - [x]  Integration tests _must_ be ran as part of the CI
1. - [x] Integration tests _should_ be quick and easy to run

---

# PR

<WHEN IN PROGRESS>
<List of all PRs related to this ticket.>

| Number   | Base      |
| ---      | ---       |
| #126       | `develop` |
| #140 | `develop` |
| #149 | `develop` | 

# QA 

<WHEN IN PROGRESS> 
<How are we covering acceptance criteria. Give here manual steps or tests that
are covering the above criteria.>

| Criteria | Coverage |
| ---      | ---      |
| 1.        | Checked with @piotr-iohk; Also, created an integration tests DSL with focus on reading |
[...]

PR #126, opened 2018-11-30
Integration Test Framework - Part 1

KtorZ:
# Linked Issue

<p align="right">#117</p>

# Overview

1. - [x] Define a framework baseline for integration tests
2. - [x] Translate (almost) each existing integration tests to this new framework (are missing: the two tests related to asset-locked wallets, will do later). 
3. - [x] Remove existing integration tests

# Comments

This is... a lot, but most of the work happens only in a few place:

- `Test.Integration.Framework.Scenario`: defines a `Scenario` monad within which each test is running. This is basically a state monad, that is used as a reader for most test action except "setup" (cf remark below). There's a bit of plumbing here to make sure that things are hooked up into `HSpec` framework in such way that the footprint is rather small. In the end, we can leverage all of the `HSpec` tooling around this.

- `Test.Integration.Framework.DSL`: defines a DSL that is used to define integration tests. Ideally, we would use only combinators from that module to write test. This is to favor a very consistent, succinct and readable format for writing integration tests (remember that this may be read and analyzed by QA on the long run, and we can even imagine some of them being written by QA as well). So in a way, if I am only interested in writing and reading integration tests, I should only be looking at the DSL top-level interface. Note that, there's quite some boilerplate in that file and I hope that in the long run, we may be able to replace most of it with a fresh version of our API types :) 

Important remark: I haven't ran any of this yet :upside_down_face:  ... Actually, there's some work to be done with the cluster to decouple it from the wallet (hopefully next week). Before that, it should be possible to create a WalletClient by hand and test everything out, we will simply not get coverage out of that and need to start tests and server separately. That's probably what I'll do first to test that swiftly. 

Finally, there's also some "bootstrapping" necessary here. Ideally, we would like to reset the state of the blockchain between each test run. That's really not practical and would make all tests very long to run. Instead, I have been following a slightly different approach: some tests can be ran without any coins, some (most of them) needs a few funds. We have a few "genesis" wallets at our disposal with millions of ADA one each. So, before starting any test, we actually divide this amount in `k` and provide each test with some ADA they can use for testing. The amount of ADA available depends on the amount of tests, but we should have enough ADA to run A LOT of them without missing any. 

# Checklist

- [ ] ~~I've kept this PR simple, on-the-spot, free of cosmetic changes.~~ Sorry :( 

---


test/integration/Test/Integration/APISanityChecks.hs
```diff
@@ -0,0 +1,132 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Test.Integration.APISanityChecks
```
KtorZ, line 3: Pretty much copy-pasted from before. I only remove one check saying that all `DELETE` should be `204` (that's actually not true since some of our DELETE are `404`) and also increased the number of randomly generated transactions from 100 to 1000... 
I'd actually like to see what kind of requests are generated and if this whole thing actually does check something :thinking:  ... but for now, I just kept it.

test/integration/Test/Integration/Framework/Orphan.hs
```diff
@@ -0,0 +1,37 @@
...
+    (+)    = error numForbiddenOperation
+    (-)    = error numForbiddenOperation
+    (*)    = error numForbiddenOperation
+    abs    = error numForbiddenOperation
+    signum = error numForbiddenOperation
+    fromInteger = mkCoin . fromInteger
+numForbiddenOperation :: Text
+numForbiddenOperation =
+    "Call forbidden operation: 'Num' for 'Coin' is only defined to ease\
+    \ literal notation when writing integration tests scenarios."
```
KtorZ: Perhaps a `newtype` here.. I don't know

test/integration/Test/Integration/Framework/Request.hs
```diff
@@ -0,0 +1,61 @@
...
+
+-- | Removes some boilerplates getting the content of a 'Resp' directly.
+class Request originalResponse where
+    type Response originalResponse :: *
+
+    -- | Run a given request and transform the response
+    request
+        :: forall m ctx. (MonadIO m, MonadReader ctx m, HasHttpClient ctx)
+        => (WalletClient IO -> IO (Either ClientError originalResponse))
+        -> m (Either ClientError (Response originalResponse))
```
KtorZ, line 29: I wonder if we couldn't modify the client to actually do that for us directly. The `jsend` wrapper is pretty much useless (except for the metadata maybe .. and even... ). 

By the by, went for a typeclass here because responses in our `WalletClient` don't have uniform types .. some of them returns a `Either ClientError (WalletResponse a)` and some of them returns a `Either ClientError ()`

test/integration/Test/Integration/Framework/Scenario.hs
```diff
@@ -0,0 +1,60 @@
...
+instance Example (Scenario s IO ()) where
+    type Arg (Scenario s IO ()) = MVar s
+    evaluateExample (Scenario io) _ action _ =
+        action runAndPersist >> return (Result "" Success)
+      where
+        runAndPersist :: MVar s -> IO ()
+        runAndPersist mvar = do
+            s <- takeMVar mvar
+            (_, s') <- runStateT io s
+            putMVar mvar s'
```
KtorZ, line 47: So, here's the trick. We save the context from a scenario to another. The only thing that actually gets modified in the context are the faucet wallets. So here, we're basically keeping track of a counter, and it gets incremented between each test, like a good global variable. In theory, the tests themselves only need this to be a `MonadReader`.

test/integration/Test/Integration/Framework/Scenario.hs
```diff
@@ -0,0 +1,60 @@
...
+
+-- | Execute all scenarios in sequence providing the given context 'ctx'
+runWithContext :: forall ctx. ctx -> SpecM ctx () -> Spec
```
KtorZ, line 50: So, HSpec can't run anything else than `Spec = SpecM (ActionWith ()) ()`; So this makes the transformation, provided a context.

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -0,0 +1,172 @@
...
+            . destinations [RandomDestination]
+
+        response <- postTransaction
+            $ source wallet
+            . destination (addr, 14)
+
+        verify response
+            [ expectTxInHistoryOf wallet
+            , expectTxStatusEventually [Creating, Applying, InNewestBlocks, Persisted]
+            ]
```
KtorZ, line 26: And finally, some test scenario ! Notice they're all divided in 3 "paragraphs"

- Setup
- Action
- Expectations

The DSL makes is possible to almost read it like a sentence and combinators are re-usable between types to make things quite smooth.

.stylish-haskell.yaml
```diff
@@ -222,7 +222,9 @@ language_extensions:
   - DefaultSignatures
   - DeriveDataTypeable
   - DeriveGeneric
+  - DerivingStrategies
```
parsonsmatt, line 225: :joy:

cardano-wallet.cabal
```diff
@@ -345,87 +345,6 @@ executable cardano-generate-swagger-file
...
-      HardwareWalletSpecs
-      QuickCheckSpecs
-      RandomStateWalk
-      SetupTestEnv
-      TransactionSpecs
-      Types
-      Util
-      WalletSpecs
-
-
```
parsonsmatt, line 427: :joy:

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,859 @@
...
+
+-- | Just a type-alias to 'SpecM', like 'scenario'. Ultimately, everything is
+-- made in such way that we can use normal (albeit lifted) HSpec functions and
+-- utilities if needed (and rely on its CLI as well when needed).
+type Scenarios ctx = SpecM (MVar ctx) ()
+
+-- | Just a slightly-specialized alias for 'it' to help lil'GHC. Also, makes
+-- sure the wallet has been cleared out completely before running the scenario.
+scenario
+    :: (HasCallStack, Example (Scenario Context IO ()))
```
parsonsmatt: This constraint is dodgy. It is fully saturated, which means we expect some orphan instance to come by and satisfy it.

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,859 @@
...
+    , expectTxStatusNever
+    , expectWalletError
+    , expectWalletUTxO
+    ) where
+
+import           Universum hiding (fromRight, getArgs, second)
+
+import           Control.Concurrent (threadDelay)
+import           Control.Concurrent.Async (Async, async, race, wait)
+import           Data.Generics.Product.Fields (HasField, field)
```
parsonsmatt: `generic-lens` <3 <3 <3

test/integration/Test/Integration/Framework/Orphan.hs
parsonsmatt: Why not just use a constructor? or a function? I'm not a fan of abusing `Num` for things like this, where *all* the methods are errors.

test/integration/Test/Integration/Framework/Scenario.hs
```diff
@@ -0,0 +1,60 @@
...
+
+import           Universum
+
+import           Test.Hspec.Core.Spec (ActionWith, Example (..), Item (..),
+                     Result (..), ResultStatus (..), Spec, SpecM, mapSpecItem)
+
+
+-- | A Wrapper around 'StateT' around which we define a few instances. The most
+-- interesting one is 'Example'
+newtype Scenario s m a = Scenario (StateT s m a)
```
parsonsmatt, line 14: Beware -- `StateT` is single-threaded state, and this is usually not what folks want in th epresence of concurrency and exceptions.

parsonsmatt (dismissed):
These look great!

I would suggest using `HasType` rather than `HasField`, as it's a bit more general and a bit less fragile (eg changing a field name breaks `HasField` instances, and `HasType` is capable of digging into datatypes recursively to find a value).

test/integration/Test/Integration/Framework/Orphan.hs
KtorZ: Well, the focus is really on keeping the end-to-end scenarios simple (read as: for non-haskellers), so I am really trying to remove as much noise as possible.

test/integration/Test/Integration/Framework/DSL.hs
KtorZ: Hmmm... right my bad, there's no need for a constraint actually. I just "copy/pasted" the signature of `it` without much thinking :man_shrugging:

test/integration/Test/Integration/Framework/Scenario.hs
KtorZ: Indeed, though here all the tests are running sequentially in a single-thread. We _could_ possibly run them in parallel I believe, and it shouldn't matter much because we don't do any concurrent business inside the `StateT`.
It'd would require to adjust where the `MVar` is manipulated however; but that's clearly not the intent anyway or at least, not in the first iteration :)

test/integration/Test/Integration/Framework/Orphan.hs
parsonsmatt: Confusion can come from many things -- excess polymorphism can certainly do it. `coins 0` is slightly noisier than `0` but is more explicit and monomorphic, which should provide for better error messages too.

test/integration/Test/Integration/Framework/Scenario.hs
parsonsmatt: This can also be defined as `runWithContext ctx = beforeAll (pure ctx)`

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+-- utilities if needed (and rely on its CLI as well when needed).
+type Scenarios ctx = SpecM (MVar ctx) ()
+
+-- | Just a slightly-specialized alias for 'it' to help lil'GHC. Also, makes
+-- sure the wallet has been cleared out completely before running the scenario.
+scenario
+    :: String
+    -> Scenario Context IO ()
+    -> Scenarios Context
+scenario title spec = it title (successfulRequest Client.resetWalletState >> spec)
```
parsonsmatt, line 129: so this is basically `before_ (successfulRequest Client.resetWalletState) $ it title spec` and `before` basically applies to every item in an `hspec` test.

parsonsmatt: So, another way to structure this would be:

```haskell
before (successfulRequest Client.resetWalletState) $ do
  it "does a thing" $ do
     ...

  it "does another thing" $ do
    blaaah
```

Not necessary, just a tip on hspec :)

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+-- MODIFIERS
+--
+
+accountIndex
+    :: (HasType (Maybe AccountIndex) s)
+    => AccountIndex
+    -> s
+    -> s
+accountIndex i =
+    (.~) typed (Just i)
```
parsonsmatt: I'd rather see the operator used infix or the named function used:

```suggestion
    typed .~ Just i
```

```suggestion
    set typed (Just i)
```

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+
+assuranceLevel
+    :: (HasType AssuranceLevel s)
+    => AssuranceLevel
+    -> s
+    -> s
+assuranceLevel =
+    (.~) typed
```
parsonsmatt: likewise, this is

```suggestion
    set typed
```

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+
+destination
+    :: HasType [PaymentDistribution] s
+    => (Address, Coin)
+    -> s
+    -> s
+destination (a, c) =
+    (%~) typed (PaymentDistribution (V1 a) (V1 c) :)
```
parsonsmatt:
```suggestion
    over typed (PaymentDistribution (V1 a) (V1 c) :)
```

or 

```suggestion
    typed %~ (PaymentDistribution (V1 a) (V1 c) :)
```

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+walletName =
+    (.~) typed
+
+walletOperation
+    :: HasType WalletOperation s
+    => WalletOperation
+    -> s
+    -> s
+walletOperation =
+    (.~) typed
```
parsonsmatt: note that `set typed` exists in the class as `setType`: http://hackage.haskell.org/package/generic-lens-1.1.0.0/docs/Data-Generics-Product-Typed.html#v:setTyped

Therefore, many of these definitions can simply be:

```haskell
walletOperation :: HasType WalletOperation s => WalletOperation -> s -> s
walletOperation = setTyped
```

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+
+-- | The type signature is more scary than it seems. This drills into a given
+--   `a` type through the provided lens and sees whether field matches.
+--
+--   e.g.
+--     do
+--       expectFieldEqual #walAssuranceLevel AssuranceStrict response
+--       expectFieldEqual #walName "My Wallet" response
+expectFieldEqual
+    :: ( MonadIO m, MonadFail m, Show a, Eq a, lens ~ ((a -> Const a a) -> s -> Const a s))
```
parsonsmatt: Why not use the type alias, `Getter s a`?

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+    Right s -> a `shouldBe` view lens s
+
+-- | Expects entire equality of two types
+expectEqual
+    :: (MonadIO m, MonadFail m, Show a, Eq a)
+    => a
+    -> Either ClientError a
+    -> m ()
+expectEqual =
+    expectFieldEqual ($)
```
parsonsmatt: This threw me for a loop for a bit -- I'd expect `expectFieldEqual id` to be a bit clearer.

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+
+--
+-- HELPERS
+--
+
+notInitialized
+    :: (HasCallStack, Typeable s)
+    => Proxy s
+    -> a
+notInitialized proxy =
```
parsonsmatt: I'm feeling *really* iffy on this. It's used in 12 other places, and basically means that we're throwing away the "If it type checks, it works!" feature for Haskell for *anything* that calls one of the functions. I don't like that.

It seems that we're using this for almost all of the fields in all of the use sites. We only provide a few default values. It would be more direct to have the caller pass in the entire data structure -- right now, uses are using lensy setters (eg `walletId wallet :: HasType WalletId s => s -> s`). A more direct construction would use:

```haskell
getAccountBalance :: AccountBalance -> ...

getAccountBalance AccountBalance { _walletId = walId wallet } 
```

Then, with strict fields, the user gets an error if they forget to set something.

If we want to allow defaults, then we can use functions of the form `(defParam -> defParam2 -> Thing)`, like:

```haskell
getAccountBalance (\defWalletId defOtherThing -> AccountBalance myThing defWalletId defOtherThing) ...
```

The least boilerplate-y thing might just be to take the required things as parameters directly.

I'm happy to help with hashing out an alternative to this.

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,855 @@
...
+    error $ "Incorrect usage of DSL; required field not initialized for " <> constr
+  where
+    constr = toText $ tyConName $ typeRepTyCon $ typeRep proxy
+
+fromRight :: Either e a -> a
+fromRight =
+    either (error "fromRight: can't take right side of 'Left' value") id
```
parsonsmatt: I would really prefer to see this as a `MonadFail` rather than a pure exception. There's only one use, and it's to set a backup phrase -- we *really* shouldn't be throwing away Haskell's type safety just because we're writing tests.

test/integration/Test/Integration/Framework/Request.hs
```diff
@@ -0,0 +1,60 @@
...
+import           Universum
+
+import           Data.Generics.Product.Fields (HasField)
+
+import           Cardano.Wallet.Client.Http (ClientError (..), WalletClient)
+import           Pos.Util.Servant (WalletResponse (..))
+import           Test.Integration.Framework.Orphan ()
+
+
+type HasHttpClient ctx = HasField "_client" ctx ctx (WalletClient IO) (WalletClient IO)
```
parsonsmatt: note that `HasType (WalletClient IO) ctx` also works here and is more general :)

parsonsmatt: I'm of the opinion that a class like this could also be a class alias:

```haskell
class (HasType (WalletClient IO) ctx) => HasHttpClient ctx where
    httpClient :: Lens' ctx (WalletClient IO)
    httpClient = typed @(WalletClient IO)

instance (HasType (WalletClient IO) ctx) => HasHttpClient ctx
```

This allows someone to import `HasHttpClient(..)` and get the type *and* lens imported at the same time.

test/integration/Test/Integration/Framework/Request.hs
```diff
@@ -0,0 +1,60 @@
...
+        request >=> \case
+            Left e  ->
+                fail . ("expected a successful response but got an error: " <>) . show $ e
+            Right a ->
+                return a
+
+instance Request (WalletResponse a) where
+    type Response (WalletResponse a) = a
+    request action = do
+        httpClient <- asks (^. #_client)
```
parsonsmatt: note that `view` works in a `MonadReader`, so this is also:


```suggestion
        httpClient <- view #_client
```

test/integration/Test/Integration/Framework/Scenario.hs
```diff
@@ -0,0 +1,60 @@
...
+        s <- get
+        put (f s) *> m <* put s
+
+-- | This gives us the ability to define our spec as `Scenario` instead of just
+-- plain `IO`. This way, each scenario runs within a context and has access to
+-- a wallet client and a dedicated faucet wallet.
+instance Example (Scenario s IO ()) where
+    type Arg (Scenario s IO ()) = MVar s
+    evaluateExample (Scenario io) _ action _ =
+        action runAndPersist >> return (Result "" Success)
```
parsonsmatt, line 41: It seems *really* wrong to me to unconditionally return `Result "" Success`. Don't we care about the result of `action`?

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -0,0 +1,172 @@
...
+            [ expectSuccess
+            ]
+
+    scenario "payment fails when wallet has no funds" $ do
+        (wallet, [addr]) <- setup
+            $ destinations [RandomDestination]
+
+        response <- postTransaction
+            $ source wallet
+            . destination (addr, 14)
```
parsonsmatt: Now that I know these functions are all unsafe, I have an urge to check the definition of every single one and verify that you didn't forget to set the required parameters. It would be *much* easier and simpler if this were simply:

```haskell
response <- postTransaction
    (source wallet)
    (destination (addr, 14))
```

(or whatever the right lensy thing is)

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -0,0 +1,90 @@
+module Test.Integration.Scenario.Wallets
+  ( spec
+  ) where
```
parsonsmatt: 2 space indent

parsonsmatt (changes requested):

test/integration/Test/Integration/Framework/DSL.hs
KtorZ: True :) and to be honest, I thought about it. Though, this is a `before` we always want and can easily forget, hence it being shoved and hidden up here :stuck_out_tongue:

test/integration/Test/Integration/Framework/DSL.hs
KtorZ: Haha, true that :+1:

test/integration/Test/Integration/Framework/DSL.hs
KtorZ: I am rather unsatisfied as well about this one and I knew you'd point that out :upside_down_face:  ... It's really about the end goal in the end: having the end scenario written in the most simple and readable way. Ideally, I'd like to add something at the type-level in the data-structure to mark which fields are required in such way that the whole object becomes a monoid. Fields can be either defaulted to an empty value, or we use this "Required" type-wrapper to initialize the non-defaultable fields. 

```hs
data RequestParameters = RequestParameters
    { myField :: Required String
    , otherField :: [Whatever]
    }
```

Meh, I'll try to get something better :+1:

test/integration/Test/Integration/Framework/Request.hs
KtorZ: mind = blown.

test/integration/Test/Integration/Framework/Scenario.hs
KtorZ: Well, I caught that in the original implementation from HSpec's source code. That's how they implement the `Example (IO ())`. My two cents is that this `Result` wrapper is only used for scenario wrapping QuickCheck properties.

test/integration/Test/Integration/Scenario/Transactions.hs
KtorZ: Well, my only regrets with the fonction vs monoid combinator approach is that, with the function, we are forced to specify every arguments and we loose some cognitive load when reading the test, trying to figure out which of those parameters are we actually acting upon. 
The monoidal combinator approach allows to only use the one that matters for the test. Doesn't buy us much and maybe it's not worth it in the end :/

test/integration/Test/Integration/Framework/Scenario.hs
parsonsmatt: Interesting.

Turns out `hspec` and the underlying libraries just assume an `IO` action completing at all is a success, and all test failures are done by exceptions. So if `action runAndPersist` throws, then `return` never executes!

test/integration/Test/Integration/Framework/DSL.hs
parsonsmatt: There's a concept called Monoid Actions, with a class:

```haskell
class Action m a where
  act :: m -> a -> a
```

where `m` is a Monoid or Semigroup and the laws are basically:

```
act a . act b = act (a <> b)
```

This gives a nice abstraction on "partial configurations acting on a default" where a partial configuration has an "override" `<>` operation.

However, the abstraction ends up being kind-of a lot of boilerplate for not much gain over just being specific about what you mean.

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,656 @@
...
+    -> [a -> m ()]
+    -> m ()
+verify a =
+    mapM_ (a &)
+
+
+--
+-- DEFAULT VALUES
+--
+
```
KtorZ, line 195: All those default are quite "verbose" but they make the scenarios nicer to read in the end :+1:

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,656 @@
...
+noRedemptionMnemonic = Nothing
+
+noSpendingPassword :: Maybe SpendingPassword
+noSpendingPassword = Nothing
+
+
+--
+-- HELPERS
+--
+
```
KtorZ, line 250: @parsonsmatt 
I'd appreciate your inputs on the ones below :upside_down_face: Some are maybe a bit ad-hoc.

test/integration/Test/Integration/Framework/Scenario.hs
KtorZ: TODO :+1:

test/integration/Test/Integration/Scenario/Wallets.hs
KtorZ: TODO :+1:

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,656 @@
...
+--
+-- TYPES
+--
+
+type Getter s a = (a -> Const a a) -> s -> Const a s
+
+lens :: (s -> a) -> ((s, b) -> t) -> Lens s t a b
+lens _get _set = \f x -> curry _set x <$> f (_get x)
+{-# INLINE[0] lens #-}
```
parsonsmatt: why aren't we just using `Control.Lens`? the library already depends on it iirc

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,656 @@
...
+amount
+    :: HasType (V1 Coin) s
+    => Lens' s Word64
+amount =
+    lens _get _set
+  where
+    _get :: HasType (V1 Coin) s => s -> Word64
+    _get = unsafeGetCoin . unV1 . view typed
+    _set :: HasType (V1 Coin) s => (s, Word64) -> s
+    _set (s, v) = set typed (V1 $ mkCoin v) s
```
parsonsmatt, line 260: lgtm

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -0,0 +1,656 @@
...
+    lens _get fst
+  where
+    _get :: HasType Wallet s => s -> WalletId
+    _get = walId . view typed
+
+walletName
+    :: HasType Text s
+    => Lens' s Text
+walletName =
+    typed
```
parsonsmatt, line 316: For definitions as short as these, I'd probably condense to two lines:

```haskell
walletName :: HasType Text s => Lens' s Text
```

Note, however, that `walletName` will only work on data structures that only contain a *single* value of type `Text`, and will work on values like:

```haskell
data User = User { userName :: Text } deriving Generic
```

which is not what you want. So a `newtype WalletName = WalletName Text` might be warranted here?

test/integration/Test/Integration/Framework/Request.hs
parsonsmatt: The metadata is mostly useful for pagination

parsonsmatt (approved):

test/integration/Test/Integration/Framework/DSL.hs
KtorZ: Yeah, that was to have the definition in front of me while I was grooming haha. This is the kind of polish i told you about :upside_down_face:

test/integration/Test/Integration/Framework/DSL.hs
KtorZ: Yup, thought of it and we'll probably end up doing that. I didn't want to clutter the PR with type changes in our source code but well, that's next for sure (including removing all those V1)

PR #129, opened 2018-12-04
Address Pool Module tests (WIP!).

denisshevchenko:
# Linked Issue

<p align="right">#33</p>

# Overview

1. - [x] `AddressPoolGap` functions are covered by tests.
2. - [x] `AddressPool` functions are covered by tests.

# Comments

¯\\\_(ツ)\_/¯

# Checklist

---

### DIFF
--- a/src/Cardano/Wallet/Kernel/AddressPool.hs
+++ b/src/Cardano/Wallet/Kernel/AddressPool.hs
@@ -18,6 +18,7 @@ module Cardano.Wallet.Kernel.AddressPool
     ( AddressPool
     , initAddressPool
     , lookupAddressPool
+    , getAddressPoolSize
     ) where
 
 import           Universum
@@ -102,6 +103,10 @@ lookupAddressPool target pool =
         Nothing ->
             (Nothing, pool)
 
+-- | For testing purpose.
+getAddressPoolSize
+    :: AddressPool address -> Int
+getAddressPoolSize pool = Map.size $ pool ^. addresses
 
 {-------------------------------------------------------------------------------
   Internals
--- a/src/Cardano/Wallet/Kernel/AddressPool.hs
+++ b/src/Cardano/Wallet/Kernel/AddressPool.hs
@@ -19,6 +19,7 @@ module Cardano.Wallet.Kernel.AddressPool
     , initAddressPool
     , lookupAddressPool
     , getAddressPoolSize
+    , getAddressPoolGap
     ) where
 
 import           Universum
@@ -103,10 +104,15 @@ lookupAddressPool target pool =
         Nothing ->
             (Nothing, pool)
 
--- | For testing purpose.
+-- | Get the underlying pool's size
 getAddressPoolSize
     :: AddressPool address -> Int
 getAddressPoolSize pool = Map.size $ pool ^. addresses
+
+-- | Get the underlying pool's gap
+getAddressPoolGap
+    :: AddressPool address -> AddressPoolGap
+getAddressPoolGap pool = pool ^. gap
 
 {-------------------------------------------------------------------------------
   Internals
--- a/test/unit/Test/Spec/AddressPool.hs
+++ b/test/unit/Test/Spec/AddressPool.hs
@@ -1,69 +1,38 @@
-{-# LANGUAGE DataKinds        #-}
-{-# LANGUAGE LambdaCase       #-}
-{-# LANGUAGE TypeApplications #-}
-
 module Test.Spec.AddressPool (spec) where
 
 import           Universum
 
-import           Cardano.Wallet.Kernel.AddressPool (getAddressPoolSize,
-                     initAddressPool, lookupAddressPool)
-import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap,
-                     mkAddressPoolGap)
+import           Cardano.Wallet.Kernel.AddressPool (AddressPool,
+                     getAddressPoolGap, getAddressPoolSize, initAddressPool,
+                     lookupAddressPool)
+import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 
 import           Test.Hspec (Spec, describe, it)
 import           Test.Pos.Core.Arbitrary ()
-import           Test.QuickCheck (Property, property, (===))
-
--- | Since the pool is polymorphic by address type, we can use any type, even 'Word'.
-mkRandomALaAddress :: Word -> Word
-mkRandomALaAddress ix = ix
-
-smallestGap, biggestGap :: AddressPoolGap
-smallestGap = minBound
-biggestGap  = maxBound
-
--- | It proves that we:
--- 1. can create 'AddressPoolGap' from any number within the min-max range,
--- 2. cannot create 'AddressPoolGap' from any other number.
-prop_gapGeneratorIsValid
-    :: Int
-    -> Property
-prop_gapGeneratorIsValid gap =
-    case mkAddressPoolGap (fromIntegral gap) of
-        Left _  -> isWithinBound gap === False
-        Right _ -> isWithinBound gap === True
-  where
-    isWithinBound g =
-        g >= fromIntegral smallestGap &&
-        g <= fromIntegral biggestGap
-
--- | It proves that pool extension works as expected.
-prop_poolExtensionIsValid
-    :: AddressPoolGap
-    -> Word
-    -> Property
-prop_poolExtensionIsValid validGap aLaAddress = do
-    let pool = initAddressPool validGap mkRandomALaAddress
-        (result, pool') = lookupAddressPool aLaAddress pool
-    property $ case result of
-        Nothing -> getAddressPoolSize pool == getAddressPoolSize pool'
-        Just (foundAddr, foundAddrIx) -> do
-            -- Since we found this address in the pool, it could be extended.
-            if getAddressPoolSize pool == getAddressPoolSize pool'
-                then aLaAddress == foundAddr
-                else aLaAddress == foundAddr &&
-                        -- Pool was extended, so found address was "on edge":
-                        -- new addresses was added in it so number of "consecutive
-                        -- undiscovered addresses" is equal to 'validGap'.
-                        getAddressPoolSize pool' - (fromIntegral foundAddrIx + 1) == fromIntegral validGap
+import           Test.QuickCheck (Property, label, property, (.&&.), (===))
 
 spec :: Spec
 spec = describe "AddressPool" $ do
-    describe "Gap" $
-        it "gap generator works as expected" $
-            property prop_gapGeneratorIsValid
+    it "lookupAddressPool" $ property prop_lookupAddressPool
 
-    describe "Pool" $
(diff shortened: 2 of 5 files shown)


src/Cardano/Wallet/Kernel/AddressPool.hs
```diff
@@ -18,6 +18,8 @@ module Cardano.Wallet.Kernel.AddressPool
     ( AddressPool
     , initAddressPool
     , lookupAddressPool
+      -- ** For testing purpose.
```
KtorZ: Not necessarily. Let's remove the comment :+1:

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,81 @@
...
+biggestGap  = maxBound
+
+spec :: Spec
+spec = describe "AddressPool" $ do
+    describe "Gap" $ do
+        it "fails if we try to make too small or too big gap" $ do
+            let tooSmallGap = mkAddressPoolGap $ fromIntegral smallestGap - 1
+                tooBigGap   = mkAddressPoolGap $ fromIntegral biggestGap + 1
+            tooSmallGap `shouldSatisfy` isLeft
+            tooBigGap   `shouldSatisfy` isLeft
```
KtorZ: This isn't really testing 'AddressPool', is it ? Rather, this is testing `mkAddressPoolGap` constructor and our `Bounded` instance.

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,81 @@
...
+
+        it "works as expected if we make a gap from some valid number" $ do
+            validGapNum <- liftIO $ randomRIO (fromIntegral smallestGap, fromIntegral biggestGap)
+            let validGap = mkAddressPoolGap validGapNum
+            validGap `shouldSatisfy` isRight
+            let Right gap = validGap
+            fromIntegral gap `shouldBe` validGapNum
```
KtorZ: The two tests above could actually be generated! That's the whole point of property testing after all!
For instance:

```hs
prop_generatorIsValid
    :: Int
    -> Property
prop_generatorIsValid gap =
    case mkAddressPoolGap (fromIntegral gap) of
        Left _  -> not (isWithinBound gap)
        Right _ -> isWithinBound gap
  where
    maxGap = fromIntegral (maxBound :: AddressPoolGap)
    minGap = fromIntegral (minBound :: AddressPoolGap)
    isWithinBound x = x <= maxGap && x >= minGap
```

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,81 @@
...
+            pool == samePool `shouldBe` True
+            addressInfo `shouldSatisfy` isNothing
+
+        it "works as expected if we lookup existing address in the pool" $ do
+            let Right addr = decodeTextAddress "DdzFFzCqrhsuR7de9K8yPa1piWnxux2fcEmU6chozrzb1FUDeMZxsQtB9nkta4huQ4yusHcsAY7r6wBvcsa4ZibZ9FSLfme3CsYWm5KB"
+                pool = initAddressPool smallestGap mkAddress
+                (addressInfo, _pool) = lookupAddressPool addr pool
+            addressInfo `shouldSatisfy` isJust
+            let Just (sameAddress, _addrIx) = addressInfo
+            addr `shouldBe` sameAddress
```
KtorZ: Here, we could use instead a small state-machine testing comparing about pool against a given model. Here, our model can just be the pool's size :stuck_out_tongue:

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,81 @@
...
+    , "DdzFFzCqrhsnK2qrKkgwchSDbCkycDnEnxrZ5Bt9n63fifgGcoWpomC1m7VFU5FAaB7LMwVSCRNN9wLvVbgj9Ud9A3zh1pCfxZqALAQE"
+    , "DdzFFzCqrhsuR7de9K8yPa1piWnxux2fcEmU6chozrzb1FUDeMZxsQtB9nkta4huQ4yusHcsAY7r6wBvcsa4ZibZ9FSLfme3CsYWm5KB"
+    , "DdzFFzCqrhsgMMHi3HkGusygzb3ntLyhCyG7cE7Q82DvjT2eeznVcNqhKLkkbzjn6QeXJ2HncaQv1C3893QmiihYtQNbEk1ms9f7cADs"
+    , "DdzFFzCqrht4YtuHPRbfzyADibMC7k6L4NFuDbpRAfN34u1cqDwpDJP3oFmoRfqSSNG8Qd3v6BfCwbkM9wPkAYQrAGiwTMWjmd3sDhZy"
+    , "DdzFFzCqrht2jmfwryBNKvaypAWYypBJ2j9c8bKYSSxrV1ihC5PjWWvEZhzCztKVj7gpNRX6rfNvtydi8oW8vf1hpx9vPeMXJQcasRHc"
+    , "DdzFFzCqrhshmzJsq1NJwZn9gv7vMr8HRokKuYURCjeakWJAJbTv4dJf2dGxz8kTHVwg5ntX6wDxeNxJD7fDWFXJ1P7eUFHf36HgauEF"
+    , "DdzFFzCqrhstjjqn1E1brKmwNQnrLeXx9r9DKiR2YCoGM7xJvF1W2x5Sq6TesCtNjtrqpJi6BUEHh58bkF1C2gP33Tk147Lx3QENhnAE"
+    , "DdzFFzCqrhsmyUMwRv8Dr1Q6NgbU57bcGjXdiMLv4NFqUEAfzwVNg2Qw1jKv6LJAovzDT8CXapE9NKR7YUBZnXrZPJN43HXoYGGJX6fF"
+    , "DdzFFzCqrht9FGQY5zJZnhnDwvra1TwmSJ9kZYzXut7TNwsy6EfAAWBJy7SoYpb2KAYtpppeHiiFk1Crdvo8Wi4GvRXTxyfgjoWp4zcP"
+    ]
```
KtorZ: Don't we have arbitrary instances for those :smile: ?

test/unit/Test/Spec/AddressPool.hs
denisshevchenko: Yes, this test proves that it's _impossible_ to create too small or too big gap. It's a useful test, right?

test/unit/Test/Spec/AddressPool.hs
denisshevchenko: Well, no:

> No instance for (Test.QuickCheck.Arbitrary.Arbitrary Address)

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,100 @@
...
+-- | It proves that we can create a valid pool from any valid gap.
+prop_poolInitializerIsValid
+    :: Int
+    -> Property
+prop_poolInitializerIsValid gap =
+    case mkAddressPoolGap (fromIntegral gap) of
+        Left _ -> True === True -- This gap is wrong, just ignore it.
+        Right validGap -> do
+            let poolWithRandomAddresses = initAddressPool validGap mkRandomAddress
+            getAddressPoolSize poolWithRandomAddresses === fromIntegral validGap
```
KtorZ: This property mixes two things:

- The validity test of `initAddressPool`
- The creation of `AddressPoolGap`

Plus.. I am not really sure what are we testing here? We can't create address pools from invalid address gap, that's a guarantee from the type system already. `initAddressPool` doesn't use a raw `Int`, it uses a `AddressPoolGap`, and since the only way to get a instance of such type is through `mkAddressPoolGap` and we already ensured that this was correct, there's no reason to test that further...

denisshevchenko: You're right, this test is unnecessary.

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,95 @@
...
+import           System.IO.Unsafe (unsafePerformIO)
+import           System.Random (Random (..))
+
+-- | Generate some random address.
+mkRandomAddress :: Word -> Address
+mkRandomAddress _ix = unsafePerformIO $ generate arbitrary
+
+-- | Generate some random UUID.
+mkRandomUUID :: Word -> UUID
+mkRandomUUID _ix = unsafePerformIO $ nextRandom
```
KtorZ: This generates *exactly one* value, once. `unsafePerformIO` is really not what you're looking for here. 
Quickcheck generates those values for you through `Arbitrary` instances ... This should be a parameter of your property, that's really what Quickcheck is for in the end....

KtorZ (changes requested):

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,95 @@
...
+    :: Int
+    -> Property
+prop_poolExtensionIsValid gap =
+    case mkAddressPoolGap (fromIntegral gap) of
+        Left _ -> property True -- Gap is invalid, just ignore it.
+        Right validGap -> do
+            let pool = initAddressPool validGap mkRandomUUID
+                allAddresses = getAddresses pool
+            -- Lookup in the pool.
+            monadicIO $ do
```
KtorZ: There's nothing monadic in this property to be honest .... This is all pure stuff.

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,95 @@
...
+        g >= fromIntegral smallestGap &&
+        g <= fromIntegral biggestGap
+
+-- | It proves that pool extension works as expected.
+prop_poolExtensionIsValid
+    :: Int
+    -> Property
+prop_poolExtensionIsValid gap =
+    case mkAddressPoolGap (fromIntegral gap) of
+        Left _ -> property True -- Gap is invalid, just ignore it.
```
KtorZ: Then, make sure to get one as parameter instead...

```hs
prop_poolExtensionIsValid
    :: AddressPoolGap
    -> Property
prop_poolExtensionIsValid gap =
```

Again, that's the whole point of defining `Arbitrary` instances.

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,95 @@
...
+        Right validGap -> do
+            let pool = initAddressPool validGap mkRandomUUID
+                allAddresses = getAddresses pool
```
KtorZ: Nope, this we don't want to expose. `getAddresses` shouldn't exist.

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -0,0 +1,95 @@
...
+            -- Lookup in the pool.
+            monadicIO $ do
+                -- If all addresses in the pool are really unique -
+                -- pool size is equal to 'validGap' for now.
+                assert $ getAddressPoolSize pool == fromIntegral validGap
+                someIndex <- run (liftIO $ randomRIO (0, length allAddresses - 1))
```
KtorZ: This should be a property parameter and not generated as part of the property.

KtorZ (approved):

Issue #31, opened 2018-11-07
BIP-44 Utils Part I: Module Skeleton & Public Derivations


KtorZ:
# Context

<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

BIP-0044 Multi-Account Hierarchy for Deterministic Wallets is a Bitcoin standards defining a structure and algorithm to build a hieararchy tree of keys from a single root private key.

It it built upon [BIP-0032](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) and is a direct application of [BIP-0043](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki).  It defines a common representation of addresses as a multi-level tree of derivations:

```
m / purpose' / coin_type' / account' / change / address_index
```

Paths with a tild `'` refer to BIP-0032 hardened derivation path (meaning that the private key and passphrase are required to derive children). This leads to a couple of interesting properties:

1. New addresses (change or not) can be generated from account's public keys only.
1. The derivation of addresses is done sequentially / deterministically. 
1. If an account private key is compromised, it doesn't compromise other
   accounts. 

This allows for external keystores and off-loading of key derivation to some external source such that a wallet could be tracking a set of accounts without the need for knowing private keys. This approach is discussed more in details below. 

> NOTE:  
> One other important aspect is more of a security-concern. The introduction of
> such new address scheme makes it possible to change the underlying derivation
> function to a new better one with stronger cryptographic properties. This is
> rather ad-hoc to the structural considerations above, but is still a major
> motivation.

See also: https://github.com/input-output-hk/cardano-wallet/wiki/Address-Derivation-a-la-BIP---44#ed25519-bip-44

# Decision

<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

Instead of mixing up the sequential derivation with new scheme in the existing `deriveLvl2KeyPair` and related. We'll instead create and test in isolation a BIP-44 modules with a few helpers to achieve key derivation. 
Note that initially, there's no need to support derivation related to private keys, this is out-of-scope for E.O.S wallets.

# Acceptance Criteria

<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

[...]

PR #133, opened 2018-12-06
BIP-44 Utils, public derivations (WIP!).

denisshevchenko:
# Linked Issue

<p align="right">#31</p>

# Overview

1. - [x] Function for Lvl3-derivation is provided.

# Comments

¯\\\_(ツ)\_/¯

# Checklist

---

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -61,12 +61,10 @@ module Cardano.Wallet.API.V1.Types (
   -- * Addresses
   , WalletAddress (..)
   , NewAddress (..)
-  , AddressPath
   , AddressLevel
   , addressLevelToWord32
   , word32ToAddressLevel
   , IsChangeAddress (..)
-  , mkAddressPathBIP44
   -- * Payments
   , Payment (..)
   , PaymentSource (..)
@@ -1373,81 +1371,6 @@ instance FromJSON AddressLevel where
 
 newtype IsChangeAddress = IsChangeAddress Bool deriving (Show, Eq)
 
--- | BIP44 derivation path, for work with external wallets, for example:
--- m / purpose' / coin_type' / account' / change / address_index
--- m /      44' /      1815' /       0' /      0 /             1
---
--- NOTE See:
---   - https://github.com/satoshilabs/slips/blob/master/slip-0044.md
---   - https://github.com/satoshilabs/slips/pull/123
-data AddressPath = AddressPath
-    { addrpathPurpose      :: AddressLevel
-    , addrpathCoinType     :: AddressLevel
-    , addrpathAccount      :: AddressLevel
-    , addrpathChange       :: AddressLevel
-    , addrpathAddressIndex :: AddressLevel
-    } deriving (Show, Eq, Generic)
-
-deriveJSON Aeson.defaultOptions ''AddressPath
-
-instance ToSchema AddressPath where
-    declareNamedSchema =
-        genericSchemaDroppingPrefix "addrpath" (\(--^) props -> props
-            & ("purpose"      --^ "44, refers to BIP-44.")
-            & ("coinType"     --^ "1815 for ADA (Ada Lovelace's birthdate).")
-            & ("account"      --^ "Account index, used as child index in BIP-32 derivation.")
-            & ("change"       --^ "0 if external (e.g. payment addr), 1 if internal (e.g. change addr).")
-            & ("addressIndex" --^ "Address index counter, incremental.")
-        )
-
-instance Arbitrary AddressPath where
-    arbitrary = AddressPath <$> arbitrary
-                            <*> arbitrary
-                            <*> arbitrary
-                            <*> arbitrary
-                            <*> arbitrary
-
-deriveSafeBuildable ''AddressPath
-instance BuildableSafeGen AddressPath where
-    buildSafeGen sl AddressPath{..} = bprint (""
-        %"{ purpose="%buildSafe sl
-        %", coin_type="%buildSafe sl
-        %", account="%buildSafe sl
-        %", change="%buildSafe sl
-        %", address_index="%buildSafe sl
-        %" }")
-        addrpathPurpose
-        addrpathCoinType
-        addrpathAccount
-        addrpathChange
-        addrpathAddressIndex
-
--- Smart constructor to create BIP44 derivation paths
---
--- NOTE Our account indexes are already referred to as "hardened" and
--- are therefore referred to as indexes above 2^31 == maxBound `div` 2 + 1
--- AddressPath makes it explicit whether a path should be hardened or not
--- instead of relying on a convention on number.
-mkAddressPathBIP44
-    :: IsChangeAddress   -- ^ Whether this is an internal (for change) or external address (for payments)
-    -> Account           -- ^ Underlying account
-    -> Maybe AddressPath -- ^ If 'Nothing' - address level is out-of-bound (31-byte unsigned integer).
-mkAddressPathBIP44 (IsChangeAddress isChange) acct = AddressPath
-    <$> pure (AddressLevelHardened 44)
-    <*> pure (AddressLevelHardened 1815)
-    <*> pure (word32ToAddressLevel actualIndex)
-    <*> pure (AddressLevelNormal $ if isChange then 1 else 0)
-    <*> checkWord31 AddressLevelNormal (getAddressIndex acct)
-  where
-    getAddressIndex = fromIntegral . length . accAddresses
-    checkWord31 mkLevel n =
-        if n >= hardenedValue then
-            Nothing
-        else
-            Just (mkLevel n)
-    AccountIndex actualIndex = accIndex acct
-
-
 -- | A type incapsulating a password update request.
 data PasswordUpdate = PasswordUpdate {
     pwdOld :: !SpendingPassword
@@ -2200,7 +2123,6 @@ instance Example AccountIndex
 instance Example AccountBalance
 instance Example AccountAddresses
 instance Example AddressLevel
-instance Example AddressPath
 instance Example AddressPoolGap
 instance Example WalletId
 instance Example AssuranceLevel
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -1,0 +1,26 @@
+{-------------------------------------------------------------------------------
+
+  This module provides utils to perform account public key child derivation:
+
+    account (parent) extended public key -> address (child) extended public key
+
+-------------------------------------------------------------------------------}
+
+module Cardano.Wallet.Kernel.Ed25519Bip44 (
+    deriveLvl3PublicKey
+    ) where
+
+import           Universum
+
+import           Pos.Crypto (PublicKey (..))
+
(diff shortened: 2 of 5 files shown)


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1370,81 +1368,6 @@ instance FromJSON AddressLevel where
...
-  where
-    getAddressIndex = fromIntegral . length . accAddresses
-    checkWord31 mkLevel n =
-        if n >= hardenedValue then
-            Nothing
-        else
-            Just (mkLevel n)
-    AccountIndex actualIndex = accIndex acct
-
-
```
KtorZ, line 1450: :+1:

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -0,0 +1,35 @@
...
+    maybe (error "Unable to derive address public key from account public key!")
+          id
+          $ deriveHDPublicKeyScheme2 accountPublicKey (fromIntegral addressIx)
+
+-- | Derives extended public key from extended public key using derivation scheme 2
+-- (please see 'cardano-crypto' package).
+deriveHDPublicKeyScheme2 :: PublicKey -> Word32 -> Maybe PublicKey
+deriveHDPublicKeyScheme2 (PublicKey xpub) childIndex =
+    maybe Nothing
+          (Just . PublicKey)
```
KtorZ: `fmap` ?

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -0,0 +1,35 @@
...
+
+import           Cardano.Crypto.Wallet (DerivationScheme (..), deriveXPub)
+
+-- | Derives address public key from the given account public key.
+deriveLvl3PublicKey
+    :: PublicKey
+    -> Word
+    -> PublicKey
+deriveLvl3PublicKey accountPublicKey addressIx =
+    maybe (error "Unable to derive address public key from account public key!")
```
KtorZ: Let's not throw in the wild like this. Especially in this module. Leave that to the caller.

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -0,0 +1,35 @@
...
+          id
+          $ deriveHDPublicKeyScheme2 accountPublicKey (fromIntegral addressIx)
+
+-- | Derives extended public key from extended public key using derivation scheme 2
+-- (please see 'cardano-crypto' package).
+deriveHDPublicKeyScheme2 :: PublicKey -> Word32 -> Maybe PublicKey
+deriveHDPublicKeyScheme2 (PublicKey xpub) childIndex =
```
KtorZ: No need for this indirection, let just inline this function in the call above.

KtorZ (approved):

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -0,0 +1,26 @@
...
+
+import           Universum
+
+import           Pos.Crypto (PublicKey (..))
+
+import           Cardano.Crypto.Wallet (DerivationScheme (..), deriveXPub)
+
+-- | Derives address public key from the given account public key,
+-- using derivation scheme 2 (please see 'cardano-crypto' package).
+deriveLvl3PublicKey
```
akegalj, line 21: @denisshevchenko @KtorZ 
Hm, I don't think this is right. If we are about to implement bip44 - derivation path is defined as
```
m / purpose' / coin_type' / account' / change / address_index
```
if we are providing account public key here, we still have to first derive internal/external change derivation before deriving an address.

Or we decided not to bother with change derivation in our implementation?

I am not sure do we have a notion of internal/external wallet atm (v0 had this adhoc) 

Also, if we are implementing bip44 , why not sticking with proposed naming convention:

m => level 0
purpose' => level 1
coin_type' => level 2
account' => level 3
change => level 4
address_index => level 5 (instead of level 3)

KtorZ: We don't manipulate derivation path so to speak. `cardano-crypto` does it. We only use the abstract type known as `PublicKey` / `PrivateKey`. The rest is irrelevant to us as we aren't implementing the cryptographic details of it.

akegalj: its a shame wallet can't fetch change derivation index as this would be useful information (to show an address or to hide it; it would also be useful for explorer to hide change addresses).

NicolasDP: @KtorZ `cardano-crypto` does not implement the Bip44 scheme. only the ed25519-bip32 derivation scheme (which is a cryptographic primitive). You have to implement the bip44 scheme on your side of the wallet (i.e. hardened derivation for purpose and coin_type and account, and then soft derivation for change and address).

so like @akegalj said, if you have the account public key, you need to derive your internal (or external) first and then derive the address key. 2 derivation levels.

There is no understanding of path in the cardano-crypto library, only the derivation index from one public key to another. so you need to implement every derivation levels of the BIP44 on your side.

KtorZ: Indeed. Bad choice of wording on my side. What I meant was that, there's no need to define "paths" on our side and keep track of those, so to speak. Paths are just another round of derivation. We just need to provide utils that perform the correct number of derivations for the right type given the corresponding address or account indexes (for intermediate derivation like purpose and coin_type, values are just hardcoded). In the end, we only manipulate `XPub` and `XPrv` keys.

From an external user, this looks roughly like:

```hs
deriveAccountPrv :: (XPrv, PassPhrase) -> Word32 -> XPrv
   -- perform triple derivation from the given wallet XPrv 
deriveAddressPub :: XPub -> IsChange -> Word32 -> XPub
   -- perform double derivation from the given account XPub 
```

akegalj: yes - seems like there was a noise in communication. Thanks for having a look @NicolasDP

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -0,0 +1,36 @@
...
+    -> Property
+prop_cannotDerivelvl3PublicKeyForBigIx accountPublicKey addressIx = property $
+    if addressIx >= maxIx
+        then isNothing result
+        else isJust result
+  where
+    result = deriveLvl3PublicKey accountPublicKey addressIx
+    -- This is maximum value for soft derivation (only soft derivation
+    -- is allowed to derive public key from public key).
+    maxIx = 0x8000000
```
akegalj, line 31: @denisshevchenko I guess you missed one zero here. It should be 0x80000000 (seven zeros).

(the test is failing then :D)

akegalj: but it seems to be a bug in cardano-crypto, not in your code (or I am not understanding bip32 well) - will have a closer look at it

akegalj: funny how things turn out. 

Seems like Denis accidentally dropped one zero from the test here https://github.com/input-output-hk/cardano-wallet/pull/133#discussion_r243491986 and this property was looking ok. And I also haven't noticed the mistake until I refactored his `maxIx` hardcoded value to use `isHardened` (from cardano-sl, defined as `(>=2^31)`. And lost a bunch of time to figure out the typo mistake (I though I had problems with `stack`s build/caching/linking mechanism).

Then I noticed failing test:
```
  1) Ed25519Bip44, Deriving address public key, fails if address index is too big
       Falsifiable (after 57 tests and 11 shrinks):
         PublicKey (XPub {xpubPublicKey = ")\129\177\221\SUB8\\\t)\"\ESC\DEL\128\&0\154\157\144\201\192O\249\144\203\167\247\131\146\225\237\249\133E", xpubChaincode = ChainCode "9\202(19A2#)\217\189\237\231\NUL\r\\]\160\CAN\195\144\195\196^\192m\181\220\225\155^+"})
         ExternalChain
         134217728
```
contains address index  134217728 , and this index was always popping out from quickcheck (no matter the seed - shrinking got it to here). it turns out this is 2^27 . And quick check for cardano-crypto revealed `deriveXPub` is failing at this index:
```
> PublicKey ppk <- Test.QuickCheck.generate (arbitrary :: Gen PublicKey)
> deriveXPub DerivationScheme2 ppk 0                         
Just (XPub {xpubPublicKey = "\143\&6v\159\SO\220K\ETB\rD2P\227\193\t\194_l\159\144\&7\149\USb\ETXT\DC2E\EM\r\188n", xpubChaincode = ChainCode "M?\245(\254\231\155\&9L\238\254\148b{}q\SUB\201\a\185E\165\176v\242\231\RS\184j\196\198["})
> deriveXPub DerivationScheme2 ppk 134217727                 
Just (XPub {xpubPublicKey = "d\253c\199\180\128\180\201\162\198\153\US\180c\SIq\248abfH1x0\253\245R= \195\211\224", xpubChaincode = ChainCode "\135l3\216\133\236\ETX\254\158\197:(\153\NAK\215\191\183P\246\171\167\210\&4C\ESC\215\188\182\173\SO\250\165"})
> deriveXPub DerivationScheme2 ppk 134217728
Nothing
```

it might be the problem in how `PublicKey` was generated - will have a closer look at that before pinging our crypto team

(accidentily this seems to be the same problem catched by property https://github.com/input-output-hk/cardano-wallet/pull/169#discussion_r242691873 https://github.com/input-output-hk/cardano-wallet/pull/169#discussion_r242741196)

akegalj: When single zero is dropped as @denisshevchenko did it, it is actually the same boundery where `deriveXPub` is failing:
```
> logBase 2 0x8000000
27.0
```

so either both @denisshevchenko and `deriveXPub` are correct and I am missing some crutial understanding of bip32 https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys or when 0 zero is added we get:
```
> logBase 2 0x80000000
31.000000000000004
```

which is actually boundary defined for hardened keys https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys

It seems very unlikely that both Denis and crypto is wrong. It is also unlikely that denis accidentily dropped one zero and this is exact boundary where crypto guys made a mistake. Looks more likely that I am missing some info?

akegalj: bip32 https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#public-parent-key--public-child-key also doesn't reveal why `deriveXPub` should work only for `<2^27` . Having a look at `cardano-crypto` internals a bit...

Seems like the same mistake with dropped zero is there as well https://github.com/input-output-hk/cardano-crypto/blob/master/src/Cardano/Crypto/Wallet.hs#L169 :
```haskell
deriveXPub ds (XPub pub (ChainCode cc)) n
    | n >= 0x8000000 = Nothing
```

using `0x8000000` == 2^27 , instead of `0x80000000` == 2^31 .

I wonder why this is hardcoded and not reused as `git grep 0x8000000` in cardano-crypto reveals multiple usages.

akegalj: It should be fixed here https://github.com/input-output-hk/cardano-crypto/pull/52

Issue #119, opened 2018-11-29
(CI) Remove use of `V1` wrapper. 


KtorZ:
# Context

<WHEN PROPOSED>
<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

We historically defined a newtype `V1` to help with maintaining multiple versions of our API types. Truth is that this has now gone out of hand and clutters the code more than it helps it. 
We use it rather randomly, and mostly as a quick wrapper around already existing types we import from core or networking to work around orphan instances.

e.g.

- https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/API/V1/Types.hs#L288

- https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/API/V1/Types.hs#L321

In most (all) cases, it would make more sense to actually define a dedicated newtype, for instance `WalletTimestamp` or simply `Timestamp` in a dedicated module and have abstract away the dependency to core types. 

In addition, this is slightly erroneous since the API is now, technically, at V2.

# Decision

<WHEN PROPOSED>
<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

In a first step towards cleaning up our code, we can remove `V1` wrappers, and replace them, when appropriated, with dedicated newtypes. When not appropriated, we'll simply use the raw type.

# Acceptance Criteria

<WHEN PROPOSED>
<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

1. - [x] `V1` _mustn't_ be used throughout the code base
2. - [x] API type representation _must_ remain identical without `V1` than what they are with `V1`

---

# Development Plan

- [x] Change V1 wrapper types with dedicated new types
    - [x] PassPhrase
    - [x] Timestamp
    - [x] InputSelectionPolicy
    - [x] Address
    - [x] AddressOwnership
    - [x] Coin
    - [x] TxId
    - [x] SoftwareVersion
    - [x] Version
- [x] Cleanup `V1` from existing code

Please note that there're _two_ modules in `cardano-wallet` where we still use `V1` and `unV1`, but it's imported explicitly from `Pos.Node.API`.

[...]

PR #135, opened 2018-12-06
Remove use of V1 wrapper: PassPhrase

denisshevchenko:
# Linked Issue

<p align="right">#119</p>

# Overview

1. - [x] Type `Core.PassPhrase` doesn't use `V1` wrapper.
2. - [x] newtype `WalletPassPhrase` is introduced instead.

# Comments

¯\\\_(ツ)\_/¯

# Checklist

---

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -40,6 +40,7 @@ module Cardano.Wallet.API.V1.Types (
   , WalletOperation (..)
   , SpendingPassword
   , mkSpendingPassword
+  , WalletPassPhrase (..)
   , EosWallet (..)
   , NewEosWallet (..)
   , WalletAndTxHistory (..)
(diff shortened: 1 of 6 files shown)


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -237,21 +236,18 @@ mkPassPhrase text =
...
         Left e    -> fail (toString e)
-        Right pp' -> pure (V1 pp')
+        Right pp' -> pure pp'
     parseJSON x           = typeMismatch "parseJSON failed for PassPhrase" x
 
-instance Arbitrary (V1 Core.PassPhrase) where
-    arbitrary = fmap V1 arbitrary
-
-instance ToSchema (V1 Core.PassPhrase) where
+instance ToSchema Core.PassPhrase where
```
KtorZ: Here we introduce orphan instances, which we really don't want. Instead, let's come up with our own dedicated newtype: `PassPhrase` or `WalletPassPhrase`. The idea isn't to simply remove `V1` but to replace it with appropriated newtypes!

akegalj (approved):

KtorZ (changes requested):
PR looks good, but we're missing two things:

- Golden tests
- Acid-State Migration

Ideally, we need write golden tests before making any change, such that they are written using the old code, and we control via the new code that everything is correct. 
Can you maybe issue another PR that we'll merge prior to this one, only adding at least one golden for the `ToJSON` and `FromJSON` instance, even though I am quite sure nothing got broken along the way.

But for all others, that should become the normal procedure:

1/ Golden tests
2/ Removing / Replacing V1
3/ Acid-state Migration.

KtorZ:
@denisshevchenko 

Remember to update golden tests here as a consequence of #144  :+1:

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -343,20 +338,30 @@ instance ToSchema (V1 Core.Timestamp) where
...
 
-instance Semigroup (V1 Core.PassPhrase) where
-    V1 a <> V1 b = V1 (a <> b)
+instance Semigroup WalletPassPhrase where
+    WalletPassPhrase a <> WalletPassPhrase b = WalletPassPhrase (a <> b)
 
-instance Monoid (V1 Core.PassPhrase) where
-    mempty = V1 mempty
+instance Monoid WalletPassPhrase where
+    mempty = WalletPassPhrase mempty
```
KtorZ, line 347: Let's avoid defining instances on type aliases. Could we actually remove the "SpendingPassword" alias entirely (while preserving the existing fields in various data types?)

KtorZ (approved):
LGTM :+1:

One minor gotcha that we could fix as part of another PR I believe.

PR #136, opened 2018-12-07
Fix @akegalj return date from holidays

akegalj:


KtorZ (approved):
Ho right ... :( Looked at the wrong week.

Issue #34, opened 2018-11-07
Review Prefiltering 


KtorZ:
# Context

With #33, we should have the necessary tools to perform account discovery while
applying blocks. Answering the question `isOurs` happens during pre-filtering
where we skim through incoming blocks, looking for inputs that are related to
our wallet(s).

# Decision

Using the address pool module, we should be able to abstract away the account
discovery to use the relevant method depending on whether we have a F.O. wallet
or an E.O. one.

# Acceptance Criterias

1. Addresses derived sequentially _can_ be identified during pre-filtering
1. Addresses derived randomly _must_ still be identifiable during pre-filtering
1. Pre-filtering _may_ make use of the module from #33.1

---

# Development Plan

- [x] Abstract over wallet/account/address ids
- [x] Parametrise prefiltering code with IsOurs + PrefilterKey, separate HdRnd from generic prefiltering module
- [x] Review pre-filtering implementation all-in-all
    - [x] Review definition of "PrefilteredBlock" and remove addresses from it
    - [x] Unify "AddrWithId" and "HdAddress" types
    - [x] Rewrite pre-filtering for single accounts, without considering metadata (yet).
    - [x] Introduce multi-accounts pre-filtering to avoid traversing a block n times (n = number of accounts)
    - [x] Compute metadata on the prefiltered result (instead of during prefiltering)
    - [x] Update usage of prefiltering functions in BListener and Restoration
- [ ] ~~Instantiate prefiltering with sequential derivation~~
- [x] Review corresponding unit tests to work with either sequential or random derivation (and test both)

# PR

| Number | Base      |
| ---    | ---       |
| #139   | `develop` |
| #179 | `develop` |
| #180 | `develop` |
| #181 | `develop` |
| #183 | `develop` |
| #192 | `develop` |
| #204 | `develop` |

# QA

### IsOurs

This was a big one but has finally landed. The _major_ and most important change is actually fairly simple but allows to abstract over the pre-filtering (resp. recovery) process by having the implementation working with a generic "IsOurs" interface.

https://github.com/input-output-hk/cardano-wallet/blob/e1526cbc931af11d4d1ff05c8563813a4772dd22/src/Cardano/Wallet/Kernel/DB/HdWallet.hs#L214-L225

which goes together with:

[...]

PR #139, opened 2018-12-07
[BIP44-D-#34] Part 1 - Abstract ids for Wallet/Account/Address in pre…

uroboros:
…filtering module

# Linked Issue

https://github.com/input-output-hk/cardano-wallet/issues/34

<p align="right">#</p>

# Overview

## Commit 1 - Abstract Ids in Prefiltering module 

* to prepare for abstraction over different wallet types (HdRnd/EOS for now) in the prefiltering code, we need generic `WalletId/AccountId/AddressId` types instead of HdRnd specific ones
* the return types of the core prefiltering api have a non-trivial impact on the wallet Kernel, including the AcidState DB types. To limit the impact of this, I've added scaffolding to `PrefilterTx.hs` to keep the API types intact for now. This code will be removed as part of #34 

## Commit 2 - Polymorphic prefiltering with IsOurs + PrefilterKey

For the Prefiltering API - prefilterUtxo/prefilterBlock/filterOurs: 

* refactor to generic prefiltering (from HdRnd specific)
* define polymorhic API by dispatching isOurs on the new PrefilterKey sum type (defined for multiple wallets types)
* simplify API by refactoring some custom prefiltering happening in Restore.hs
* decouple HdRnd specific code from generic prefiltering (includes context e.g. NetworkMagic)

# Comments

# Checklist

- [X] I've kept this PR simple, on-the-spot, free of cosmetic changes.
- [X] I have reviewed my commit messages and history.
- [X] I've assigned a reviewer.
- [ ] I acknowledge my changes may require an update in the wiki.
- [X ] I have added a reference to this PR to the corresponding issue.

---

- [X] I've checked the checklist

### DIFF
--- a/src/Cardano/Wallet/Kernel/PrefilterTx.hs
+++ b/src/Cardano/Wallet/Kernel/PrefilterTx.hs
@@ -4,17 +4,22 @@
 {-# LANGUAGE TypeFamilies        #-}
 
 module Cardano.Wallet.Kernel.PrefilterTx
-       ( PrefilteredBlock(..)
-       , emptyPrefilteredBlock
+       ( PrefilteredBlock (..)
+       , PrefilterKey (..)
        , AddrWithId
-       , prefilterBlock
-       , prefilterUtxo
        , UtxoWithAddrId
-       , prefilterUtxo'
-       , filterOurs
-       , toHdAddressId
-       , WalletKey
-       , toPrefilteredUtxo
+       -- Prefilter API
+       , filterOursHdRnd        -- TODO #34 deprecate -> use filterOurs
+       , prefilterBlockHdRnd    -- TODO #34 deprecate -> use prefilterBlock
+       , prefilterUtxoHdRnd     -- TODO #34 deprecate -> use prefilterUtxo
+       -- HdRnd Helpers         -- TODO #34 move to PrefilterHdRnd
+       , addressIdFromMeta
+       , toHdRndPrefKey
+       , toHdRndPrefKey'
+       , toHdRndPrefKeys
+       -- Auxiliary
+       , emptyPrefilteredBlock
+       , toPrefilteredUtxoHdRnd -- TODO #34 deprecate -> use toPrefilteredUtxo
        ) where
 
 import           Universum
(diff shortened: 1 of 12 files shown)


KtorZ:
@uroboros 

I am not sure to fully understand the direction you're taking here. Are you able to detail this (as a comment on #34) in prose or as a list of points you're going to tackle?

src/Cardano/Wallet/Kernel/PrefilterTx.hs
```diff
@@ -147,17 +128,88 @@ data AddressSummary = AddressSummary {
...
+    , pfbContext__       :: !BlockContext
+    }
+
+type UtxoWithAddrId__ = Map TxIn (TxOutAux,AddressId)
+
+{-------------------------------------------------------------------------------
+ Core definitions of discovery of "our" addresses
+-------------------------------------------------------------------------------}
+
+-- | We assume that the AddressId provides access to AccountId
```
KtorZ, line 171: This assumption is actually wrong for sequential wallets and wasn't a right decision in random derivation we ought to follow.

src/Cardano/Wallet/Kernel/PrefilterTx.hs
```diff
@@ -147,17 +128,88 @@ data AddressSummary = AddressSummary {
...
+
+-- | For each wallet type, captures all the data necessary to discover "our" addresses.
+--   This type provides the basis for defining a polymorphic isOurs for different
+--   wallet types.
+data PrefilterKey
+    = PrefilterKeyHdRnd WalletId WalletDecrCredentials
+    -- ^ for HdRnd wallets, the WalletId is based on the HdRootId
+    -- and the credentials are derived from the EncryptedSecretKey
+    | PrefilterKeyEOS V1.EosWalletId PublicKey
+    -- ^ for EOS wallets, we need the root EosWalletId and the account public key
```
KtorZ, line 182: I am not sure about this kind of sum types. Why Prefiltering should have to know about the distinction between externally owned and fully owned wallets? I mean, we can inject the behavior without introducing the such higher level concepts. 

I am still not convinced about the need for abstracting `***Id` types here. Being polymorphic on the relevant types should be enough IMO.

src/Cardano/Wallet/Kernel/Types.hs
```diff
@@ -47,15 +51,10 @@ import qualified Cardano.Wallet.Kernel.Util.Core as Core
...
 
-    {- potential future kinds of wallet IDs:
-    -- | HD wallet with sequentially generated addresses
-    | WalletIdHdSeq ...
-
-    -- | External wallet (all crypto done off-site, like hardware wallets)
-    | WalletIdExt ...
-    -}
+  -- | External wallet (all crypto done off-site, like hardware wallets)
+  | WalletIdEOS EosWalletId
```
KtorZ, line 57: Minor thought but, names feel wrong here. Sounds like we are joining two different things under a same "WalletId" concept, only to have some branches irrelevant for `WalletIdEOS`. In the end, we should ask ourselves: is this `WalletId` really a type we want here ?

Issue #88, opened 2018-11-22
Remove readerT mechanism that uses MonadBListener from the wallet and fetches blocks through the diffusion layer


paweljakubas:
# Context

In order to operate and maintain its own UTxO state, the wallet applies blocks successively and extract information from them. This current implementation has been described in:

https://github.com/input-output-hk/cardano-wallet/wiki/ApplyingBlocks-Rollback-After-Decoupling-Byron

Here, we can see how blocks are received through a shared environment mechanism between the wallet and its underlying node. This creates strong memory coupling between both processes.

# Decision

We want to get rid of this shared environment (through `readerT`) and rely on the diffusion layer capabilities instead. This includes refactoring the current `BListener`, the overall block application and rollbacks mechanism. Note that the wallet currently receives `Blund`s from the `BListener` instead of mere `Blocks`. However, `Undo`s from `Blunds` are only used to retrieve some transaction Metadata which are otherwise available in (and will have to be looked up from) the `UTxO` maintained by the wallet. 

# Acceptance Criterias

1. The `BListener` _mustn't_ use any strong memory coupling to retrieve blocks from the node
2. The `BListener` _must_ use the diffusion layer to receive blocks from the node
3. The diffusion layer interface _mustn't_ change

---

# Development Plan

- [x] getBlocks and requestTip support is added in diffusion layer
- [x] The additional thread is instantiated that is responsible for downloading blocks from the node
- [x] The added mechanism is based on pull paradigm, ie., the wallet is the initiator for pulling blocks, that takes responsibility the blocks are in line with what the trusted node has
- [x] The same resolved blocks are created using the added mechanism as it is in case of old mechanism
- [x] apply blocks part of old mechanism is commented out
- [x] All integration tests pass
- [x] the pull mechanism is documented in the wiki
- [x] verify if occasionally dedicated thread for applying blocks got stuck and if yes make it stable 
- [x] rollback mechanism is proposed and added to wiki
- [x] data needed to realize pull mechanism is taken from wallet state rather than by invoke params (ie., consumed block headers can be retrieved from wallet state)
- [x] simplify tickDiffusionFunction further and check `processBlockTx` (utxo should probably be updated from the oldest to youngest block)
- [ ] document manual testing in `node-attached-to-mainnet-with-wallet` setup
- [ ] check restoration in `node-attached-to-mainnet-with-wallet` setup 
- [ ] check backfilling  in `node-attached-to-mainnet-with-wallet` setup
[...]

PR #140, opened 2018-12-07
Finalize cluster setup and running of integration test scenarios

KtorZ:
# Linked Issue

<p align="right">#117</p>

# Overview

- [x] Start a local cluster of node with an actual "real-life" topology setup
- [x] Generate initial faucet wallets with millions of ADA in it, every test scenario got assigned one
- [x] Hook up the test in the CI as an extra parallel job 
- [x] Extend coverage report to also take integration tests into account

# Comments

Two bugs discovered as part of this integration :( 

# Checklist

- [ ] I have added a reference to this PR to the corresponding issue.

---


test/integration/Test/Integration/APISanityChecks.hs
```diff
@@ -1,132 +0,0 @@
-{-# OPTIONS_GHC -fno-warn-orphans #-}
```
KtorZ, line 1: So, I've entirely removed that one. It was... buggy and I am actually not even sure we use to run it before. 
I like the idea however and we might reconsider later. But for now, let's leave it out.

test/integration/Test/Integration/Framework/Cluster.hs
```diff
@@ -0,0 +1,182 @@
+module Test.Integration.Framework.Cluster
```
KtorZ, line 1: So hopefully, a lot of this will go when we achieve full decoupling. Here, we still need to hijack an edge node's identity and sort of re-write the `startCluster` method normally provided by the cluster package.

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -329,13 +301,13 @@ walletName =
...
     :: (MonadIO m, MonadFail m, Show a, Eq a)
-    => Getter s a
+    => Lens' s a
     -> a
     -> Either ClientError s
     -> m ()
 expectFieldEqual getter a = \case
     Left e  -> wantedSuccessButError e
-    Right s -> a `shouldBe` view getter s
+    Right s -> view getter s `shouldBe` a
```
KtorZ, line 310: The error outputs renders better when we put what we expect first :)

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -569,16 +520,20 @@ withNextFaucet actionWithFaucet = do
...
-        \ wallets should you want to right more tests :)"
+        "\nFailed to setup new scenario: there's no more available faucet wallets!\
+        \\nWe import a faucet wallet for each scenario but only have a limited\
+        \ number of them. This can be modified directly in the configuration file,\
+        \ by default in: \n\n\ttest/integration/configuration.yaml\
+        \\n\ntry increasing the number of available 'poors' wallets\
+        \\n\nspec: &default_core_genesis_spec\
+        \\n\tinitializer:\
+        \\n\t\ttestBalance:\
+        \\n\t\t\tpoors: ???\n"
```
KtorZ, line 531: Gotta love error messages :)

test/integration/Test/Integration/Framework/Scenario.hs
```diff
@@ -42,19 +41,7 @@ instance Example (Scenario s IO ()) where
...
-
-    mapItem :: Item ctx -> Item ()
-    mapItem item = item
-        { itemExample =
-            \p action -> (itemExample item) p (action . mapActionWith)
-        }
+            let acquire = takeMVar mvar
+            let release = putMVar mvar
+            let between = runStateT io >=> (putMVar mvar . snd)
+            bracketOnError acquire release between
```
KtorZ, line 47: So, when a test fails, the underlying `io` actually throws. So if we don't put the `MVar` back, we're screwed because the next one holds forever.

test/integration/Test/Integration/Framework/Scenario.hs
```diff
@@ -42,19 +41,7 @@ instance Example (Scenario s IO ()) where
...
-runWithContext ctx = mapSpecItem mapActionWith mapItem
-  where
-    mapActionWith :: ActionWith ctx -> ActionWith ()
-    mapActionWith fn = \() -> fn ctx
-
-    mapItem :: Item ctx -> Item ()
-    mapItem item = item
-        { itemExample =
-            \p action -> (itemExample item) p (action . mapActionWith)
-        }
```
KtorZ, line 60: now it's just `beforeAll (newMVar ctx)` o/

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -12,7 +12,7 @@ spec :: Scenarios Context
 spec = do
     scenario "successful payment appears in the history" $ do
         fixture <- setup $ defaultSetup
-            & initialCoins .~ [100]
+            & initialCoins .~ [1000000]
```
KtorZ, line 15: Kinda forgot about fees when I wrote this.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -42,18 +43,18 @@ spec = do
...
         response <- request $ Client.getUtxoStatistics
-            $- (fixture ^. walletId)
+            $- (fixture ^. wallet . walletId)
 
         verify response
             [ expectWalletUTxO []
             ]
 
-    scenario "UTxO statistics reflect wallet's activity" $ do
+    xscenario "UTxO statistics reflect wallet's activity" $ do
```
KtorZ, line 52: So this one fails and has a weird output.. Only the tx above `1000` figures in the stats... I couldn't quite figure out why at first glance so I left it out for now and will open a corresponding ticket.

test/integration/configuration.yaml
```diff
@@ -0,0 +1,143 @@
...
+      #   file: genesis.json
+      #   hash: bcfbfcad6cf78f2363568e76af9d6e927f71c3683aafe5e289796579792dccdb
+      spec: &default_core_genesis_spec
+        initializer:
+          testBalance:
+            poors:        100
+            richmen:      3
+            richmenShare: 0.99
+            useHDAddresses: True
+            totalBalance: 11111111111111110000
```
KtorZ, line 16: So, this is roughly the part we're interested in. This gives us `100` faucet wallets which all have roughly 300B Ada each...  if we need more faucets (one per scenario), we can simply increase `poors`, and if we need more ADA per faucet, we can increase `totalBalance` or reduce `richmenShare`. Note that generating `poors` isn't "free" of time, because we have to generate all the corresponding keys (hence the little time before the test get started!). The more we have, the longer it is. For know set to 100, and when we'll lack of it, it fails with a rather explicit error (cf above).

test/integration/configuration.yaml
```diff
@@ -0,0 +1,143 @@
...
+          fakeAvvmBalance:
+            count: 10
+            oneBalance: 100000
+          avvmBalanceFactor: 1
+          useHeavyDlg: True  # it's better to use heavy delegation to test it more often
+          seed: 0
+        blockVersionData: &default_core_genesis_spec_blockVersionData
+          scriptVersion:     0
+          slotDuration:      5000
```
KtorZ, line 25: Gotta see down to what we can go here :stuck_out_tongue:  Tests ran in less than 10 minutes on my machine which is rather okay I believe. The delay are usually when we wait after transactions to be validated; by reducing the slot time, we reduce the delay.

paweljakubas (approved):
Thanks for the comments - really helped when going through the PR. 

When finished with #88 I am eager to look at UTXO statistics stuff and take the corresponding ticket !!!

KtorZ:
Merging as is. Will open another PR with documentation ASAP :+1:

Issue #141, opened 2018-12-07
Metadata store becomes inconsistent after wallet or account is deleted


KtorZ:
# Context
 
<A clear and concise description of what the bug is.>

| |  |
| --- | --- | 
| Commit Hash / Release Tag | release/2.0.0 |
| Operating System | (Windows, OSX, Linux) | 

We store some information related to transaction in the metadata store (sqlite). However, when looking up transaction with metadata referring to wallet we don't know about, we fail with a not so friendly error "WalletNotFound" despite no wallet being given as part of the query. 

This was also identified by Kostas some weeks ago: https://iohk.myjetbrains.com/youtrack/issue/CBR-468

# Steps to Reproduce

<Steps to reproduce the behavior>

1. Create or restore a wallet 
2. Move some funds from or to this wallet
3. Delete this wallet
4. Create another (different) wallet
5. Lookup the transaction history via `GET /api/v1/transactions` (not providing any filters)

```hs
    scenario "can fetch transaction history" $ do
        -- Here, the setup action does the first 4 steps. It 
        --     - restores a faucet wallet
        --     - create a fixture wallet
        --     - move the corresponding funds from the faucet wallet to the fixture wallet
        --     - then it deletes the faucet wallet
        _ <- setup $ defaultSetup 
            & initialCoins .~ [14]

        response <- request $ Client.getTransactionIndexFilterSorts
            $- Nothing
            $- Nothing
            $- Nothing
            $- defaultPage
            $- defaultPerPage
            $- NoFilters
            $- NoSorts

        verify response
            [ expectSuccess
            ]
```

# Expected behavior

<A clear and concise description of what you expected to happen.>

The API returns a 200 OK with an history containing transactions from all known wallets.

# Actual behavior
 
<A clear and concise description of what you observe instead. If applicable add screenshots to help explain your problem.>

The scenario fails with:

```
expected a successful response but got an error: ClientWalletError WalletNotFound
```

Meanwhile, in the log..

```
[cardano-sl.walletapi:Info:ThreadId 1072] [2018-12-08 13:53:59.13 UTC] 
DELETE Request #12
    :> api
    :> v1
    :> wallets
    :> 'walletId' field: <wallet id>
[cardano-sl.walletapi:Info:ThreadId 1072] [2018-12-08 13:53:59.13 UTC] 
    Response #12 OK 0.009651908s
[...]

PR #146, opened 2018-12-10
Fix Metadata Inconsistency After Wallet or Account Deletion

KtorZ:
# Linked Issue

<p align="right">#141</p>

# Overview

- [x]  It discards incoherent transactions fetched from the DB, if any, and shout a warning in the log.This is in order to make the system more resilient to concurrent calls while a wallet or account is being deleted (since metadata and accounts / wallets are stored in separated databases, we can't easily run both delete in a single transaction).

- [x] It deletes corresponding metadata when an account or a wallet is removed. This may cause extra damage? What if there are pending transactions when we delete the account or wallet.

- [x] It now ignores insertion violations thay may occur if a user restore a deleted wallet or account. Cleaning up the metadata table itself isn't sufficient; we also need (in theory) to cleanup the corresponding inputs and outputs though in practice, this isn't that simple. Inputs may be used by different tx. Ideally, we would rely on db foreign keys for that but `beam` (the library we use to talk to SQLite) doesn't support them... Therefore, dust remains in the DB after deletion which is acceptable for now.

- [x] It adds a few regression integration scenarios to illustrate the various fixes.

# Comments

This is actually a port of [cardano-sl#3943](https://github.com/input-output-hk/cardano-sl/pull/3943) already reviewed and discussed with Kostas.

# Checklist

---

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
+++ b/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
@@ -18,6 +18,7 @@ module Cardano.Wallet.Kernel.DB.Sqlite (
     , putTxMeta
     , getTxMeta
     , getTxMetas
+    , deleteTxMetas
 
     -- * Unsafe functions
     , unsafeMigrateMetaDB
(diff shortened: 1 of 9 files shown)


paweljakubas (approved):
The PR introduces precise clearing of meta that is needed in a number of cases, for instance, when the wallet is deleted and later comes back or when the account of the wallet is deleted. Without this the respective metadata was inconsistent and results in misleading response and behavior. The exact scenarios targeted are "spelled and written down" in the integration test. 

I am very fond of this way of addressing errors in the integration tests. It boosts confidence that the problem was treated properly, makes easier to review the PR and make sure the same bug is not going to elude us later and resurface. Probably it should be a requirement to add the corresponding integration tests for each error/bug not covered.

PR #149, opened 2018-12-11
Review Integration Test: Part 3 (wrapping up)

KtorZ:
# Linked Issue

<p align="right">#117</p>

# Overview

- [x] Add link to the wiki guide on integration tests
- [x] Make sure to fetch (and thus generate) the api documentation as part of the integration tests.
- [x] Remove the `Maybe` wrapper around the setup backup phrase (slightly less verbose)
- [x] Remove passing around the `wallet` when waiting for tx status. This was done temporarily to cope with #141.

# Comments

¯\\\_(ツ)\_/¯

# Checklist

- [ ] I have added a reference to this PR to the corresponding issue.

---

### DIFF
--- a/test/integration/Main.hs
+++ b/test/integration/Main.hs
@@ -2,9 +2,14 @@ module Main where
 
 import           Universum hiding (keys)
 
+import           Data.Map.Strict ((!))
 import           Test.Hspec (beforeAll, describe, hspec)
 
 import           Cardano.Cluster (MaxWaitingTime (..), NodeType (..))
+import           Cardano.Cluster.Util (ntwrkAddrToBaseUrl,
+                     unsafeNetworkAddressFromString)
+import           Cardano.Wallet.Client.Http (mkHttpClient, mkHttpDocClient)
+import qualified Test.Integration.Documentation as Documentation
 import           Test.Integration.Framework.Cluster (startCluster, waitForNode)
 import           Test.Integration.Framework.DSL (Context (..))
 import qualified Test.Integration.Scenario.Accounts as Accounts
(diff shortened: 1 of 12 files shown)


test/integration/Main.hs
```diff
@@ -2,33 +2,46 @@ module Main where
...
+    (env, keys, manager) <- startCluster
         [ ("core0", NodeCore)
         , ("core1", NodeCore)
         , ("core2", NodeCore)
         , ("relay", NodeRelay)
         , ("edge", NodeEdge)
         ]
-    waitForNode (_client ctx) (MaxWaitingTime 90)
+    let wClient = mkHttpClient (toBaseUrl $ env ! "WALLET_ADDRESS") manager
+    let dClient = mkHttpDocClient (toBaseUrl $ env ! "WALLET_DOC_ADDRESS") manager
```
KtorZ, line 34: NOTE: We are defaulting these two values, so are sure they're there!

test/integration/Test/Integration/Documentation.hs
```diff
@@ -0,0 +1,17 @@
...
+
+import           Cardano.Wallet.Client.Http (WalletDocHttpClient)
+import qualified Cardano.Wallet.Client.Http as Client
+
+
+spec :: WalletDocHttpClient -> Spec
+spec client = do
+    it "Fetches the documentation from the API" $ do
+        response <- runExceptT $ Client.getSwaggerJson client
+        response `shouldSatisfy` isRight
```
KtorZ, line 17: By fetching the documentation, we make sure that it actually gets rendered properly (basically invoking a whole bunch of `Example` classes as well as pretty much everything in `Swagger.hs`). We have quite a few "unsafe" things there that aren't caught at runtime. However, they'll also prevent the doc from being rendered, so if this test passes, we can assume that things are fine.

test/integration/Test/Integration/Framework/Cluster.hs
```diff
@@ -85,17 +84,15 @@ startCluster nodes = do
                 yield Nothing >> startNode node nodeEnv
             NodeEdge -> do
                 manager <- init topology >> init logger >> init tls
-                let waddr  = unsafeNetworkAddressFromString walletAddr
-                let client = mkHttpClient (ntwrkAddrToBaseUrl waddr) manager
-                yield (Just client) >> startWallet node nodeEnv
-    let client = Prelude.head $ catMaybes (map snd handles)
+                yield (Just (nodeEnv, manager)) >> startWallet node nodeEnv
+    let (env, manager) = Prelude.head $ catMaybes (map snd handles)
```
KtorZ, line 88: So basically, we are passing around the `Manager` and defer the creation of the client. We'll use the same manager to instantiate both a wallet client and a doc client (since the underlying server uses the same TLS certificates anyway!)

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -158,7 +158,7 @@ data Setup = Setup
     , _assuranceLevel
         :: AssuranceLevel
     , _mnemonicWords
-        :: Maybe [Text]
+        :: [Text]
```
KtorZ, line 161: The reason for this is purely esthetic. `mnemonicWords .~ ["patate", "autruche", ...]` reads better than `mnemonicWords .~ Just ["patate", "autruche" ....`. And I don't want us to be introducing every possible operator from the lens library (So yes, I am aware of `(?~)`). But let's leave it simple :)

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -393,14 +392,13 @@ expectAddressInIndexOf = \case
 -- seconds if not.
 expectTxStatusEventually
     :: (MonadIO m, MonadFail m, MonadReader ctx m, HasHttpClient ctx)
-    => Wallet
-    -> [TransactionStatus]
+    => [TransactionStatus]
     -> Either ClientError Transaction
     -> m ()
-expectTxStatusEventually w statuses = \case
+expectTxStatusEventually statuses = \case
```
KtorZ, line 398: So the `Wallet` here was only needed because of the bug described in #141. Now that is is fixed, the parameter is superfluous.

paweljakubas (approved):
LGTM!

PR #152, opened 2018-12-12
Remove V1 wrapper: InputSelectionPolicy

denisshevchenko:
<p align="right">#119</p>

# Overview

- [x] `InputSelectionPolicy` doesn't use `V1` wrapper anymore.
- [x] `InputSelectionPolicy` golden test passes for old and new versions.

# Comments

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -42,6 +42,7 @@ module Cardano.Wallet.API.V1.Types (
   , mkSpendingPassword
   , WalletPassPhrase (..)
   , WalletTimestamp (..)
+  , WalletInputSelectionPolicy (..)
   , EosWallet (..)
   , NewEosWallet (..)
   , WalletAndTxHistory (..)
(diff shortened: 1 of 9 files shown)


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1503,27 +1504,35 @@ instance BuildableSafeGen PaymentSource where
...
-  , pmtGroupingPolicy   :: !(Maybe (V1 Core.InputSelectionPolicy))
+  , pmtGroupingPolicy   :: !(Maybe WalletInputSelectionPolicy)
   , pmtSpendingPassword :: !(Maybe SpendingPassword)
   } deriving (Show, Eq, Generic)
 
-instance ToJSON (V1 Core.InputSelectionPolicy) where
-    toJSON (V1 Core.OptimizeForSecurity)       = String "OptimizeForSecurity"
-    toJSON (V1 Core.OptimizeForHighThroughput) = String "OptimizeForHighThroughput"
+newtype WalletInputSelectionPolicy = WalletInputSelectionPolicy Core.InputSelectionPolicy
+    deriving (Eq, Generic, Show)
```
KtorZ, line 1512: The "funny" thing is that this `InputSelectionPolicy` is defined in `cardano-sl/client` but only used in `wallet` :upside_down_face:  ... In the long run, I think we want to get rid of this `client` dependency. We only use it for very few things but most of the package should have already been deprecated. 

For later though.

KtorZ (approved):
Love the commit history :heart:

PR #153, opened 2018-12-13
[DEVOPS-1179] nix-tools build

rvl:
Adds a nix-tools based build which can cross-compile `cardano-node.exe` for Windows.

https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-1179


default.nix
```diff
@@ -0,0 +1,20 @@
+{ system ? builtins.currentSystem
+, crossSystem ? null
+, config ? {}
+, iohkLib ? import ./nix/iohk-common.nix { inherit system crossSystem config; application = "cardano-sl"; }
+, pkgs ? iohkLib.pkgs
+, gitrev ? iohkLib.commitIdFromGitRepo ./.
```
jbgi, line 20: `gitrev` is not used anywhere?

release.nix
```diff
@@ -0,0 +1,31 @@
+let
+  iohkLib = import ./nix/iohk-common.nix { application = "cardano-sl"; };
+  fixedNixpkgs = iohkLib.pkgs;
+
+in { supportedSystems ? [ "x86_64-linux" "x86_64-darwin" ]
+  , scrubJobs ? true
+  , cardano-wallet ? { outPath = ./.; rev = "abcdef"; }
```
jbgi: Is there a point in using an optional arg here, if the default value does not make sense (`"abcdef"`)?

jbgi, line 20: the `outPath` doesn't look like it is used?  should `cardano-wallet` be replaced by just a `rev` parameter?

default.nix
rvl: Thanks for the excellent review comments.

This `gitrev` arg is not used anywhere but a useful side-effect of its presence is that Hydra will _not_ cache the evaluation of the release jobset when the git revision changes.

I should add a comment.

release.nix
rvl: This is a bug. Hydra will supply the `cardano-wallet` argument in this format. The source should be passed to the build.

default.nix
jbgi: I see, though I am not totally clear on why we don't want hydra to use the cache when the git revision change? Is it due to possible impurities?

nix/pkgs.nix
```diff
@@ -0,0 +1,190 @@
...
+        packages.conduit.patches            = [ ./patches/conduit-1.3.0.2.patch ];
+        packages.cryptonite-openssl.patches = [ ./patches/cryptonite-openssl-0.7.patch ];
+        packages.streaming-commons.patches  = [ ./patches/streaming-commons-0.2.0.0.patch ];
+        packages.x509-system.patches        = [ ./patches/x509-system-1.6.6.patch ];
+        packages.file-embed-lzma.patches    = [ ./patches/file-embed-lzma-0.patch ];
+        packages.cardano-sl.patches         = [ ./patches/cardano-sl.patch ];
+      }
+
+      # cross compilation logic
+      ({ pkgs, buildModules, config, lib, ... }:
```
jbgi: `buildModules` is not used. bug? or to remove?

rvl: It's not used -- I will remove it.

.gitignore
```diff
@@ -31,4 +31,16 @@ state-integration
...
 wallet-db-acid/
-wallet-db-sqlite.sqlite3
\ No newline at end of file
+wallet-db-sqlite.sqlite3
+
+### Nix ###
+
+/result*
+
+# GitHub Linguist annotations.
```
angerman: Does this actually work? Doesn't this belong to the `.gitattributes` file?
``` 
# GitHub Linguist annotations.
# Hide nix/.stack.nix/*.nix
# That is stuff that is generated by nix-tools stack-to-nix

nix/.stack.nix/*.nix linguist-generated=true
.stack-to-nix.cache linguist-generated=true
nix/.stack-pkgs.nix linguist-generated=true
```

angerman (dismissed):

KtorZ:
*please let me know when you'll be willing to merge this :)

.gitignore
rvl: Ah good catch -- I cargo culted that from cardano-sl.

rvl:
> please let me know when you'll be willing to merge this

Hi @KtorZ, we would like to merge it soon. It is ready for you to review now.

disassembler (approved):
LGTM! Once we get a dev okay, I say lets merge.

KtorZ:
There isn't much I could review on that since it's mostly nix stuff, located in a `nix` folder as we discussed so, that's fine by me. 

Yet, didn't we agree to not re-do the full nix build pipeline with "legacy sl" but instead, to do it on the legacy-free code on `cardano-shell` :thinking: ? 
Or are we actually moving on with `cardano-wallet` as the replacement for building nodes? 

We are going to drop `cardano-sl` from our extra dependencies A.S.A.P to rely on `cardano-chain` and `ouroboros-network` as much as possible. 

cc: @rvl @disassembler @jbgi @angerman

rvl:
@KtorZ I don't think `cardano-shell` is ready yet -- it doesn't use `cardano-wallet` as a library. Until that happens, this repo is the best place to ensure that cross-compiled wallet builds work. We are also trying to keep as much common code as possible in the [iohk-nix](https://github.com/input-output-hk/iohk-nix) repo.

As far as this PR is concerned, it should be fine to add/drop dependencies -- as long as the stack builds still work, the nix builds should too (in theory).

rvl:
I have rebased on latest develop and (temporaily) disabled the builds on macOS ([DEVOPS-1202](https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-1202)).

KtorZ (approved):

PR #155, opened 2018-12-13
Remove V1 wrapper: Address

denisshevchenko:
<p align="right">#119</p>

# Overview

- [x] `Core.Address` doesn't use `V1` wrapper anymore.
- [x] `Core.Address` golden test passes for old and new versions.

# Comments

### DIFF
--- a/src/Cardano/Wallet/API/Indices.hs
+++ b/src/Cardano/Wallet/API/Indices.hs
@@ -62,8 +62,8 @@ instance ToIndex Transaction WalletTimestamp where
     toIndex _ = fmap WalletTimestamp . Core.parseTimestamp
     accessIx Transaction{..} = txCreationTime
 
-instance ToIndex WalletAddress (V1 Core.Address) where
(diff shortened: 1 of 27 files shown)


KtorZ (commented):
I think this one is a bit more subtle. I guess we don't want to wrapper to ripple everywhere in the Kernel. Wrapper boundaries should pretty much be set at the API-level. 

I mean, in the Kernel, what `WalAddress` buys us that `Address` doesn't already have :thinking: ?

denisshevchenko:
> I mean, in the Kernel, what WalAddress buys us that Address doesn't already have 

Actually, it buys nothing. I just found all places where `V1` was used for `Address` and changed it to `WalAddress`. :-) Anyway, we have to remove `V1` from _all_ places. Maybe we have to finish all these PRs (getting rid of `V1` wrapper completely) and after that we'll change it in `Kernel`-level?

KtorZ (approved):

KtorZ:
Okay, let's first get rid of V1 and then see if wrapper are actually needed. Could you make sure to add this as a TODO for this ticket?

Issue #41, opened 2018-11-07
BIP-44 Utils Part II: Hardened Derivations


KtorZ:
# Context

<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

This is finalizing the work started in #31 but including private key derivations and operations this time.

# Decision

<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

Extend the module defined in #31 to also support hardened derivation.

# Acceptance Criteria

<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

1. The module from #31.1 _must_ provide utils to perform child derivation of account private keys from root private keys.
2. The module from #31.1 _must_ provide utils to perform child derivation of address private keys from account private keys.
3. The module from #31.1 _must_ provide utils to perform public key derivation from private keys.

# Development Plan

 - [x] I plan on adding method to derive account private key from root/master private key
      - deriving through `purpose'` and `coin_type'` internal derivations
 - [x] I plan on adding method to derive address private key from account private key
      - deriving through `change` internal derivation
 - [x] I plan on adding method to derive extended public key from extended private key
 - [x] I plan on fixing current method for deriving address public key from account public key
      - https://github.com/input-output-hk/cardano-wallet/pull/133/files#diff-ab6bdce941f09e40581d4a2debad9d7aR26
      - should go through `change` internal derivation
 - [x] Open an issue for adding golden tests for every method above
      - Nicolas advised to use rust's `cardano-cli` to generate golden tests https://github.com/input-output-hk/cardano-cli, https://cardanorust.iohkdev.io/cli-wallet/
      - use `cardano-cli` instructions https://github.com/input-output-hk/cardano-wallet/wiki/Testing-Yoroi-and-Cardano-wallet
      - https://github.com/input-output-hk/cardano-wallet/issues/195
 - [x] respect naming convention "normal" vs "non-hardened"
      - poll https://input-output-rnd.slack.com/archives/GBT05825V/p1546277012162200
 - [x] I plan on adding additional unit/property tests
      - steal them from `cardano-crypto` lib, have a look at `cardano-cli`
 - [x] Open an issue for adding an isomorphic type to `Word31`
[...]

PR #162, opened 2018-12-17
Fix address public key derivation

akegalj:
<p align="right">#41</p>

# Overview

- [x] I have fixed current method for deriving address public key from account public key
https://github.com/input-output-hk/cardano-wallet/pull/133/files#diff-ab6bdce941f09e40581d4a2debad9d7aR26
    - added intermediate change internal derivation
- [x] Simplified naming of address public key derivation

# Comments

This commit fixes work done by #31. Instead of deriving address
public key dirrectly with account public key, bip44 specifies
intermediate Change derivation. Simplified naming.
https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#Change

### DIFF
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -1,26 +1,61 @@
 {-------------------------------------------------------------------------------
 
-  This module provides utils to perform account public key child derivation:
+  This module provides utils to perform derivation of address public key from
+  account public key:
 
     account (parent) extended public key -> address (child) extended public key
 
 -------------------------------------------------------------------------------}
 
-module Cardano.Wallet.Kernel.Ed25519Bip44 (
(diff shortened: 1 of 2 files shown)


src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -1,26 +1,61 @@
...
+    -- helpers
+    , isInternalChange
     ) where
 
 import           Universum
 
 import           Pos.Crypto (PublicKey (..))
 
-import           Cardano.Crypto.Wallet (DerivationScheme (..), deriveXPub)
+import           Cardano.Crypto.Wallet (DerivationScheme (DerivationScheme2),
```
akegalj, line 24: added explicit `DerivationScheme2` so we not confuse things

KtorZ (approved):
Nice, thanks @akegalj

PR #163, opened 2018-12-17
Derive extended public key from extended private key

akegalj:
Adds method described in the title

<p align="right">#41</p>

# Overview

- [x] I have added a method to derive extended public key from extended private key

# Comments

~I was not shure should we use `EncryptedSecretKey` or a plain `XPrv`. If we were using plain `XPrv` provided by `cardano-crypto` then we would still have to wrapp all the functions with `EncprytedSecretKey` provided by `cardano-sl` (`crypto` repo). I believe this is the way to go but here is a remark on my slight confusion.~
~EDIT: I have decided to use `SecretKey` instead as it seems more appropriate~

EDIT2: We have decided to use `EncryptedSecretKey` after all (see comments bellow)

Additionally as implemented functions just re-exports function from `cardano-sl` I was not sure how to actually test the function. When skimmed over internals how `XPrv -> XPub` is implemented low level (used internally by this method) how to test this function would be to test is part of public key binary representation a subset of private key binary representation. I would have to go into more details to closely understand format of these keys but this seems out of scope? Also - I am not implementing anything new in this case - just re-exporting (whith changed function name).

### DIFF
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -1,9 +1,9 @@
 {-------------------------------------------------------------------------------
 
-  This module provides utils to perform derivation of address public key from
-  account public key:
+  This module provides utils to perform child key derivation for bip44:
 
     account (parent) extended public key -> address (child) extended public key
+    extended private key -> extended private key
 
 -------------------------------------------------------------------------------}
 
@@ -12,6 +12,7 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
 
     -- key derivation functions
     , deriveAddressPublicKey
+    , derivePublicKey
 
     -- helpers
     , isInternalChange
@@ -19,7 +20,7 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
 
 import           Universum
 
-import           Pos.Crypto (PublicKey (..))
+import           Pos.Crypto (EncryptedSecretKey, PublicKey (..), encToPublic)
 
 import           Cardano.Crypto.Wallet (DerivationScheme (DerivationScheme2),
                      deriveXPub)
@@ -59,3 +60,8 @@ deriveAddressPublicKey (PublicKey accXPub) changeChain addressIx = do
     changeXPub <- deriveXPub DerivationScheme2 accXPub (changeToIndex changeChain)
     -- lvl5 derivation in bip44 is derivation of change address chain
     PublicKey <$> deriveXPub DerivationScheme2 changeXPub addressIx
+
+-- | Generate extend private key from extended private key
+-- (EncryptedSecretKey is a wrapper around private key)
+derivePublicKey :: EncryptedSecretKey -> PublicKey
+derivePublicKey = encToPublic


src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -24,3 +26,8 @@ deriveLvl3PublicKey
     -> Maybe PublicKey
 deriveLvl3PublicKey (PublicKey xpub) addressIx =
     PublicKey <$> deriveXPub DerivationScheme2 xpub addressIx
+
+-- | Generate extend private key from extended private key
+-- (EncryptedSecretKey is a wrapper around private key)
+derivePublicKey :: EncryptedSecretKey -> PublicKey
```
akegalj, line 66: ~I am not sure is `EncryptedSecretKey` the right type here. We might have to use more primitive type like https://github.com/input-output-hk/cardano-crypto/blob/master/src/Cardano/Crypto/Wallet.hs#L144 for some usecases but I am not sure in that yet~

I have decided to use `SecretKey` instead. We don't do anything with hash passphrase in these steps so why complicate

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -24,3 +26,7 @@ deriveLvl3PublicKey
     -> Maybe PublicKey
 deriveLvl3PublicKey (PublicKey xpub) addressIx =
     PublicKey <$> deriveXPub DerivationScheme2 xpub addressIx
+
+-- | Generate extend private key from extended private key
+derivePublicKey :: SecretKey -> PublicKey
+derivePublicKey = toPublic
```
KtorZ: Well... truth is that we want to manipulate only `EncryptedSecretKey`. We have to enforce password protection on secret key all the time. That doesn't change much what's happening here however, `toPublic` -> `encToPublic`.

akegalj: ~Well, that's what was my first thought as well. But when you think about it, bip44 and derivation of HD wallet tree doesn't really care is private key is encrypted or not. We just care about the topic of how to derive child keys from parent keys in this module and offload unwrapping `EncryptedSecretKey` and wrapping of `SecretKey` to functionalities outside this module...~

Hm - unwrapping of `EncryptedSecretKey` wouldn't be much of a problem (and doesn't lead to many bugs) - but wrapping of `SecretKey` into `EncryptedSecretKey` requires child to use same passphrase (probably used as passphrase validation). So I believe you are right, we have to use `EncryptedSecretKey`.

> We have to enforce password protection on secret key all the time
Looking at the type 
```haskell
-- | Encrypted HD secret key.
data EncryptedSecretKey = EncryptedSecretKey
    { eskPayload :: !CC.XPrv          -- ^ Secret key itself, encrypted with passphrase.
    , eskHash    :: !S.EncryptedPass  -- ^ Hash of passphrase used for key creation.
    }

```
I see `XPrv` (ie, `SecretKey`) already is password protected on its own (provided by `cardano-crypto`) but what we loose is hash of a password which is probably used by validation all over the api. Will use `EncryptedSecretKey` as you suggested

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
akegalj: See here https://github.com/input-output-hk/cardano-wallet/pull/163#discussion_r242284225

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
KtorZ: Very valid point. Though I don't think we do manipulate any `SecretKey` so to speak in the codebase, we always manipulate the wrapped version with password hash attached.

KtorZ (approved):

PR #165, opened 2018-12-17
Remove V1 wrapper: SoftwareVersion.

denisshevchenko:
<p align="right">#119</p>

# Overview

- [x] Type `SoftwareVersion` doesn't use `V1` wrapper.
- [x] Golden test both for old and new types passed.

# Comments

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -43,6 +43,7 @@ module Cardano.Wallet.API.V1.Types (
   , WalletPassPhrase (..)
   , WalletTimestamp (..)
   , WalletInputSelectionPolicy (..)
+  , WalletSoftwareVersion (..)
   , WalletTxId (..)
   , WalAddress (..)
   , WalletCoin (..)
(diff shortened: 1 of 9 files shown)


KtorZ:
@denisshevchenko Some styling issues unfortunately.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1982,6 +1984,41 @@ instance BuildableSafeGen WalletAndTxHistory where
...
+instance Arbitrary WalletSoftwareVersion where
+    arbitrary = WalletSoftwareVersion <$> arbitrary
+
+-- | NOTE: There is 'ToSchema' instance for 'V1 SoftwareVersion' type, but it's defined
+-- in 'cardano-sl-chain' package and we cannot touch it for now. So we have to define
+-- here some 'ToSchema'-instance for 'WalletSoftwareVersion'.
+instance ToSchema WalletSoftwareVersion where
+    declareNamedSchema _ =
+        pure $ NamedSchema (Just "WalletSoftwareVersion") $ mempty
+            & type_ .~ SwaggerObject
```
KtorZ, line 2030: So... If we weren't defining any custom instances for `SoftwareVersion`.. What was `V1` used for ... ? Sounds like we could just remove it entirely without going for a newtype.

denisshevchenko: I've tried it already, but it requires `ToSchema` for `SoftwareVersion`, it doesn't compile without it. But we cannot define orphan instances for `Core`-types, right?

KtorZ: but where do the previous `ToSchema` instance came from... ?

denisshevchenko: We have `instance ToSchema (V1 SoftwareVersion)` in `cardano-sl-chain` package.

KtorZ (approved):

KtorZ:
@denisshevchenko This conflicts with the PR on `Coin`.

denisshevchenko:
@KtorZ Fixed.

PR #168, opened 2018-12-18
Remove V1 wrapper: explicit import V1/unV1 from 'cardano-sl' package.

denisshevchenko:
<p align="right">#119</p>

# Overview

- [x] `V1` and `unV1` are imported explicitly from `cardano-sl` package.

# Comments

Unfortunately, there're two modules in `cardano-wallet` where we must use `V1` and `unV1`:

1. `Cardano.Wallet.WalletLayer.Kernel.Settings`
2. `Cardano.Wallet.WalletLayer.Kernel.Transactions`

This is because we use two types from `cardano-sl`, and these types require `V1` wrapper. So now, in these two modules we import `V1` and `unV1` explicitly from `Pos.Node.API`, not from `Cardano.Wallet.API.V1.Types`.

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -19,10 +19,8 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Cardano.Wallet.API.V1.Types (
-    V1 (..)
-  , unV1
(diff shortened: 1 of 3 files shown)


KtorZ (approved):

KtorZ:
Right, but at least, we don't spread the usage of `V1` to the whole codebase anymore. So that's fine.

PR #169, opened 2018-12-18
Derive address private key from account private key

akegalj:
<p align="right">#41</p>

# Overview

- [x] I have added a method to derive address private key from account private key
- [x] I have added property tests to check the above method

# Comments

~For some reason property test https://github.com/input-output-hk/cardano-wallet/compare/akegalj/41/derive-addrPrvKey-from-accPrvKey?expand=1#diff-e9a62968e1cca775db414be6d317ecadR125 fails and I have to debug where the issue is. In theory it should work~
Edit: fixed with https://github.com/input-output-hk/cardano-crypto/pull/52

### DIFF
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -23,7 +23,7 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
 import           Universum
 
 import           Pos.Crypto (EncryptedSecretKey (..), PassPhrase,
-                     PublicKey (..), checkPassMatches, encToPublic)
+                     PublicKey (..), checkPassMatches, encToPublic, isHardened)
 
 import           Cardano.Crypto.Wallet (DerivationScheme (DerivationScheme2),
                      deriveXPrv, deriveXPub)
@@ -61,9 +61,11 @@ derivePublicKey = encToPublic
 deriveAddressPublicKey
     :: PublicKey       -- Account Public Key
     -> ChangeChain     -- Change chain
-    -> Word32          -- Address Key Index
+    -> Word32          -- Non-hardened Address Key Index
     -> Maybe PublicKey -- Address Public Key
 deriveAddressPublicKey (PublicKey accXPub) changeChain addressIx = do
+    -- address index should be non hardened
+    guard (not $ isHardened addressIx)
     -- lvl4 derivation in bip44 is derivation of change chain
     changeXPub <- deriveXPub DerivationScheme2 accXPub (changeToIndex changeChain)
     -- lvl5 derivation in bip44 is derivation of address chain
@@ -76,11 +78,13 @@ deriveAddressPrivateKey
     :: PassPhrase               -- Passphrase used to encrypt Account Private Key
     -> EncryptedSecretKey       -- Account Private Key
     -> ChangeChain              -- Change chain
-    -> Word32                   -- Address Key Index
+    -> Word32                   -- Non-hardened Address Key Index
     -> Maybe EncryptedSecretKey -- Address Private Key
 deriveAddressPrivateKey passPhrase accEncPrvKey@(EncryptedSecretKey accXPrv passHash) changeChain addressIx = do
     -- enforce valid PassPhrase check
     checkPassMatches passPhrase accEncPrvKey
+    -- address index should be non hardened
+    guard (not $ isHardened addressIx)
         -- lvl4 derivation in bip44 is derivation of change chain
     let changeXPrv = deriveXPrv DerivationScheme2 passPhrase accXPrv (changeToIndex changeChain)
         -- lvl5 derivation in bip44 is derivation of address chain
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -2,8 +2,9 @@
 
   This module provides utils to perform child key derivation for bip44:
 
-    account (parent) extended public key -> address (child) extended public key
-    extended private key -> extended private key
+    extended private key -> extended public key
+    account extended public key -> address extended public key
+    account extended private key -> address extended private key
 
 -------------------------------------------------------------------------------}
 
@@ -22,8 +23,7 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
 import           Universum
 
 import           Pos.Crypto (EncryptedSecretKey (..), PassPhrase,
-                     PublicKey (..), ShouldCheckPassphrase (..),
-                     checkPassMatches, encToPublic)
+                     PublicKey (..), checkPassMatches, encToPublic)
 
 import           Cardano.Crypto.Wallet (DerivationScheme (DerivationScheme2),
                      deriveXPrv, deriveXPub)
@@ -73,15 +73,14 @@ deriveAddressPublicKey (PublicKey accXPub) changeChain addressIx = do
 -- using derivation scheme 2 (please see 'cardano-crypto' package).
 -- TODO: EncryptedSecretKey will be renamed to EncryptedPrivateKey in #164
 deriveAddressPrivateKey
-    :: ShouldCheckPassphrase    -- Weather to call @checkPassMatches@
-    -> PassPhrase               -- Passphrase used to encrypt Account Private Key
+    :: PassPhrase               -- Passphrase used to encrypt Account Private Key
     -> EncryptedSecretKey       -- Account Private Key
     -> ChangeChain              -- Change chain
     -> Word32                   -- Address Key Index
     -> Maybe EncryptedSecretKey -- Address Private Key
-deriveAddressPrivateKey (ShouldCheckPassphrase checkPass) passPhrase accEncPrvKey@(EncryptedSecretKey accXPrv passHash) changeChain addressIx = do
+deriveAddressPrivateKey passPhrase accEncPrvKey@(EncryptedSecretKey accXPrv passHash) changeChain addressIx = do
     -- enforce valid PassPhrase check
-    when checkPass $ checkPassMatches passPhrase accEncPrvKey
+    checkPassMatches passPhrase accEncPrvKey
         -- lvl4 derivation in bip44 is derivation of change chain
     let changeXPrv = deriveXPrv DerivationScheme2 passPhrase accXPrv (changeToIndex changeChain)
         -- lvl5 derivation in bip44 is derivation of address chain
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -79,9 +79,9 @@ deriveAddressPrivateKey
     -> ChangeChain              -- Change chain
     -> Word32                   -- Address Key Index
     -> Maybe EncryptedSecretKey -- Address Private Key
-deriveAddressPrivateKey (ShouldCheckPassphrase checkPass) passPhrase accEncSK@(EncryptedSecretKey accXPrv passHash) changeChain addressIx = do
+deriveAddressPrivateKey (ShouldCheckPassphrase checkPass) passPhrase accEncPrvKey@(EncryptedSecretKey accXPrv passHash) changeChain addressIx = do
     -- enforce valid PassPhrase check
-    when checkPass $ checkPassMatches passPhrase accEncSK
+    when checkPass $ checkPassMatches passPhrase accEncPrvKey
         -- lvl4 derivation in bip44 is derivation of change chain
     let changeXPrv = deriveXPrv DerivationScheme2 passPhrase accXPrv (changeToIndex changeChain)
         -- lvl5 derivation in bip44 is derivation of address chain
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -12,6 +12,7 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
 
     -- key derivation functions
     , deriveAddressPublicKey
+    , deriveAddressPrivateKey
     , derivePublicKey
 
     -- helpers
@@ -20,10 +21,12 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
 
 import           Universum
 
-import           Pos.Crypto (EncryptedSecretKey, PublicKey (..), encToPublic)
+import           Pos.Crypto (EncryptedSecretKey (..), PassPhrase,
+                     PublicKey (..), ShouldCheckPassphrase (..),
+                     checkPassMatches, encToPublic)
 
 import           Cardano.Crypto.Wallet (DerivationScheme (DerivationScheme2),
-                     deriveXPub)
+                     deriveXPrv, deriveXPub)
 import           Test.QuickCheck (Arbitrary (..), elements)
 
 -- | Change chain. External chain is used for addresses that are meant to be visible
@@ -48,6 +51,11 @@ changeToIndex :: ChangeChain -> Word32
 changeToIndex ExternalChain = 0
 changeToIndex InternalChain = 1
 
+-- | Generate extend private key from extended private key
+-- (EncryptedSecretKey is a wrapper around private key)
+derivePublicKey :: EncryptedSecretKey -> PublicKey
+derivePublicKey = encToPublic
+
 -- | Derives address public key from the given account public key,
 -- using derivation scheme 2 (please see 'cardano-crypto' package).
 deriveAddressPublicKey
@@ -58,10 +66,24 @@ deriveAddressPublicKey
 deriveAddressPublicKey (PublicKey accXPub) changeChain addressIx = do
     -- lvl4 derivation in bip44 is derivation of change chain
     changeXPub <- deriveXPub DerivationScheme2 accXPub (changeToIndex changeChain)
-    -- lvl5 derivation in bip44 is derivation of change address chain
+    -- lvl5 derivation in bip44 is derivation of address chain
     PublicKey <$> deriveXPub DerivationScheme2 changeXPub addressIx
 
--- | Generate extend private key from extended private key
--- (EncryptedSecretKey is a wrapper around private key)
-derivePublicKey :: EncryptedSecretKey -> PublicKey
-derivePublicKey = encToPublic
+-- | Derives address private key from the given account private key,
+-- using derivation scheme 2 (please see 'cardano-crypto' package).
+-- TODO: EncryptedSecretKey will be renamed to EncryptedPrivateKey in #164
+deriveAddressPrivateKey
+    :: ShouldCheckPassphrase    -- Weather to call @checkPassMatches@
+    -> PassPhrase               -- Passphrase used to encrypt Account Private Key
+    -> EncryptedSecretKey       -- Account Private Key
+    -> ChangeChain              -- Change chain
+    -> Word32                   -- Address Key Index
+    -> Maybe EncryptedSecretKey -- Address Private Key
+deriveAddressPrivateKey (ShouldCheckPassphrase checkPass) passPhrase accEncSK@(EncryptedSecretKey accXPrv passHash) changeChain addressIx = do
+    -- enforce valid PassPhrase check
+    when checkPass $ checkPassMatches passPhrase accEncSK
+        -- lvl4 derivation in bip44 is derivation of change chain
+    let changeXPrv = deriveXPrv DerivationScheme2 passPhrase accXPrv (changeToIndex changeChain)
+        -- lvl5 derivation in bip44 is derivation of address chain
+        addressXPrv = deriveXPrv DerivationScheme2 passPhrase changeXPrv addressIx
+    pure $ EncryptedSecretKey addressXPrv passHash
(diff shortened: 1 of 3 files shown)


src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -48,6 +51,11 @@ changeToIndex :: ChangeChain -> Word32
 changeToIndex ExternalChain = 0
 changeToIndex InternalChain = 1
 
+-- | Generate extend private key from extended private key
+-- (EncryptedSecretKey is a wrapper around private key)
+derivePublicKey :: EncryptedSecretKey -> PublicKey
```
akegalj, line 81: This was just moved from the bottom of the module

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -58,10 +66,24 @@ deriveAddressPublicKey
 deriveAddressPublicKey (PublicKey accXPub) changeChain addressIx = do
     -- lvl4 derivation in bip44 is derivation of change chain
     changeXPub <- deriveXPub DerivationScheme2 accXPub (changeToIndex changeChain)
-    -- lvl5 derivation in bip44 is derivation of change address chain
+    -- lvl5 derivation in bip44 is derivation of address chain
```
akegalj, line 96: typo

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -25,15 +28,104 @@ prop_cannotDeriveAddressPublicKeyForBigIx
     -> Property
 prop_cannotDeriveAddressPublicKeyForBigIx accountPublicKey change addressIx = property $
     if addressIx >= maxIx
-        then isNothing result
-        else isJust result
+        then isNothing addrPubKey
+        else isJust addrPubKey
   where
-    result = deriveAddressPublicKey accountPublicKey change addressIx
+    addrPubKey = deriveAddressPublicKey accountPublicKey change addressIx
```
akegalj: renamed to give a bit more context to binding

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -25,15 +28,104 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+    -> EncryptedSecretKey
+    -> ChangeChain
+    -> Word32
+    -> Property
+prop_deriveAddressPublicFromAccountPrivateKey passPhrase accEncSK changeChain addressIx =
+    -- TODO (akegalj): check coverage with quickcheck @cover@
+    -- FIXME (akegalj): instead of doing this create generator for non-hardened keys.
+    -- This should in average discard 50% of examples and will thus be
+    -- 50% slower.
+    isNonHardened addressIx ==> (addrPrvKey1 === addrPrvKey2)
```
akegalj: for some reason this fails in quickcheck - should work in theory based on https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--public-child-key , so should do more debugging with it

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -25,15 +31,105 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+    -> Word32
+    -> Property
+prop_deriveAddressPublicFromAccountPrivateKey (InfiniteList seed _) passPhrase@(PassPhrase passBytes) changeChain addressIx =
+    -- TODO (akegalj): check coverage with quickcheck @cover@
+    -- FIXME (akegalj): instead of doing this create generator for non-hardened keys.
+    -- This should in average discard 50% of examples and will thus be
+    -- 50% slower.
+    isNonHardened addressIx ==> (addrPubKey1 === addrPubKey2)
+  where
+    accEncPrvKey = mkEncSecretWithSaltUnsafe emptySalt passPhrase $ generate (BS.pack $ take 32 seed) passBytes
```
akegalj, line 103: to address failure of this property I have tried to manually create encrypted account private key. My initial idea was that the property might be failing because generated `EncryptedSecretKey` didn't match generated `PassPhrase`, so in this version we manually create account encrypted secret key from seed and passphrase and we reuse this passphrase for child derivations.

Unfortunatelly this doesn't help and this property still fails:
```
> quickCheck $ prop_deriveAddressPublicFromAccountPrivateKey 
*** Failed! Falsifiable (after 58 tests and 46 shrinks):     
[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0] ++ ...
<passphrase>
ExternalChain
134217728
Just (PublicKey (XPub {xpubPublicKey = "U\t\166M\EM\175\215\161w}\237F+\174\155\161\176\v\232\202\228\SOH0[\189\151\206i\133\177\ETX\174", xpubChaincode = ChainCode "\140V\195\149\168\253\237\219/\173q\209|%\135\rJ\200H\240\208\&2\168K\165\133)<\GS\207S\188"})) /= Nothing
```
so I wonder why :| 

Seems like it fails when used seed is `[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]` , but looking at https://github.com/input-output-hk/cardano-crypto/blob/master/src/Cardano/Crypto/Wallet.hs#L92 doesn't reveal why this seed shouldn't work. Will manually do step by step to reveal why this doesn't pass

I might be doing something really wrong that I am not aware?

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -58,10 +66,24 @@ deriveAddressPublicKey
...
 
--- | Generate extend private key from extended private key
--- (EncryptedSecretKey is a wrapper around private key)
-derivePublicKey :: EncryptedSecretKey -> PublicKey
-derivePublicKey = encToPublic
+-- | Derives address private key from the given account private key,
+-- using derivation scheme 2 (please see 'cardano-crypto' package).
+-- TODO: EncryptedSecretKey will be renamed to EncryptedPrivateKey in #164
+deriveAddressPrivateKey
+    :: ShouldCheckPassphrase    -- Weather to call @checkPassMatches@
```
KtorZ: I believe this is useless here? It was present in `deriveLvl2KeyPair` because the function was doing many things at once. Here, we do want to check the passphrase every time, don't we ?

Or at least, if needed, we'll add it later on.

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -25,15 +31,105 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+    :: InfiniteList Word8
+    -> PassPhrase
+    -> ChangeChain
+    -> Word32
+    -> Property
+prop_deriveAddressPublicFromAccountPrivateKey (InfiniteList seed _) passPhrase@(PassPhrase passBytes) changeChain addressIx =
+    -- TODO (akegalj): check coverage with quickcheck @cover@
+    -- FIXME (akegalj): instead of doing this create generator for non-hardened keys.
+    -- This should in average discard 50% of examples and will thus be
+    -- 50% slower.
```
KtorZ: Great idea :+1:

test/unit/Test/Spec/Ed25519Bip44.hs
KtorZ: One general remark: it's usually a good idea to narrow down the number of parameters you have and create instead, several properties. For instance, you could write this one with a fixed passphrase (or, at the very most, a Maybe with a static passphrase, or nothing), and then, another property where the key is fixed but trying out many passphrases.

akegalj: yes - that way search space will be more specialized (around those free variables)

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
akegalj: Yes - I was confused why we don't check passphrase every time. I thought its some kind of optimization. But hd wallet child key derivation doesn't happen that often (at least this is my intuition) and in most cases check is needed.

I think your suggestion is sound - lets remove it for now and we will add it later if really needed (which I doubt)

KtorZ:
@akegalj , so what's up with this one ?

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: fixed https://github.com/input-output-hk/cardano-crypto/pull/52

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: fixed https://github.com/input-output-hk/cardano-crypto/pull/52

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
akegalj: Removed

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -23,17 +29,104 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+-- For details see https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--public-child-key
+prop_deriveAddressPublicFromAccountPrivateKey
+    :: InfiniteList Word8
+    -> PassPhrase
+    -> ChangeChain
+    -> Word32
+    -> Property
+prop_deriveAddressPublicFromAccountPrivateKey (InfiniteList seed _) passPhrase@(PassPhrase passBytes) changeChain addressIx =
+    -- TODO (akegalj): check coverage with quickcheck @cover@
+    -- FIXME (akegalj): instead of doing this create generator for non-hardened keys.
```
KtorZ: Make sure to put those on the ticket as bullet points :+1:, otherwise, we won't likely every tackle those points...

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -23,17 +29,104 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+            passPhrase
+            accEncPrvKey
+            changeChain
+            addressIx
+    -- CKDpub(N(kpar, cpar), i)
+    addrPubKey2 =
+        deriveAddressPublicKey
+            (derivePublicKey accEncPrvKey)
+            changeChain
+            addressIx
```
KtorZ, line 116: I am really a big fan of this property :purple_heart:

test/unit/Test/Spec/Ed25519Bip44.hs
KtorZ: Would be nice to at least make sure we aren't just comparing `Nothing` with `Nothing` at each step :man_shrugging:

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -23,17 +29,104 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+
+-- | Deriving address private key should always fail
+-- if account index is hardened
```
KtorZ: Comment feels wrong here :thinking: ? Shouldn't it be `address index` ?

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -23,17 +29,104 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+prop_deriveAddressPrivateKeyHardened
+    :: InfiniteList Word8
+    -> PassPhrase
+    -> ChangeChain
+    -> Word32
+    -> Property
+prop_deriveAddressPrivateKeyHardened (InfiniteList seed _) passPhrase@(PassPhrase passBytes) changeChain addressIx =
+    isHardened addressIx ==> property (isJust addrPrvKey)
```
KtorZ: How come this passes :thinking: ? Shouldn't it be `property (isNothing addrPrvKey)` ?

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -23,17 +29,104 @@ prop_cannotDeriveAddressPublicKeyForBigIx
...
+    describe "Deriving address public key" $ do
+        it "fails if address index is too big (hardened)" $
+            property prop_cannotDeriveAddressPublicKeyForBigIx
+        it "equals to deriving address private key and extracting public part from it: N(CKDpriv((kpar, cpar), i)) === CKDpub(N(kpar, cpar), i)" $
+            property prop_deriveAddressPublicFromAccountPrivateKey
+    describe "Deriving address private key" $ do
+        it "fails if password differs" $
+            expectFailure prop_deriveAddressPrivateKeyWrongPassword
+        it "fails if address index is hardened" $
+            expectFailure prop_deriveAddressPrivateKeyHardened
```
KtorZ: Ah! `expectFailure`, I get it. This feels a bit wrong here because, we aren't expecting the property to pass even once. I find it quite confusing here and having the property being written the other way around (expecting always `Nothing` when using a hardened index) makes the intent clearer IMO. 

Don't you think?

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: > Would be nice to at least make sure we aren't just comparing Nothing with Nothing at each step

true that - will update

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: found by accident actually - but I would like to add similar property more low level, to `cardano-crypto` where it actually bellongs (its bip32 property, not bip44). 

`cardano-wallet` is checked and safe this way - but some other applications/libs which are using `cardano-crypto` might not implement this safe net - so long term it would be wise to move it to `cardano-crypto`

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: fails because this property is expected to fail with `expectFailure prop_derivateAddressPrivateKeyHardened` 

Althought I am to sure is my usage of `expectFailure` correct. It seems like property succeeds if at least one failure is found (and quickcheck will stop at the first one which doesn't provide much coverage).

Alternative to `expectFailure` approach would be to cover all cases as @denisshevchenko did with https://github.com/input-output-hk/cardano-wallet/pull/169/files#diff-e9a62968e1cca775db414be6d317ecadR33 . I was not sure which one is better suited. 

https://github.com/input-output-hk/cardano-wallet/pull/169/files#diff-e9a62968e1cca775db414be6d317ecadR33 is a bit harder to read/understand in my opinion but covers both cases, where in my scenario one case is discarded by prefilter `==>`

what's your opinion on this topic?

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: would you prefer 
```haskell
isHardened addressIx ==> property (isNothing addrPrvKey)
```

or how denis covered both branches:
```haskell
     if isHardened addressIx
   	         then isNothing addrPrvKey
	         else isJust addrPrvKey
```
?

test/unit/Test/Spec/Ed25519Bip44.hs
KtorZ: Yeah, I noticed after the `expectFailure` below. The precondition arrow reads quite nicely to be honest. But I'd rather write a valid property here (so `IsNothing ...`). `expectFailure` can be use to put emphasize on things we may think of as properties but aren't in reality. Or, to cover highlight failing scenarios. 

About _precondition_ vs _if_, it'd personally favor two properties with precondition to really make the intent clear about what's being tested (also makes the failure clearer because we know exactly what part has failed). Also, if one of the cases never happens because of a poorly designed generator, then QuickCheck will give up trying to satisfy the precondition. Without instrumenting on the coverage, this will remain unnoticed in the `if .. then ... else` approach.

test/unit/Test/Spec/Ed25519Bip44.hs
KtorZ: Cf comment above :)

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: very good point - didn't thought about the later and quickcheck giving up

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: Added a point to #41 development plan

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: > Make sure to put those on the ticket as bullet points +1, otherwise, we won't likely every tackle those points...

added a point to #41 development plan

> Would be nice to at least make sure we aren't just comparing Nothing with Nothing at each step

pending...

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: added point to #41 development plan

akegalj: added as a point to development plan #41

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: Resolved

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: Implementation of this requirement requires me to implement point from #41 development plan:
> Open an issue to add arbitrary generator for hardened and non-hardened indexes and keys

as quickcheck is not able to pick big enough numbers for hardened indexes and gives up, see https://github.com/input-output-hk/cardano-wallet/pull/169/commits/96043b7e460c232c7793f76aa3ce954bda8c5344#diff-e9a62968e1cca775db414be6d317ecadR28  : 

```haskell
quickCheck $ withMaxSuccess 1000 prop_cannotDeriveAddressPublicKeyForHardenedIx 
*** Gave up! Passed only 17 tests.
```

quickcheck `verbose` modificator told me what is going on: quickcheck is picking small numbers and precondition `isHardened ix` fails most of the time. I have tried to use `Large` from quickcheck but it fails as well. So instead this refactor requires me to add two specialized generators for hardened and non hardened keys which should be fairly simple

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: Resolved with https://github.com/input-output-hk/cardano-wallet/pull/169/commits/ce4cedfe9fe0a878416dd89ea71a1f5ad25c75d9

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: Resolved with https://github.com/input-output-hk/cardano-wallet/pull/169/commits/ce4cedfe9fe0a878416dd89ea71a1f5ad25c75d9

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: Resolved with https://github.com/input-output-hk/cardano-wallet/pull/169/commits/ce4cedfe9fe0a878416dd89ea71a1f5ad25c75d9

KtorZ:
@akegalj This one conflicts with the previous one :no_mouth: , may you look into that please ?

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -75,6 +76,119 @@ prop_deriveAddressPublicKeyNonHardened accPubKey change (NonHardened addressIx)
...
+    -> Property
+prop_deriveAddressPrivateKeyCorrectPassword passPhrase accEncPrvKey changeChain (NonHardened addressIx) =
+    isJust (checkPassMatches passPhrase accEncPrvKey) ==> isJust addrPrvKey
+  where
+    addrPrvKey =
+        deriveAddressPrivateKey
+            passPhrase
+            accEncPrvKey
+            changeChain
+            addressIx
```
KtorZ, line 190: I am afraid that the two properties above aren't testing much :/ .. I was surprised at first that this particular one even pass without QuickCheck giving up (actually, the generator for `PassPhrase` isn't as broad as I thought). 

Still, the only case where the precondition may possibly (realistically) be satisfied is when the passphrase is empty. Chances to get an `EncryptedSecretKey` generated with a passphrase identical to `PassPhrase` are actually quite null here.

To get convinced, we can write:

```hs
    isJust (checkPassMatches passPhrase accEncPrvKey) ==>
         classify (passPhrase == mempty) "Empty PassPhrase" (property $ isJust addrPrvKey)
```

which yields:

```
succeeds if passwords are equal
      +++ OK, passed 100 tests (100% Empty PassPhrase).
```

KtorZ (changes requested):

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: :/ good catch!

I had a comment about that actually before last refactor and was also suspicious to see quickcheck passed the test - but was lazy to check the cause in this case :-1:  

Thanks for having a look - will add custom generator for this case.

akegalj: this would eventually be caught by https://github.com/input-output-hk/cardano-wallet/issues/188

KtorZ (approved):
Merging as is. We'll tackle the pending point about the test in #188

akegalj:
thanks - added a point to development plan of https://github.com/input-output-hk/cardano-wallet/issues/188

PR #170, opened 2018-12-20
Derive account private key from master private key

akegalj:
<p align="right">#41</p>

# Overview

- [x] I have added a method to derive account private key from master private key
- [x] I have property tests to cover the change

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -2,17 +2,21 @@
 
   This module provides utils to perform child key derivation for bip44:
 
-    account (parent) extended public key -> address (child) extended public key
-    extended private key -> extended private key
+    extended private key -> extended public key
+    account extended public key -> address extended public key
+    master extended private key -> account extended private key
 
 -------------------------------------------------------------------------------}
 
 module Cardano.Wallet.Kernel.Ed25519Bip44
     ( ChangeChain(..)
+    , purposeIndex
+    , coinTypeIndex
 
     -- key derivation functions
     , deriveAddressPublicKey
     , derivePublicKey
+    , deriveAccountPrivateKey
 
     -- helpers
     , isInternalChange
@@ -20,11 +24,33 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
 
 import           Universum
 
-import           Pos.Crypto (EncryptedSecretKey, PublicKey (..), encToPublic)
+import           Pos.Crypto (EncryptedSecretKey (..), PassPhrase (..),
+                     PublicKey (..), checkPassMatches, encToPublic)
 
 import           Cardano.Crypto.Wallet (DerivationScheme (DerivationScheme2),
-                     deriveXPub)
+                     deriveXPrv, deriveXPub)
 import           Test.QuickCheck (Arbitrary (..), elements)
+
+-- | Purpose is a constant set to 44' (or 0x8000002C) following the BIP43 recommendation.
+-- It indicates that the subtree of this node is used according to this specification.
+--
+-- Hardened derivation is used at this level.
+purposeIndex :: Word32
+purposeIndex = 0x8000002C
+
+-- | One master node (seed) can be used for unlimited number of independent cryptocoins
+-- such as Bitcoin, Litecoin or Namecoin. However, sharing the same space for various
+-- cryptocoins has some disadvantages.
+--
+-- This level creates a separate subtree for every cryptocoin, avoiding reusing addresses
+-- across cryptocoins and improving privacy issues.
+--
+-- Coin type is a constant, set for each cryptocoin. For Cardano this constant is set
+-- to 1815' (or 0x80000717). 1815 is the birthyear of Ada Lovelace.
+--
+-- Hardened derivation is used at this level.
+coinTypeIndex :: Word32
+coinTypeIndex = 0x80000717
 
 -- | Change chain. External chain is used for addresses that are meant to be visible
 -- outside of the wallet (e.g. for receiving payments). Internal chain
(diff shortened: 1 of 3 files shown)


src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -53,15 +84,37 @@ changeToIndex InternalChain = 1
 deriveAddressPublicKey
     :: PublicKey       -- Account Public Key
     -> ChangeChain     -- Change chain
-    -> Word32          -- Address Key Index
+    -> Word32          -- Non-hardened Address Key Index
     -> Maybe PublicKey -- Address Public Key
 deriveAddressPublicKey (PublicKey accXPub) changeChain addressIx = do
-    -- lvl4 derivation in bip44 is derivation of change chain
+    -- address index should be non-hardened
+    guard (not $ isHardened addressIx)
```
akegalj, line 86: This guard is extra as underlying `cardano-crypto` does the same check https://github.com/input-output-hk/cardano-crypto/blob/master/src/Cardano/Crypto/Wallet.hs#L169 , but I find it nice to have here because user of this module might be confused (details in the next comment)

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -53,15 +84,37 @@ changeToIndex InternalChain = 1
...
+deriveAccountPrivateKey
+    :: EncryptedSecretKey       -- Master Private Key
+    -> PassPhrase               -- Passphrase used to encrypt Master Private Key
+    -> Word32                   -- Hardened Account Key Index
+    -> Maybe EncryptedSecretKey -- Account Private Key
+deriveAccountPrivateKey masterEncPrvKey@(EncryptedSecretKey masterXPrv passHash) passPhrase accountIx = do
+    -- enforce valid PassPhrase check
+    checkPassMatches passPhrase masterEncPrvKey
+    -- account index should be hardened
+    guard (isHardened accountIx)
```
akegalj, line 108: this guard is necessary here as underlying lib doesn't enforce any checks. Account derivation should be hardened in bip44 https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account and if non-hardened/normal key index is given - we should abort.

Not to confuse users of this module we might:
 1) Leave it as it is. Account index is represented as `Word32` and `deriveAccountPrivateKey` expects it to be hardened (>=2^31). If normal key index (<2^31) is given we will return `Nothing`
 
 2) Implement advice from https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys : "To ease notation for hardened key indices, a number iH represents i+2^31". We would still use `Word32` as account key index but internally would increase this index `accountIx` by 2^31 to get hardened key. This might confuse the user of the module if he is informed and knows how bip44 works. We would also have to fail if index is larger then 2^31 as this way we exceed number of possible indices.

 3) Instead of using `Word32` lets encode derivation index as:
```haskell
newtype NormalIndex = NormalIndex Word32
newtype HardenedIndex = HardenedIndex Word32

firstNormalIndex :: NormalIndex
firstNormalIndex = NormalIndex 0

firstHardenedIndex :: HardenedIndex
firstHardenedIndex = HardenedIndex 2^31

nextNormalIndex :: NormalIndex -> Maybe NormalIndex
nextNormalIndex (NormalIndex ix) | ix >= firstHardenedIndex = Nothing
                                 | otherwise = Just $ NormalIndex (ix + 1)

nextHardenedIndex :: HardenedIndex -> Maybe HardenedIndex
nextHardenedIndex (HardenedIndex ix) | ix > maxBound :: Word32 = Nothing
                                     | otherwise = Just $ HardenedIndex (ix + 1)
```

we would then expose only smart constructors (not `DerivationIndex` constructors). We could simplify last one to 
```haskell
nextHardenedIndex :: HardenedIndex -> HardenedIndex
nextHardenedIndex (HardenedIndex ix) = HardenedIndex (ix + 1)
```
if there is no practical risk we will exceed the index (I think there is?)

As for 
```haskell
nextNormalIndex :: NormalIndex -> Maybe NormalIndex
nextNormalIndex (NormalIndex ix) | ix >= firstHardenedIndex = Nothing
                                 | otherwise = Just $ NormalIndex (ix + 1)
```
we might also drop `Maybe` if for practical reason there is no chance we will exceed index limit. Personally I find `Maybe` good here as it pushes error handling up the stack and if there is a practical risk index can be exceed then we should cover that case and inform the user (ie Daedalus user) that he had exceeded account or address index.

Having `nextNormalIndex` and `nextHardenedIndex` makes sense as well because we are implementing sequential derivation scheme (instead of current random scheme). Index bound check is made explicit this way - which enforces proper error handling and error report up the stack. `NormalIndex` and `HardenedIndex` would also serve as a good wrapper types for adding proper quickcheck `Arbitrary` generators (thus optimizing tests).

This should probably be separate task for new sequential derivation scheme (or we can create task to refactor it later).

@KtorZ thoughts ?

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
KtorZ: Yeah, ideally, we would use an appropriate type isomorphic to `Word31`. We have to shove in a `Maybe` here because the input domain we allow is too broad. If we were to enforce that `Word31`, we wouldn't need any `Maybe` at all.

KtorZ: Having said that, it's something that would ripple in a lot of places. Not sure we actually want that _now_.

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
KtorZ: That's one of those cases where it's hard to choose between a more complex type that buys us little type-safety or, a slightly less friendly API. 
In the end, the result is rather similar, we will, at some point, get a `Nothing` value. Better sooner than later indeed to prevent errors from propagating. 

Nevertheless, looking at the big pictures, those indexes directly comes from the `HdAddressIx` and `HdAccountIx` types we have in the Kernel that are also represented as `Word32` so far. Thus, to be really fair, these are actually the type we should modify to enforce valid ranges for them. 

I'd suggest here to stick with `Word32` for now, and to open a ADR giving the context and the detailed decision above. That'd be something to tackle later but I don't think it buys us much to do it now and only for this library.

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
akegalj: added point to #41 development plan

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
akegalj: added point to #41 development plan

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -1,39 +1,123 @@
...
+                     property, shrinkIntegral)
 
--- | It proves that we cannot derive address public key
--- if address index is too big. We should be able to derive
--- Address public key only with non-hardened address index
-prop_cannotDeriveAddressPublicKeyForBigIx
+-- A wrapper type for hardened keys generator
+newtype Hardened
+    = Hardened Word32
+    deriving (Show, Eq, Ord, Enum, Real, Integral, Num)
```
KtorZ, line 27: Note: when using `GeneralizzedNewtypeDeriving` always use `DerivingStrategies` and choose `deriving newtype` and `deriving stock` when appropriate to make the intent clear.

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -1,39 +1,123 @@
...
+
+-- A wrapper type for non-hardened keys generator
+newtype NonHardened
+    = NonHardened Word32
+    deriving (Show, Eq, Ord, Enum, Real, Integral, Num)
```
KtorZ, line 32: same here ^^^

KtorZ (approved):

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: I didn't know about it - will read more at https://ghc.haskell.org/trac/ghc/wiki/Commentary/Compiler/DerivingStrategies

thanks!

PR #171, opened 2018-12-20
Remove V1 wrapper: reduce usage of WalAddress in Kernel-layer.

denisshevchenko:
<p align="right">#119</p>

# Overview

- [x] Usage of `WalAddress` in Kernel-layer is reduced.

# Comments

Since `WalAddress` is defined at API-level, it shouldn't be used in Kernel-layer. Unfortunately, there is still one module `Cardano.Wallet.Kernel.Decrypt` that using `WalAddress`.

### DIFF
--- a/src/Cardano/Wallet/API/Indices.hs
+++ b/src/Cardano/Wallet/API/Indices.hs
@@ -65,6 +65,10 @@ instance ToIndex WalletAddress WalAddress where
     toIndex _ = fmap WalAddress . either (const Nothing) Just . Core.decodeTextAddress
     accessIx WalletAddress{..} = addrId
 
+instance ToIndex WalletAddress Core.Address where
+    toIndex _ = either (const Nothing) Just . Core.decodeTextAddress
+    accessIx WalletAddress{..} = let (WalAddress addr) = addrId in addr
+
 --
 -- Primary and secondary indices for V1 types
 --
@@ -138,7 +142,8 @@ type family IndexToQueryParam resource ix where
     IndexToQueryParam Wallet  WalletId                = "id"
     IndexToQueryParam Wallet  WalletTimestamp         = "created_at"
 
-    IndexToQueryParam WalletAddress (WalAddress)      = "address"
+    IndexToQueryParam WalletAddress Core.Address      = "address"
+    IndexToQueryParam WalletAddress WalAddress        = "address"
 
     IndexToQueryParam Transaction WalletTxId         = "id"
     IndexToQueryParam Transaction WalletTimestamp    = "created_at"
@@ -167,6 +172,7 @@ instance KnownQueryParam Account AccountIndex
 instance KnownQueryParam Wallet Core.Coin
 instance KnownQueryParam Wallet WalletId
 instance KnownQueryParam Wallet WalletTimestamp
+instance KnownQueryParam WalletAddress Core.Address
 instance KnownQueryParam WalletAddress WalAddress
 instance KnownQueryParam Transaction WalletTxId
 instance KnownQueryParam Transaction WalletTimestamp
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -334,6 +334,14 @@ instance FromHttpApiData WalAddress where
 instance ToHttpApiData WalAddress where
     toQueryParam (WalAddress a) = sformat build a
 
+-- | We need 'FromHttpApiData' and 'ToHttpApiData' instances for 'Core.Address',
+-- it is required for 'filterHdAddress' in 'Cardano.Wallet.WalletLayer.Kernel.Accounts'.
+instance FromHttpApiData Core.Address where
+    parseQueryParam = Core.decodeTextAddress
+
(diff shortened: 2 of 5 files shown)


src/Cardano/Wallet/API/Indices.hs
```diff
@@ -138,7 +142,8 @@ type family IndexToQueryParam resource ix where
     IndexToQueryParam Wallet  WalletId                = "id"
     IndexToQueryParam Wallet  WalletTimestamp         = "created_at"
 
-    IndexToQueryParam WalletAddress (WalAddress)      = "address"
+    IndexToQueryParam WalletAddress Core.Address      = "address"
+    IndexToQueryParam WalletAddress WalAddress        = "address"
```
KtorZ, line 146: How come do we still need `IndexToQueryParam` for `WalAddress` ?

denisshevchenko: It doesn't compile without it. AFAIU, this is because we use a filter by `WalAddress` on API-level:

```
:<|> "wallets" :> CaptureWalletId :> "accounts"
          :> CaptureAccountId :> "addresses"
          :> Summary "Retrieve only account's addresses."
          :> WalletRequestParams
          :> FilterBy '[WalAddress] WalletAddress
          :> Get '[ValidJSON] (APIResponse AccountAddresses)
```

KtorZ: Make sense :+1:

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -334,6 +334,14 @@ instance FromHttpApiData WalAddress where
 instance ToHttpApiData WalAddress where
     toQueryParam (WalAddress a) = sformat build a
 
+-- | We need 'FromHttpApiData' and 'ToHttpApiData' instances for 'Core.Address',
+-- it is required for 'filterHdAddress' in 'Cardano.Wallet.WalletLayer.Kernel.Accounts'.
+instance FromHttpApiData Core.Address where
+    parseQueryParam = Core.decodeTextAddress
+
+instance ToHttpApiData Core.Address where
+    toQueryParam = sformat build
```
KtorZ, line 343: That sounds wrong? We _shouldn't_ render `Core.Address` directly through the API, so both instances `shouldn't` be needed, unless I am missing something. 
The comment says it's needed for `filterHdAddress`, but it doesn't seem so?

denisshevchenko: > We shouldn't render Core.Address directly through the API...

Of course, but `filterHdAddress` uses `coerce` for `FilterOp`, I didn't find another solution... :-(

KtorZ: Well, let's stick with `WalAddress` in `filterHdAddress` then.

denisshevchenko: If I return `filterHdAddress` function to its previous state, i.e.:

```
filterHdAddress :: FilterOperations '[V1.WalAddress] WalletAddress
                -> FilterOperations '[V1.WalAddress] (Indexed HD.HdAddress)
filterHdAddress NoFilters               = NoFilters
filterHdAddress (FilterNop NoFilters)   = FilterNop NoFilters
filterHdAddress (FilterOp op NoFilters) = FilterOp (coerce op) NoFilters
```

I get this error:

```
    • Could not deduce (Data.IxSet.Typed.IsIndexOf V1.WalAddress '[])                                                                                                                          
        arising from a use of ‘FilterOp’
```

To fix this error, I have to add `WalAddress` in indices type-list:

```
type SecondaryIndexedHdAddressIxs = '[AutoIncrementKey, HdRootId, HdAccountId, V1.WalAddress, Core.Address]
```

in module `Cardano.Wallet.Kernel.DB.HdWallet`. And it _will_ work, but in this case we'll return to the same situation: `WalAddress` is using in Kernel-layer. :-)

KtorZ: Then, what do you suggest :) ?

denisshevchenko: Anyway, we _still_ use API-types in Kernel-layer (for example, `WalletId` in the module `Cardano.Wallet.Kernel.Decrypt`). I mean if we want to get rid of API-types in Kernel-layer _completely_ - we have to create another additional PRs, not just this one. So we have two options:

1. Merge this PR for now and move on.
2. Close this PR and create another PRs (in the _separate_ issue, not in 119) for complete removing _all_ API-types from Kernel-layer.

@KtorZ What do you think about it? Anyway, "removing V1 wrapper" issue is done.

KtorZ: Well. This PR mostly adds orphan instances on `Core.Address` ... I don't think that's a step in the right direction to be honest. So it seems `WalAddress` isn't something we can easily get rid of.

denisshevchenko: Ok, agree. We don't love orphan instances. IMHO, we should open (in some future) new ticket for removing API-types from Kernel-layer.

Issue #173, opened 2018-12-20
Ability to write raw (falsy) requests in integration tests to bypass type-safety


KtorZ:
# Context

The integration tests currently relies on a typed haskell http client when it comes to making request to the underlying cluster. This prevents us from creating malformed requests (which is the whole point of using such typed client) and hence, prevent us from testing a few "wrong" / failing interactions with the API.

# Decision

Integration tests could provide an extra constructor `rawRequest` to create a request from raw JSON, potentially malformed. Such constructor should only be used to test ill cases; for normal scenario, one should favor the typed version.

# Acceptance Criterias

1. Integration DSL _must_ provide a constructor to submit request from raw JSON object
2. Other DSL combinators _must_ remain compatible with this new constructor

---

# Development Plan

- [x] I intend to extend the `Request` class from `Framework/Request` with a `rawRequest` constructor. 
  This constructor would take an endpoint as a plain string (or by proxying a Servant endpoint) and a JSON value which could be created through quasi-quoting. 

- [x] I plan on writing a "demo" scenario using such feature to demonstrate its capability

- [ ] ~~I will also use this opportunity to add a few combinators as discussed with @piotr-iohk (assert the response status of a request and a typed-way to provide spending password).~~

# PR

| Number   | Base      |
| ---      | ---       |
| #175       | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| 1.        | Added a `unsafeRequest` function to perform a request with a "free" body. Body still has to be a valid JSON object, but that gives us quite some freedom. Also added a way to assert that the response obtained from the API is a JSON Validation error, matching a given string. An example of failure (when the string isn't matched is given below).
| 2.        | The change was almost seamless and is only added as an extra "topping", so nothing else had to change.       |

```
  1) Wallets demo unsafeRequest
       "Error in $: When parsing the record walletUpdate of type Cardano.Wallet.API.V1.Types.WalletUpdate the key assuranceLevel was not present." does not contain "key assuranceLevel was not present patate"
```

---

# Retrospective

This now makes it fairly easy to write falsy requests for integration testing which gives us the ability to test more failing cases. Some possible areas of improvements coming from the first usage:

- Having the ability to provide custom headers as well
- Being able to perform some similar general-purpose JSON matching on response bodies
- Being able to assert on the response code

PR #175, opened 2018-12-20
Add support for unsafe requests in the integration DSL

KtorZ:
<p align="right">#173</p>

# Overview

- [x] I have added an `unsafeRequest` combinator that can be used to submit raw requests to the API, specifying methods, endpoint and body by hand (can't yet customize headers, but that's an improvement we could imagine)

- [x] I have added an extra `(</>)` combinator to easily create paths containing variables from ids

- [x] I have added an extra `expectJSONError` combinator to expect a response to return a JSON validation error containing a given message.

- [x] I have added a dummy integration test example that illustrate all the above.

# Comments

We may want to remove the example in the future as it serves no real purpose :)

### DIFF
--- a/test/integration/Test/Integration/Framework/Request.hs
+++ b/test/integration/Test/Integration/Framework/Request.hs
@@ -3,21 +3,42 @@ module Test.Integration.Framework.Request
     , request
     , request_
     , successfulRequest
+    , unsafeRequest
     , ($-)
     ) where
 
-import           Universum
+import           Universum hiding (ByteString)
 
+import           Data.Aeson (FromJSON)
+import qualified Data.Aeson as Aeson
+import           Data.ByteString.Lazy (ByteString)
 import           Data.Generics.Product.Typed (HasType, typed)
+import qualified Data.Text as T
+import           GHC.Exts (IsList (..))
+import           Network.HTTP.Client (RequestBody (..), httpLbs, method,
+                     parseRequest, requestBody, requestHeaders, responseBody,
+                     responseHeaders, responseStatus, responseVersion)
+import qualified Network.HTTP.Client as HTTP
+import           Network.HTTP.Types.Method (Method)
+import           Network.HTTP.Types.Status (status200, status300)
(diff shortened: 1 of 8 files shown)


piotr-iohk (approved):
nice! :+1:

test/integration/Test/Integration/Framework/Request.hs
```diff
@@ -59,6 +80,47 @@ instance Request () where
...
+        , HasManager ctx
+        )
+    => (Method, Text)
+    -> Maybe Aeson.Value
+    -> m (Either ClientError a)
+unsafeRequest (verb, path) body = do
+    (base, man) <- view manager
+    req <- parseRequest (showBaseUrl $ base { baseUrlPath = T.unpack path })
+    res <- liftIO $ httpLbs (prepare req) man
+    return $ case responseStatus res of
```
akegalj, line 98: minor - but this looks like a good candidate for `if` (not `case`) as you used guards to be able to compare status

```haskell
let resStatus = responseStatus res
in return $
  if resStatus >= status200 && resStatus <= status300
     then...
     else ...
```

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -4,12 +4,24 @@ module Test.Integration.Scenario.Wallets
...
 
-
 spec :: Scenarios Context
 spec = do
+    scenario "demo unsafeRequest" $ do
+        fixture <- setup $ defaultSetup
+
+        let endpoint = "api" </> "v1" </> "wallets" </> fixture ^. wallet . walletId
+        response <- unsafeRequest ("PUT", endpoint) $ Just $ [json|{
+            "invalidField": #{fixture ^. wallet}
```
akegalj, line 18: useful :+1:

akegalj (approved):

test/integration/Test/Integration/Framework/Request.hs
KtorZ: Indeed. I was expecting more cases initially :sweat_smile:

PR #177, opened 2018-12-21
#88 Add applying blocks via diffusion and comment out old mechanism

paweljakubas:
#88 

# Overview

- [x] I have added getBlocks and requestTip support in diffusion layer
- [x] The additional thread is instantiated that is responsible for downloading blocks from the node
- [x] The added mechanism is based on pull paradigm, ie., the wallet is the initiator for pulling blocks, makes sure the blocks are in line with what the node has
- [x] The same resolved blocks are created using the added mechanism as it is in case of old mechanism
- [x] apply blocks part of old mechanism is commented out  
- [x] I have handled a number of pathological cases, like corrupt data coming from the node or no communication with the node for some time 
- [x] Added `pull-mechanism` flag to switch on pull mechanism. Otherwise, push mechanism, the readerT based - well tested but introducing coupling, is used 

# Comments

The thorough description of the proposed mechanism are
[here](https://github.com/input-output-hk/cardano-wallet/wiki/ApplyingBlocks-Rollback-After-Decoupling-Byron) 

# Testing
## unit tests
When run as it is push mechanism is used as default. In order to run unit tests in **pull mechanism** first change in `test/unit/Test/Spec/Fixture.hs`:
```
41  usePullMechanism :: Bool                                                                                                                                                                  
42  usePullMechanism = True
```
and then 
```
$ stack test cardano-wallet:test:unit
```

## integration tests
In order to run integration tests using pull mechanism first change in `test/integration/Test/Integration/Framework/Cluster.hs` :
```
72      , ("PULL_MECHANISM", "True")
``` 
and then
```
$ stack test cardano-wallet:test:integration
```
One should see the following log in `edge.log.pub`:
```
[pawel@arch logs]$ ag "mechanism" edge.log.pub
43602:[cardano-sl.edge:Debug:ThreadId 1063] [2019-01-26 19:07:54.34 UTC] pull mechanism (diffusion layer based) for block syncing is active
```
In default case one should see instead : 
 ```
[pawel@arch logs]$ ag "mechanism" edge.log.pub
538:[cardano-sl.edge:Debug:ThreadId 1163] [2019-01-26 18:57:03.74 UTC] push mechanism (readerT based) for block syncing is active
```

## wallet-node syncing with mainnet
When running wallet with node attached to mainnet see [here](https://github.com/input-output-hk/cardano-wallet/wiki/Building-Running-and-Testing-a-wallet-node)
In order to switch on **pull mechanism** of block syncing use additionally `--pull-mechanism```
One should see the same logs as above confirming switching on the pull mechanism

A number of problems were solved during this PR:
**Update1**
Now after the metaTx fix coin selection tests in Transactions testsuite and Utxo test in Wallets testsuite are alright. Occasionally, I am witnessing the thread responsible for block application to be stuck (and then all tests that rely on block application fail) for the reason I do not understand at this moment. 

**Update 2**
I addressed those rare cases : it was due to (a) no communication with node, (b) corrupt data flowing from the node, (c) combination of (a) and (b). 
I tested using integration tests more than 100 - found all cases and handled them properly (ie., integration tests pass)
 
Very occasionally the node response of `requestTip` (from `lib/src/Pos/Diffusion/Full/Block.hs` from `cardano-sl`): 
```
479	    handleTip _ t = do
480	        traceWith logTrace (Warning, sformat ("requestTip: got enexpected response: "%shown) t)
481	        throwIO $ DialogUnexpected "peer sent more than one tip"
```

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/BlockContext.hs
+++ b/src/Cardano/Wallet/Kernel/DB/BlockContext.hs
@@ -46,7 +46,7 @@ data BlockContext = BlockContext {
       -- we do some work to figure out what the previous /main/ block was.
       -- See 'mostRecentMainBlock'.
     , _bcPrevMain :: !(Strict.Maybe (InDb Core.HeaderHash))
-    } deriving Eq
+    } deriving (Eq, Show)
 
 makeLenses ''BlockContext
 deriveSafeCopy 1 'base ''BlockContext
--- a/src/Cardano/Wallet/Kernel/DB/Resolved.hs
+++ b/src/Cardano/Wallet/Kernel/DB/Resolved.hs
@@ -61,7 +61,7 @@ data ResolvedTx = ResolvedTx {
 
      -- | Transaction Meta
     , _rtxMeta    :: InDb (TxId, Timestamp)
-    }
+    } deriving Show
 
 -- | (Unsigned block) containing resolved transactions
 --
@@ -77,7 +77,7 @@ data ResolvedBlock = ResolvedBlock {
 
       -- | Creation time of this block
     , _rbMeta    :: !Timestamp
-    }
+    } deriving Show
 
 makeLenses ''ResolvedTx
 makeLenses ''ResolvedBlock
--- a/src/Cardano/Wallet/Kernel/Diffusion.hs
+++ b/src/Cardano/Wallet/Kernel/Diffusion.hs
@@ -1,17 +1,29 @@
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 module Cardano.Wallet.Kernel.Diffusion (
     WalletDiffusion(..)
   , fromDiffusion
+  , tickDiffusionLayer
   ) where
 
 import           Universum
 
-import           Pos.Chain.Txp (TxAux)
+import           Control.Concurrent (threadDelay)
+import qualified Data.Map.Strict as Map
+import           Formatting (sformat, (%))
+import qualified Formatting as F
+
+import           Pos.Chain.Block (Block, BlockHeader, HeaderHash, prevBlockL)
+import           Pos.Chain.Txp (TxAux, Utxo)
 import           Pos.Core ()
+import           Pos.Core.Chrono (OldestFirst (..))
+import           Pos.DB.Block (MonadBlockVerify)
 import           Pos.Infra.Communication.Types.Protocol (NodeId)
 import           Pos.Infra.Diffusion.Subscription.Status (SubscriptionStatus,
                      ssMap)
 import           Pos.Infra.Diffusion.Types
+import           Pos.Util.Wlog (Severity (..))
+
 
 -- | Wallet diffusion layer
 --
@@ -33,13 +45,90 @@ data WalletDiffusion = WalletDiffusion {
 
       -- | Get subscription status (needed for the node settings endpoint)
     , walletGetSubscriptionStatus :: IO (Map NodeId SubscriptionStatus)
+
+      -- | Request tip-of-chain from the network
+    , walletRequestTip            :: IO (Map NodeId (IO BlockHeader))
+
+      -- | Get blocks from checkpoint up to the header (not-included) from the node
+    , walletGetBlocks
+        :: NodeId
+        -> HeaderHash
+        -> [HeaderHash]
+        -> IO (OldestFirst [] Block)
+
     }
 
 -- | Extract necessary functionality from the full diffusion layer
-fromDiffusion :: (forall a. m a -> IO a)
+fromDiffusion :: forall m ctx .
+                 (MonadBlockVerify ctx m)
+              => (forall a. m a -> IO a)
               -> Diffusion m
               -> WalletDiffusion
 fromDiffusion nat d = WalletDiffusion {
       walletSendTx                = nat . sendTx d
     , walletGetSubscriptionStatus = readTVarIO $ ssMap (subscriptionStates d)
+    , walletRequestTip            = do
+           fromDiff <- nat $ requestTip d
+           pure $ Map.map nat fromDiff
+    , walletGetBlocks             = \nodeId from toNotIncluding -> do
+            case toNotIncluding of
+                [lastConsumedHeader] ->
+                    nat $ getPrevBlock nodeId from (Just lastConsumedHeader) []
+                _ ->
+                    nat $ getPrevBlock nodeId from Nothing []
     }
+    where
+        getPrevBlock
+            :: NodeId
+            -> HeaderHash
+            -> Maybe HeaderHash
+            -> [Block]
+            -> m (OldestFirst [] Block)
+        getPrevBlock nodeId from toMaybe blocksInThisBatch = do
+            blocks <- getOldestFirst <$> getBlocks d nodeId from [from]
+            case blocks of
+                [block@(Right _)] -> do
+                    -- here we are dealing with MainBlock
+                    let prevBlockHeader = block ^. prevBlockL
+                    case toMaybe of
+                        Just upToHeader ->
+                            if (prevBlockHeader /= upToHeader) then
+                                -- still need to download backwards
+                                getPrevBlock nodeId prevBlockHeader toMaybe (block : blocksInThisBatch)
+                            else
+                                -- we just downloaded the last missing block
+                                pure $ OldestFirst $ block : blocksInThisBatch
+                        Nothing -> do
+                            -- just one block downloaded
+                            pure $ OldestFirst (block : blocksInThisBatch)
+                _ -> do
+                    pure $ OldestFirst blocksInThisBatch
+
+
+tickDiffusionLayer
+    :: forall m. (MonadCatch m, MonadIO m)
+    => (Severity -> Text -> m ())
+       -- ^ A logging function
+    -> ( (Severity -> Text -> m ()) -> (([HeaderHash],[HeaderHash]),Utxo) -> m (([HeaderHash],[HeaderHash]), Utxo) )
+       -- ^ A function to call at each 'tick' of the worker.
+       -- This callback will be responsible for doing any pre
+       -- and post processing of the state.
+    -> (([HeaderHash],[HeaderHash]),Utxo)
+    -> m ()
+tickDiffusionLayer logFunction tick initialState =
+    go initialState
+    `catch`
+    (\(e :: SomeException) ->
+            let msg = "Terminating tickDiffusionLayer due to " % F.shown
+            in logFunction Error (sformat msg e)
+    )
+    where
+      go :: (([HeaderHash],[HeaderHash]), Utxo) -> m ()
+      go previousState  = do
+          logFunction Debug "ticking the slot in the diffusion layer..."
+          currentState <- tick logFunction previousState
+          liftIO $ threadDelay tickDiffusionRate
+          go currentState
+
+tickDiffusionRate :: Int
+tickDiffusionRate = 1000000
--- a/src/Cardano/Wallet/Kernel/Submission/Worker.hs
+++ b/src/Cardano/Wallet/Kernel/Submission/Worker.hs
@@ -11,14 +11,15 @@ import qualified Formatting as F
 import           Pos.Util.Wlog (Severity (..))
 
 
-tickSubmissionLayer :: forall m. (MonadCatch m, MonadIO m)
-                    => (Severity -> Text -> m ())
-                    -- ^ A logging function
-                    -> IO ()
-                    -- ^ A function to call at each 'tick' of the worker.
-                    -- This callback will be responsible for doing any pre
-                    -- and post processing of the state.
-                    -> m ()
+tickSubmissionLayer
+    :: forall m. (MonadCatch m, MonadIO m)
+    => (Severity -> Text -> m ())
+       -- ^ A logging function
+    -> IO ()
+       -- ^ A function to call at each 'tick' of the worker.
+       -- This callback will be responsible for doing any pre
+       -- and post processing of the state.
+    -> m ()
 tickSubmissionLayer logFunction tick =
     go `catch` (\(e :: SomeException) ->
                    let msg = "Terminating tickSubmissionLayer due to " % F.shown
--- a/src/Cardano/Wallet/Server/Plugins.hs
+++ b/src/Cardano/Wallet/Server/Plugins.hs
@@ -60,6 +60,8 @@ import           Cardano.Wallet.WalletLayer (ActiveWalletLayer,
                      PassiveWalletLayer)
 import qualified Cardano.Wallet.WalletLayer.Kernel as WalletLayer.Kernel
 
+import           Data.Reflection (Given (..))
+import           Pos.Chain.Ssc (SscConfiguration, sscConfiguration)
 import           Pos.Infra.Diffusion.Types (Diffusion (..))
 import           Pos.Infra.Shutdown (HasShutdownContext (shutdownContext),
                      ShutdownContext)
@@ -69,6 +71,10 @@ import           Pos.Util.Wlog (logError, logInfo, modifyLoggerName,
                      usingLoggerName)
 import           Pos.Web (serveDocImpl, serveImpl)
 import qualified Pos.Web.Server
+
+
+instance Given SscConfiguration where
+    given = sscConfiguration
 
 -- A @Plugin@ running in the monad @m@.
 type Plugin m = Diffusion m -> m ()
@@ -104,7 +110,7 @@ apiServer
             <> show err
         Right _ -> do
             logInfo "The node responded successfully."
-    WalletLayer.Kernel.bracketActiveWallet passiveLayer passiveWallet diffusion' $ \active _ -> do
+    WalletLayer.Kernel.bracketActiveWallet passiveLayer passiveWallet diffusion' walletApplyBlockPullMechanism $ \active _ -> do
         ctx <- view shutdownContext
         serveImpl
             (getApplication active)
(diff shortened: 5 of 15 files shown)

### CODE REVIEW

File: src/Cardano/Wallet/WalletLayer/Kernel.hs [hunk 1]
```diff
@@ -82,12 +103,26 @@ bracketPassiveWallet pm mode logFunction keystore node fInjects f = do
 
       -- Start the wallet worker
       let wai = Actions.WalletActionInterp
-                 { Actions.applyBlocks = \blunds -> do
+                 --{ Actions.applyBlocks = \blunds -> do
+                { Actions.applyBlocks = \_ -> do
+                    return ()
```
paweljakubas: Commenting out readerT mechanism

src/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -64,6 +66,10 @@ import           Pos.Util.Wlog (logInfo, modifyLoggerName, usingLoggerName)
 import           Pos.Web (serveDocImpl, serveImpl)
 import qualified Pos.Web.Server
 
+
+instance Given SscConfiguration where
+    given = sscConfiguration
+
```
paweljakubas, line 78: need this in order to use getBlocks and service MonadBlockVerify constraint

src/Cardano/Wallet/Kernel/Submission/Worker.hs
```diff
@@ -43,3 +46,29 @@ tickSubmissionLayer logFunction tick =
...
+
+
+tickDiffusionLayer :: forall m. (MonadCatch m, MonadIO m)
+                   => (Severity -> Text -> m ())
+                      -- ^ A logging function
+                   -> ((([HeaderHash],[HeaderHash]),Utxo) -> m (([HeaderHash],[HeaderHash]), Utxo))
+                      -- ^ A function to call at each 'tick' of the worker.
+                      -- This callback will be responsible for doing any pre
+                      -- and post processing of the state.
+                   -> (([HeaderHash],[HeaderHash]),Utxo)
```
paweljakubas: This will probably be different types as we will switch to relying on wallet state (for Utxo, and block headers accommodated)

src/Cardano/Wallet/Kernel/DB/BlockContext.hs
```diff
@@ -46,7 +46,7 @@ data BlockContext = BlockContext {
       -- we do some work to figure out what the previous /main/ block was.
       -- See 'mostRecentMainBlock'.
     , _bcPrevMain :: !(Strict.Maybe (InDb Core.HeaderHash))
-    } deriving Eq
+    } deriving (Eq, Show)
```
paweljakubas, line 49: needed this to print resolved blocks - will clean this in the final version

src/Cardano/Wallet/Kernel/DB/Resolved.hs
```diff
@@ -105,7 +105,7 @@ data ResolvedBlock = ResolvedBlock {
 
       -- | Creation time of this block
     , _rbMeta    :: !Timestamp
-    }
+    } deriving Show
```
paweljakubas, line 80: needed this to print resolved blocks - will clean this in the final version

src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -165,9 +200,23 @@ bracketActiveWallet
     -> (ActiveWalletLayer n -> Kernel.ActiveWallet -> m a) -> m a
 bracketActiveWallet walletPassiveLayer passiveWallet walletDiffusion runActiveLayer =
     Kernel.bracketActiveWallet passiveWallet walletDiffusion $ \w -> do
+
+        genesisConfig <- liftIO $ getCoreConfig $ passiveWallet ^. Kernel.walletNode
+        let initialUtxo = ccUtxo $ initCardanoContext genesisConfig
```
paweljakubas: I have to do it in order to have initial Utxo - in the readerT mechanism which relies on blunds it comes with in Undo value

src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -82,12 +103,26 @@ bracketPassiveWallet pm mode logFunction keystore node fInjects f = do
...
+                    --let bs = map (\(_,u)-> u) $ toList (getOldestFirst blunds) ---
+                    --let bs = map (\(b,_)-> (headerHash . getBlockHeader) b) $ toList (getOldestFirst blunds) ---
+                    --let mp = Debug.trace (("####### WalletActionInterp:\n" <> show bs)::String) $ catMaybes ls
+                    {--let resolved = map (\b -> case b of
+                                               (Just block) -> _rbTxs block
+                                               Nothing      -> []
+                                       ) ls--}
+                    let mp = Debug.trace (("####### WalletActionInterp:\n" <> show ls)::String) $ catMaybes ls
+                    --let mp = catMaybes ls
                     mapM_ (Kernel.applyBlock w) mp
```
paweljakubas: here it was the place where BListener applyBlocks was invoked

src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -165,9 +200,23 @@ bracketActiveWallet
...
+
+        genesisConfig <- liftIO $ getCoreConfig $ passiveWallet ^. Kernel.walletNode
+        let initialUtxo = ccUtxo $ initCardanoContext genesisConfig
+
+        applyingBlockTicker <- liftIO $ async $
+            tickDiffusionLayer
+            (passiveWallet ^. Kernel.walletLogMessage)
+            tickDiffusionFunction
+            (([],[]), initialUtxo)
+
```
paweljakubas: Instantiating dedicated diffusion ticker thread

src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -6,13 +7,19 @@ module Cardano.Wallet.Kernel.Diffusion (
 
 import           Universum
 
+import qualified Data.Map.Strict as Map
+
+import           Pos.Chain.Block (Block, BlockHeader, HeaderHash, prevBlockL)
 import           Pos.Chain.Txp (TxAux)
 import           Pos.Core ()
+import           Pos.Core.Chrono (OldestFirst (..))
+import           Pos.DB.Block
```
KtorZ: Favor explicit imports over exposing all identifiers :+1:

src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
```
KtorZ: I guess this will be removed later :upside_down_face:

src/Cardano/Wallet/Kernel/DB/Resolved.hs
KtorZ: Can leave it there actually. Always good to have a little debug capability.

src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -33,13 +40,76 @@ data WalletDiffusion = WalletDiffusion {
...
+                            case toMaybe of
+                                Just upToHeader ->
+                                    if (prevBlockHeader /= upToHeader) then
+                                        -- still need to download backwards
+                                        getPrevBlock nodeId prevBlockHeader toMaybe (block : blocksInThisBatch)
+                                    else
+                                    -- we just downloaded the last missing block
+                                        pure $ OldestFirst $ block : blocksInThisBatch  --
+                                Nothing -> do
+                                    getPrevBlock nodeId prevBlockHeader Nothing (block : blocksInThisBatch)  --
```
KtorZ: What is the flow exactly here :thinking: ? We download blocks going backward from a given a point until we reach either a certain point or, the very first block (when `toMaybe = Nothing`) ?

src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -33,13 +40,76 @@ data WalletDiffusion = WalletDiffusion {
...
+                        [(Left _)] -> do
+                            -- here we come to the GenesisBlock, we stop and we do not want to include it in a batch
+                            --void $ print "Left"
+                            pure $ OldestFirst blocksInThisBatch --
+                        _ -> do
+                            -- unexpected so returning current blocks in the batch
+                            --void $ print "N"
+                            pure $ OldestFirst blocksInThisBatch --
```
KtorZ: Is this really unexpected, or is it simply the terminal case of the recursion, once there's no more blocks to retrieve?

src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -33,13 +40,76 @@ data WalletDiffusion = WalletDiffusion {
...
+                    -> m (OldestFirst [] Block)
+                getPrevBlock nodeId from toMaybe blocksInThisBatch = do
+                    --void $ print "*********"
+                    --void $ print from
+                    --void $ print toMaybe
+                    --void $ print "*********"
+                    downloadedBlock <- getBlocks d nodeId from [from]
+                    let downloadedBlockStripped = getOldestFirst downloadedBlock
+                    --void $ print toMaybe
+                    case downloadedBlockStripped of
```
KtorZ:
```hs
blocks <- getOldestFirst <$> getBlocks d nodeId from [from]
case blocks of 
```

src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -33,13 +40,76 @@ data WalletDiffusion = WalletDiffusion {
...
+                        [(Left _)] -> do
+                            -- here we come to the GenesisBlock, we stop and we do not want to include it in a batch
+                            --void $ print "Left"
+                            pure $ OldestFirst blocksInThisBatch --
+                        _ -> do
+                            -- unexpected so returning current blocks in the batch
+                            --void $ print "N"
+                            pure $ OldestFirst blocksInThisBatch --
+            in
+            \nodeId from toNotIncluding -> do
```
KtorZ: I'd write this the other way around to make things a bit clearer:

```hs
walletGetBlocks nodeId from toNotIncluding = do
    ...
  where
    getPrevBlocks :: ...
```

src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -33,13 +40,76 @@ data WalletDiffusion = WalletDiffusion {
...
+                 case toNotIncluding of
+                     [lastConsumedHeader] -> do
```
KtorZ: Why is this a list if we only care about this being a singleton or.. something else? Wouldn't a `Maybe` better reflect the intent here?

src/Cardano/Wallet/Server/Plugins.hs
KtorZ: We will find a way around this :) ... Having it passed as an extra parameter isn't a bad choice. The reflection stuff is... not pragmatic.

src/Cardano/Wallet/Kernel/Submission/Worker.hs
```diff
@@ -43,3 +46,29 @@ tickSubmissionLayer logFunction tick =
 -- section 10.4 should be updated.
 tickRate :: Int
 tickRate = 5000000
+
+
+tickDiffusionLayer :: forall m. (MonadCatch m, MonadIO m)
+                   => (Severity -> Text -> m ())
```
KtorZ: Please, no more variable-length indentation :(

src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -180,3 +229,90 @@ bracketActiveWallet walletPassiveLayer passiveWallet walletDiffusion runActiveLa
         , redeemAda          = Active.redeemAda        w
         , getNodeInfo        = Info.getNodeInfo        w
         }
+
+    tickDiffusionFunction
+        :: (([HeaderHash], [HeaderHash]), Utxo)
+        -> IO (([HeaderHash], [HeaderHash]), Utxo)
+    tickDiffusionFunction ((inProgressHeader, consumedHeader),utxo) = do
```
KtorZ: This sounds like something that should be in `Kernel/Diffusion` instead.

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -7,9 +7,31 @@ import           Universum
...
+            & initialCoins .~ [1000000]
+
+        response <- request $ Client.postTransaction $- Payment
+            (defaultSource fixture)
+            (defaultDistribution 14 fixture)
+            defaultGroupingPolicy
+            noSpendingPassword
+
+        --need this to see transition between epochs
+        liftIO $ threadDelay $ 120 * oneSecond
```
KtorZ: I suggest introducing a dedicated DSL extension for this `expectNextEpochEventually` or something like this.

src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -180,3 +229,90 @@ bracketActiveWallet walletPassiveLayer passiveWallet walletDiffusion runActiveLa
...
+                --void $ print resolvedBlocks
+
+                mapM_ (Kernel.applyBlock passiveWallet) resolvedBlocks
+
+                --void $ print "-------------- end tick 1 "
+                let utxoToAdd = map (snd . snd) $ catMaybes txPayload
+                let utxoAfterAddition = utxoUnions $ utxo : utxoToAdd
+                let utxoSetToDel = Set.fromList $ concat $ map fst $ catMaybes txPayload
+                let updatedUtxo = utxoRemoveInputs utxoAfterAddition utxoSetToDel
+                pure $ (([h1], (reverse accomodatedHeaders) ++ consumedHeader), updatedUtxo)
```
KtorZ: Wow. That's one big function which I honestly didn't try to read much :( ... Sounds like this would deserve to be clarified and decomposed a bit in smaller pieces.

KtorZ (changes requested):

src/Cardano/Wallet/Kernel/Diffusion.hs
paweljakubas: yes, I need this for temporary logging

src/Cardano/Wallet/Kernel/Diffusion.hs
paweljakubas: We use `getPrevBlock nodeId from toMaybe blocksInThisBatch` here. Case when `toMaybe = Nothing` is used refers to the initial situation when up till now no block was downloaded. In that case we are expected to stop when genesis block is reached. In `getPrevBlock` we download the block having `from` header and add it to blocksInThisBatch. We stop when the previous header of just downloaded block is equal to `toMaybe` (if present) or when (if `toMaybe` is absent) genesis block is found. We use for the detection of this the fact that `Block` is either `MainBlock` or `GenesisBlock`. The GenesisBlock is not included to  `blocksInThisBatch`.  In the another case we take the previous header of the just downloaded block and call `getPrevBlock` with it.

src/Cardano/Wallet/Kernel/Diffusion.hs
paweljakubas: I agree this is superfluous. Simplified

src/Cardano/Wallet/Kernel/Diffusion.hs
paweljakubas: If servicing applying blocks only using `Maybe` would be the best. But, I would like to reuse this call in case of rollback, and in that case I will deliver here **N lastly downloaded headers**

src/Cardano/Wallet/Kernel/Diffusion.hs
paweljakubas: done

src/Cardano/Wallet/Kernel/Diffusion.hs
paweljakubas: done

src/Cardano/Wallet/Kernel/Submission/Worker.hs
paweljakubas: done

src/Cardano/Wallet/WalletLayer/Kernel.hs
paweljakubas: done

test/integration/Test/Integration/Scenario/Transactions.hs
paweljakubas: I will do it in the separate PR

src/Cardano/Wallet/WalletLayer/Kernel.hs
paweljakubas: simplified

paweljakubas:
after successful addressing review remarks I will rebase branch against develop and squash the commits

src/Cardano/Wallet/Server/Plugins.hs
paweljakubas: I will leave it for another PR

paweljakubas:
rebased

paweljakubas:
rebased and added flag pull-mechanism to switch on this mechanism. Otherwise it readerT mechanism, push one, well tested but introducing coupling, is used

paweljakubas:
When ok, I will squash the commit

KtorZ (approved):
Merging this in a side branch for now. We still need to experiment on that and, we won't be releasing that into 1.6 anyway. Still, it doesn't make sense to keep a PR / ticket hanging for months.

Issue #172, opened 2018-12-20
Integration Tests Coverage: Wallets (Part 1)


KtorZ:
# Context

Our current coverage from integration tests is quite low and could be greatly improved. We now have the power to write many test scenarios rather easily and increase the code coverage. The [coverage reports](https://coveralls.io/github/input-output-hk/cardano-wallet) generated by the CI may help to identify cases where coverage is missing.

# Decision

We will identify (visible) wallet capabilities from the API perspective and, for each of those capabilities, define and qualify requirements. Then, We'll increase integration tests coverage by verifying that those requirements are indeed met.

# Acceptance Criterias

1. We _must_ establish a list of the Wallets endpoints capabilities
1. For each of those capability, we _must_ establish a list of requirements 
1. We _must_ cover those requirements with integration tests
1. We _should_ provide a reference ID for each requirement and use them to identify corresponding tests

---

# Development Plan

cf [Wallet Capabilities](https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk), category 'Wallets'

# PR

| Number   | Base      |
| ---      | ---       |
| #178  | `develop` |
| #206 | `develop` | 
| #217 | `develop` | 
| #219 | `develop`

# QA 

| Criteria | Coverage |
| ---      | ---      |
| [WALLETS_CREATE]        | Capabilities related to wallet creation are now (partially) covered as part of the integration tests and clearly identified with a requirement number. See [Test/Integration/Scenario/Wallets.hs](https://github.com/input-output-hk/cardano-wallet/blob/e76921af79e1189d17867c2336112d7bfc76803f/test/integration/Test/Integration/Scenario/Wallets.hs)     |
| [WALLETS_LIST]        | Capabilities related to wallet listing are now (partially) covered as part of the integration tests and clearly identified with a requirement number. See [Test/Integration/Scenario/Wallets.hs](https://github.com/input-output-hk/cardano-wallet/blob/e76921af79e1189d17867c2336112d7bfc76803f/test/integration/Test/Integration/Scenario/Wallets.hs)     |

Some items couldn't be properly covered because of discovered bugs, cf:

- #193 
- #213 
- #214
- #220

---

# Retrospective

KtorZ:
I am going to reduce the scope of this one and close it _now_. 
[...]

PR #178, opened 2018-12-21
Integration tests covering WALLETS_CREATE capability

piotr-iohk:
<p align="right">#172</p>

# Overview

- [x] Added integration tests covering WALLETS_CREATE capability from [Capabilities & Requirements](https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk) document
- [x] Failed attempt to create `expectJSONSuccess` to be used similar as as `expectJSONError`. But will go back to that later (in the next PR).

# Comments
Method mentioned above that I was not able to add:
```
-- Would be nice to have this method. The one commented above does not compile.
-- `T.unpack` is `Text - > String` and  a is `a rigid type variable bound by the type signature  for: expectJSONSuccess...`
expectJSONSuccess
	:: (MonadIO m, MonadFail m, Show a)
	=> String
	-> Either ClientError a
	-> m ()
expectJSONSuccess excerpt = \case
	Left e  -> wantedSuccessButError e
	Right a -> T.unpack a `shouldContain` excerpt
```

### DIFF
--- a/test/integration/Test/Integration/Framework/DSL.hs
+++ b/test/integration/Test/Integration/Framework/DSL.hs
@@ -14,6 +14,7 @@ module Test.Integration.Framework.DSL
     , setup
     , request
     , request_
+    , successfulRequest
     , unsafeRequest
     , verify
 
@@ -69,9 +70,12 @@ module Test.Integration.Framework.DSL
     , initialCoins
     , mnemonicWords
     , wallet
+    , wallets
     , walletId
     , walletName
     , json
+    , hasSpendingPassword
+    , mkBackupPhrase
     ) where
 
 import           Universum hiding (getArgs, second)
@@ -298,8 +302,14 @@ initialCoins =
 mnemonicWords :: HasType [Text] s => Lens' s [Text]
 mnemonicWords = typed
 
+hasSpendingPassword :: HasType Bool s => Lens' s Bool
+hasSpendingPassword = typed
+
 wallet :: HasType Wallet s => Lens' s Wallet
 wallet = typed
+
+wallets :: HasType [Wallet] s => Lens' s [Wallet]
+wallets = typed
 
 walletId :: HasType WalletId s => Lens' s WalletId
 walletId = typed
@@ -329,7 +339,6 @@ expectFieldEqual getter a = \case
     Left e  -> wantedSuccessButError e
     Right s -> view getter s `shouldBe` a
 
-
 -- | Expects entire equality of two types
 expectEqual
     :: (MonadIO m, MonadFail m, Show a, Eq a)
@@ -338,7 +347,6 @@ expectEqual
     -> m ()
 expectEqual =
     expectFieldEqual id
-
 
 -- | Expect an errored response, without any further assumptions
 expectError
@@ -475,7 +483,6 @@ expectJSONError excerpt = \case
         T.unpack msg `shouldContain` excerpt
     Left e ->
         fail $ "expectJSONError: got something else than a JSON validation failure: " <> show e
-
 
 expectWalletUTxO
     :: (MonadIO m, MonadFail m)
--- a/test/integration/Test/Integration/Scenario/Wallets.hs
+++ b/test/integration/Test/Integration/Scenario/Wallets.hs
@@ -10,35 +10,792 @@ import           Test.Integration.Framework.DSL
 
 spec :: Scenarios Context
 spec = do
-    scenario "demo unsafeRequest" $ do
-        fixture <- setup $ defaultSetup
 
-        let endpoint = "api" </> "v1" </> "wallets" </> fixture ^. wallet . walletId
-        response <- unsafeRequest ("PUT", endpoint) $ Just $ [json|{
-            "invalidField": #{fixture ^. wallet}
-        }|]
+    scenario "WALLETS_CREATE_01 - One can restore previously deleted wallet" $ do
+      fixture <- setup $ defaultSetup
+        & initialCoins .~ [1000111]
 
-        verify (response :: Either ClientError Wallet)
-            [ expectJSONError "key assuranceLevel was not present"
-            ]
+      -- 1. Delete wallet
+      successfulRequest $ Client.deleteWallet
+        $- (fixture ^. wallet . walletId)
 
-    scenario "wallet is available after it's been created" $ do
-        fixture <- setup $ defaultSetup
-            & walletName .~ "漢patate字"
-            & assuranceLevel .~ StrictAssurance
+      -- 1. Restore wallet
+      restoration <- request $ Client.postWallet $- NewWallet
+        (fixture ^. backupPhrase)
+        noSpendingPassword -- defaultSpendingPassword does not work
+        StrictAssurance
+        kanjiPolishWalletName
+        RestoreWallet
+      verify restoration
+        [ expectWalletEventuallyRestored
+        , expectFieldEqual walletName kanjiPolishWalletName
+        , expectFieldEqual assuranceLevel StrictAssurance
+        , expectFieldEqual hasSpendingPassword False
+        , expectFieldEqual amount 1000111
+        ]
 
-        response <- request $ Client.getWallet
-            $- (fixture ^. wallet . walletId)
+    scenario "WALLETS_CREATE_02a - Restore wallet with No spending password" $ do
+      -- 1. Setup new wallet
+      fixture <- setup $ defaultSetup
 
-        verify response
-            [ expectFieldEqual walletName "漢patate字"
-            , expectFieldEqual assuranceLevel StrictAssurance
-            ]
+      -- 2. Delete wallet
+      successfulRequest $ Client.deleteWallet
+        $- (fixture ^. wallet . walletId)
 
-    scenario "updating a wallet persists the update" $ do
+      -- 3. Restore wallet with noSpendingPassword
+      restoreResp <- request $ Client.postWallet $- NewWallet
+        (fixture ^. backupPhrase)
+        noSpendingPassword
+        StrictAssurance
+        kanjiPolishWalletName
+        RestoreWallet
+      verify restoreResp
+        [ expectWalletEventuallyRestored
+        , expectFieldEqual walletName kanjiPolishWalletName
+        , expectFieldEqual assuranceLevel StrictAssurance
+        , expectFieldEqual hasSpendingPassword False
+        ]
+
+      -- 4. Get wallet details
+      getResp <- request $ Client.getWallet
+        $- (fixture ^. wallet . walletId)
+      verify getResp
+        [
+          expectFieldEqual walletName kanjiPolishWalletName
+        , expectFieldEqual assuranceLevel StrictAssurance
+        , expectFieldEqual hasSpendingPassword False
+        ]
+
+    scenario "WALLETS_CREATE_02b - Create wallet with No spending password" $ do
+      -- 1. Create yet new wallet with noSpendingPassword
+      bkpPhrase <- mkBackupPhrase testBackupPhrase2
+      let newWalletName = "πœę©ß←↓→óþąśðæŋ’ə…łżźć„”ńµ"
+      createResp <- successfulRequest $ Client.postWallet $- NewWallet
+        bkpPhrase
+        noSpendingPassword
+        NormalAssurance
+        newWalletName
+        CreateWallet
+      verify (Right createResp)
+        [
+          expectFieldEqual walletName newWalletName
+        , expectFieldEqual assuranceLevel NormalAssurance
+        , expectFieldEqual hasSpendingPassword False
+        ]
+
+      -- 2. Verify noSpendingPassword on new wallet
+      getAgainResp <- request $ Client.getWallet
+        $- (createResp ^. walletId)
+      verify getAgainResp
+        [
+          expectFieldEqual walletName newWalletName
+        , expectFieldEqual assuranceLevel NormalAssurance
+        , expectFieldEqual hasSpendingPassword False
+        ]
+
+    scenario "WALLETS_CREATE_02c - Create wallet with spending password" $ do
+
+      createResp <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "create",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "My HODL Wallet",
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
+      }|]
+      verify (createResp :: Either ClientError Wallet)
+        [
+          expectFieldEqual hasSpendingPassword True
+        , expectFieldEqual walletName "My HODL Wallet"
+        , expectFieldEqual assuranceLevel NormalAssurance
+        ]
+
+    scenario "WALLETS_CREATE_02d - Restore wallet with spending password" $ do
+
+      restoResp <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "restore",
+        "backupPhrase": #{testBackupPhrase2},
+        "assuranceLevel": "strict",
+        "name": "My HODL Wallet 2",
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
+      }|]
+
+      verify (restoResp :: Either ClientError Wallet)
+        [
+          expectWalletEventuallyRestored
+        , expectFieldEqual hasSpendingPassword True
+        , expectFieldEqual walletName "My HODL Wallet 2"
+        , expectFieldEqual assuranceLevel StrictAssurance
+        ]
+
+    scenario "WALLETS_CREATE_02e - Create wallet if spendingPassword is null" $ do
+
+      response3 <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "create",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "MyFirstWallet",
+        "spendingPassword": null
+      }|]
+      verify (response3 :: Either ClientError Wallet)
+        [
+          expectFieldEqual walletName "MyFirstWallet"
+        , expectFieldEqual assuranceLevel NormalAssurance
+        , expectFieldEqual hasSpendingPassword False
+        ]
+
+
+    scenario "WALLETS_CREATE_02f - Restore wallet if spendingPassword is null" $ do
+
+      response3 <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "restore",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "strict",
+        "name": "MyFirstWallet",
+        "spendingPassword": null
+      }|]
+      verify (response3 :: Either ClientError Wallet)
+        [
+          expectFieldEqual walletName "MyFirstWallet"
+        , expectFieldEqual assuranceLevel StrictAssurance
+        , expectFieldEqual hasSpendingPassword False
+        ]
+
+    scenario "WALLETS_CREATE_03a - Cannot create wallet if spendingPassword is hex|base16 string with length less than 64" $ do
+
+      response <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "create",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "MyFirstWallet",
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c6"
+      }|]
+
+      verify (response :: Either ClientError Wallet)
+        [ expectJSONError "Error in $.spendingPassword: Expected spending password to be of either length 0 or 32, not 26"
+        ]
+
+    scenario "WALLETS_CREATE_03b - Cannot restore wallet if spendingPassword is hex|base16 string with length less than 64" $ do
+
+      response <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "restore",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "MyFirstWallet",
+        "spendingPassword": "541"
+      }|]
+
+      verify (response :: Either ClientError Wallet)
+        [ expectJSONError "suffix is not in base-16 format: 1"
+        ]
+
+    scenario "WALLETS_CREATE_03c - Cannot create wallet if spendingPassword is hex|base16 string with length MORE than 64" $ do
+
+      response <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "create",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "MyFirstWallet",
+        "spendingPassword": "c0b75cebcd14403d7abba4227cea5b99b1b09148623cd927fa7bb40c6cca5583c"
+      }|]
+
+      verify (response :: Either ClientError Wallet)
+        [ expectJSONError "Error in $.spendingPassword: suffix is not in base-16 format: c"
+        ]
+
+    scenario "WALLETS_CREATE_03d - Cannot restore wallet if spendingPassword is hex|base16 string with length MORE than 64" $ do
+
+      response <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "restore",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "MyFirstWallet",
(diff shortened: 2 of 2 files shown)


test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
 spec :: Scenarios Context
 spec = do
-    scenario "demo unsafeRequest" $ do
+
+    scenario "WALLETS_CREATE_01 - One can restore previously deleted wallet" $ do
+      fixture <- setup $ defaultSetup
+          & initialCoins .~ [1000111]
+          & walletName .~ "漢ę ó ł ąś ł żźćń字"
+          & mnemonicWords .~ testBackupPhrase
+          & assuranceLevel .~ NormalAssurance
```
KtorZ: If those aren't important for the scenario (since the wallet is deleted right after), I'd recommend to simply go with:

```hs
           & initialCoins .~ [1000111]
```

This `initialCoins` makes sense here in order to have had some activity on the wallet before trying to restore it.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+
+      -- 1. Delete wallet
+      deletion <- request $ Client.deleteWallet
+          $- (fixture ^. wallet . walletId)
+
+      verify deletion
+          [ expectSuccess
+          ]
```
KtorZ: if the goal is just to make a successful request, then, you can just go with 

```hs
       deletion <- successfulRequest $ Client.deleteWallet
           $- (fixture ^. wallet . walletId)
```

:+1:

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+          noSpendingPassword -- defaultSpendingPassword does not work
+          StrictAssurance
+          kanjiPolishWalletName
+          RestoreWallet
+      verify restoration
+          [ expectWalletEventuallyRestored
+            , expectFieldEqual walletName kanjiPolishWalletName
+            , expectFieldEqual assuranceLevel StrictAssurance
+            , expectFieldEqual hasSpendingPassword False
+          ]
```
KtorZ: Minor things, but comma should be aligned with the brackets here :)

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+
+      -- 2. Make sure balance is correct
+      balance <- request $ Client.getAccountBalance
+          $- (fixture ^. wallet . walletId)
+          $- defaultAccountId
+      verify balance
+          [ expectFieldEqual amount 1000111
+          ]
```
KtorZ: So this is a nice case for a new "expectation" that could just be plugged in with the others above

```hs
    verify restoration
        [ ...
        , expectBalanceEqual 1000111
        ]
```

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+          & walletName .~ "漢ę ó ł ąś ł żźćń字"
+          & mnemonicWords .~ testBackupPhrase
+          & assuranceLevel .~ NormalAssurance
+
+      -- 2. Delete wallet
+      deleteResp <- request $ Client.deleteWallet
+          $- (fixture ^. wallet . walletId)
+      verify deleteResp
+          [ expectSuccess
+          ]
```
KtorZ: Same as above :+1:

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+      verify balance
+          [ expectFieldEqual amount 1000111
+          ]
+
+    scenario "WALLETS_CREATE_02a - Create/restore wallet with No spending password" $ do
+      -- 1. Setup new wallet
+      fixture <- setup $ defaultSetup
+          & walletName .~ "漢ę ó ł ąś ł żźćń字"
+          & mnemonicWords .~ testBackupPhrase
+          & assuranceLevel .~ NormalAssurance
```
KtorZ: Same as above :+1:

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+          noSpendingPassword
+          StrictAssurance
+          kanjiPolishWalletName
+          RestoreWallet
+      verify (Right restoreResp)
+          [ expectWalletEventuallyRestored
+            , expectFieldEqual walletName kanjiPolishWalletName
+            , expectFieldEqual assuranceLevel StrictAssurance
+            , expectFieldEqual hasSpendingPassword False
+          ]
```
KtorZ: So, if you intend to make some assertions on a request, use `request` instead of `successfulRequest`. The later is really for request you want to fire without caring much about the result, but still making sure they succeeded.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+
+      -- 6. Verify noSpendingPassword on new wallet
+      getAgainResp <- request $ Client.getWallet
+          $- (createResp ^. walletId)
+      verify getAgainResp
+          [
+              expectFieldEqual walletName newWalletName
+            , expectFieldEqual assuranceLevel NormalAssurance
+            , expectFieldEqual hasSpendingPassword False
+          ]
```
KtorZ: 5 & 6 sounds like a separate scenario to me, different from what's tested in 1,2,3 & 4?

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+
+      let endpoint = "api" </> "v1" </> ("wallets" :: Text)
+
+      createResp <- unsafeRequest ("POST", endpoint) $ Just $ [json|{
+        "operation": "create",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "My HODL Wallet",
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
+      }|]
```
KtorZ, line 104: There's no need for an unsafe request here, we gotta define a proper setter for the spending password instead :smile:

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+    scenario "WALLETS_CREATE_03 - Cannot create/restore wallet it spendingPassword is not hex|base16 string" $ do
+
+      let endpoint = "api" </> "v1" </> ("wallets" :: Text)
+
+      response <- unsafeRequest ("POST", endpoint) $ Just $ [json|{
+        "operation": "create",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "normal",
+        "name": "MyFirstWallet",
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c6"
```
KtorZ, line 170: This doesn't quite match the title :p Password is an hex string, the length is incorrect thought. I'd go for two separate scenarios here as well.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+            , expectFieldEqual hasSpendingPassword True
+            , expectFieldEqual walletName "My HODL Wallet 2"
+            , expectFieldEqual assuranceLevel StrictAssurance
+          ]
+
+    scenario "WALLETS_CREATE_03 - Cannot create/restore wallet it spendingPassword is not hex|base16 string" $ do
+
+      let endpoint = "api" </> "v1" </> ("wallets" :: Text)
+
+      response <- unsafeRequest ("POST", endpoint) $ Just $ [json|{
```
KtorZ: I'd recommend to just write

```hs
unsafeRequest("POST", "api/v1/wallets") $ Just $ [json|{ ...
```

The combinator `(</>)` helps to combine `Text` with non-text types you may get from a previous response like a `WalletId` or `AccountIndex`. But as long as you just have plain text, no need for this :)

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+        "operation": "restore",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "strict",
+        "name": "MyFirstWallet",
+        "spendingPassword": " "
+      }|]
+
+      verify (response3 :: Either ClientError Wallet)
+          [ expectJSONError "Error in $.spendingPassword: suffix is not in base-16 format"
+          ]
```
KtorZ: So, well, actually three separate scenario then ^^. The idea is to really keep scenario about testing one and one thing. This way, when one of them fails, we can quickly identify the reason why it's failing. Otherwise, it's not necessarily obvious to know where in a scenario things have failed.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
+      }|]
+
+      verify (response2 :: Either ClientError Wallet)
+          [ expectJSONError "Error in $.operation: When parsing Cardano.Wallet.API.V1.Types.WalletOperation expected a String with the tag of a constructor but got"
+          ]
+
+    scenario "WALLETS_CREATE_05a - can't create wallet if backupPhrase is already known" $ do
+        fixture <- setup $ defaultSetup
+            & mnemonicWords .~ testBackupPhrase
```
KtorZ: Note that there's no need to have a custom backupPhrase here, we could just leave this as `setup defaultSetup` and use the one from the fixture anyway.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+          [ expectJSONError "Invalid number of mnemonic words: got 9 words, expected 12 words"
+          ]
+
+      response2 <- unsafeRequest ("POST", endpoint) $ Just $ [json|{
+        "operation": "restore",
+        "backupPhrase": #{mnemonicsWith9Words},
+        "assuranceLevel": "strict",
+        "name": #{russianWalletName},
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
+      }|]
```
KtorZ, line 435: again, these should be two separate scenarios.

KtorZ (changes requested):
Great stuff in there @piotr-iohk ! Impressive :stuck_out_tongue: 

There's room for more separation between scenarios however, and some simplification to keep the code tidy.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: This was actually a workaround proposed by you. :) I wanted to use `walletId` returned by this response in the next response, like:

```      
      getAgainResp <- request $ Client.getWallet
                      $- (createResp ^. walletId)
```
and since the result of `successfulRequest` is `response` then it was easier, I guess, than with the `request` which returns `Either...`

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: oh, thanks :)

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: sure, separated.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok

test/integration/Test/Integration/Framework/Request.hs
```diff
@@ -70,10 +72,13 @@ class Request originalResponse where
...
 
-instance Request (APIResponse a) where
+instance Show a => Request (APIResponse a) where
     type Response (APIResponse a) = a
     request action =
-        view httpClient >>= liftIO . fmap (fmap wrData) . action
+        view httpClient >>= liftIO . fmap debug . fmap (fmap wrData) . action
+      where
+        debug x = x --traceShow x x
+        -- replace `x` with `traceShow x x` for having Json responses in the test log
```
KtorZ: We'll have to remove this :) ... Let's not commit commented out code, and debug artifacts. By the by, I just realized you could have go with `print` in the scenarios if you just needed to debug a particular response... 

```hs
do
    response <- request $ .... 
    print response
    ....
```

test/integration/Test/Integration/Scenario/Wallets.hs
KtorZ: Hehe, got you, was lacking a bit of context here maybe :upside_down_face: 
So, my suggestion: if what you need is the wallet id, it's already in the fixture :sweat_smile:  ... So you don't need the one from the previous request, the ID is unique per wallet and strictly determined by the mnemonic. If you use a same mnemonic, you get the same id (hey look, an internal requirement / capability here :D)

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,590 @@ import           Test.Integration.Framework.DSL
...
+
+      getWalletsResp <- request $ Client.getWallets
+      verify getWalletsResp
+          [ expectSuccess
+          -- additional checks needs to be added
+          -- expect returned list of wallets has size 3
+          -- expect there is a wallet on the list with certain field value
+          ]
+
+    scenario "WALLETS_UPDATE_01 - updating a wallet persists the update" $ do
```
KtorZ, line 795: Let's separate PRs here and have UPDATE_xx as part of another one :+1:

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: this one was already there. Just added the WALLETS_UPDATE_01 and some additional requests, which I now see are unnecessary, so removing them.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: so once we have it we can update the test? :)

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Well, trying to add it, but failing...
```

expectBalanceEqual
    => a
    -> Either ClientError a
    -> m ()
expectBalanceEqual = expectFieldEqual amount
```

test/integration/Test/Integration/Framework/Request.hs
piotr-iohk: ok

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -476,6 +484,16 @@ expectJSONError excerpt = \case
...
+-- Would be nice to have this method. The one commented above does not compile.
+-- `T.unpack` is `Text - > String` and  a is `a rigid type variable bound by the type signature  for: expectJSONSuccess...`
+-- expectJSONSuccess
+--     :: (MonadIO m, MonadFail m, Show a)
+--     => String
+--     -> Either ClientError a
+--     -> m ()
+-- expectJSONSuccess excerpt = \case
+--     Left e  -> wantedSuccessButError e
+--     Right a -> T.unpack a `shouldContain` excerpt
```
KtorZ: This should better be in a ticket than as commented-out code :+1:

Comments should be reserved for precisions and explaining _why_; we do not want commented-out code, unless if used to illustrate usage of a function in a documentation header.

test/integration/Test/Integration/Framework/Request.hs
```diff
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
+
```
KtorZ: To be removed :smile:  
I think this was added to make the code compile with the `debug` helper. Not needed anymore.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,673 @@ import           Test.Integration.Framework.DSL
...
+        ]
+
+    scenario "WALLETS_CREATE_03c - Cannot create wallet it spendingPassword is not hex|base16 string" $ do
+
+      response2 <- unsafeRequest ("POST", "api/v1/wallets") $ Just $ [json|{
+        "operation": "create",
+        "backupPhrase": #{testBackupPhrase},
+        "assuranceLevel": "strict",
+        "name": "MyFirstWallet",
+        "spendingPassword": "541"
```
KtorZ: `"541"` is actually a valid base16 string :upside_down_face:  
Perhaps `[]`, `541`, `null` or `"patate"` are better candidates here :+1:

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,673 @@ import           Test.Integration.Framework.DSL
...
+      }|]
+
+      verify (response2 :: Either ClientError Wallet)
+        [ expectJSONError "Error in $.spendingPassword: suffix is not in base-16 format"
```
KtorZ, line 258: If this passes, then the error message is slightly wrong.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,673 @@ import           Test.Integration.Framework.DSL
...
+        "operation": "restore",
+        "backupPhrase": #{invalidMnemonics},
+        "assuranceLevel": "normal",
+        "name": #{russianWalletName},
+        "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
+      }|]
+
+      verify (response :: Either ClientError Wallet)
+        [ expectJSONError "Error in $.backupPhrase: MnemonicError: Invalid entropy checksum: got Checksum 11, expected Checksum 5"
+        ]
```
KtorZ, line 551: I am starting to think that we could make use of a combinator that would run a scenario with both "restore" and "create" operation. What do you think? We have quite many cases that are kinda duplicated because of this. Would be nice to only write things once, but get both tested seamlessly.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,673 @@ import           Test.Integration.Framework.DSL
...
+      fixture <- setup $ defaultSetup
+        & walletName .~ "漢ę ó ł ąś ł żźćń字"
+        & mnemonicWords .~ testBackupPhrase
+
+      response01 <- request $ Client.deleteWallet
+        $- (fixture ^. wallet . walletId)
+
+      verify response01
+        [ expectSuccess
+        ]
```
KtorZ: This could just be `successfulRequest_ $ ...` , if the only thing expected is a success, then, there's no need for a verify clause :+1:

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,673 @@ import           Test.Integration.Framework.DSL
...
+
+      fixture02 <- setup $ defaultSetup
+        & walletName .~ "2"
+        & mnemonicWords .~ testBackupPhrase2
+        & assuranceLevel .~ NormalAssurance
+
+      fixture03 <- setup $ defaultSetup
+        & walletName .~ "3"
+        & mnemonicWords .~ testBackupPhrase3
+        & assuranceLevel .~ StrictAssurance
```
KtorZ, line 763: Note: you don't need to provide backup phrases here. They'll just be generated arbitrarily for each wallet.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -10,48 +10,673 @@ import           Test.Integration.Framework.DSL
...
+
+      -- 2. Vierify each wallet individually
+      sequence_
+        [
```
KtorZ: Yeah, I was using this pattern as a way to show that assertions are somewhat isolated and don't interfere with each other.. I guess it's more confusing than anything else. Especially with a 4th case `getWalletResp <- ...` after :disappointed:

KtorZ (changes requested):

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: yes, indeed. It passes.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Ah, ok. I'll add tests with those values too.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Yes, that would be nice thing to have.

test/integration/Test/Integration/Framework/DSL.hs
piotr-iohk: ok

test/integration/Test/Integration/Framework/Request.hs
piotr-iohk: ok

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: My motivation for using that sequence here was that... it didn't work for some reason otherwise (i.e. without having all those 3 requests in the sequence). I suppose that the reason was simply faulty indentation I had before. Removing it now.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok. From my perspective sometimes it's more explicit and readable for the test to have it. But perhaps no in this particular case.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Actually for `"spendingPassword": null` the wallet is created successfully, but I guess that it is expected?

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Raised bug https://github.com/input-output-hk/cardano-wallet/issues/193

KtorZ (approved):
:tada:

PR #179, opened 2018-12-24
Remove 'WalletId' & 'AccountId' wrappers

KtorZ:
<p align="right">#34</p>

# Overview

- [ ] I have removed `WalletId` and `AccountId` "Kernel Types" in favor of their `Hd` equivalent. 

# Comments

These were introduced in prevision of future evolution. However, we now know that we want to keep the same Hd representation for sequential and random wallets. As such, those types only add one extra unnecessary level of indirection.

This is the first PR of a serie about redesigning the prefiltering code. I'll try to proceed with incremental changes and clean up like this before submitting the actual prefiltering redesign such that the intent is made clearer at that time.

### DIFF
--- a/src/Cardano/Wallet/Kernel/Wallets.hs
+++ b/src/Cardano/Wallet/Kernel/Wallets.hs
@@ -54,7 +54,6 @@ import           Cardano.Wallet.Kernel.Internal (PassiveWallet, walletKeystore,
                      walletProtocolMagic, wallets)
 import qualified Cardano.Wallet.Kernel.Keystore as Keystore
 import qualified Cardano.Wallet.Kernel.Read as Kernel
-import           Cardano.Wallet.Kernel.Types (WalletId (..))
 import           Cardano.Wallet.Kernel.Util.Core (getCurrentTimestamp)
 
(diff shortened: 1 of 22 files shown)


src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -200,7 +199,7 @@ createHdWallet pw mnemonic spendingPassword assuranceLevel walletName = do
     case mbHdAddress of
         Nothing -> do
             -- Here, ideally, we would like to do some cleanup and call
-            -- >>> Keystore.delete (WalletIdHdRnd newRootId) (pw ^. walletKeystore)
+            -- >>> Keystore.deletef newRootId (pw ^. walletKeystore)
```
denisshevchenko: Why `deletef`? :-)

KtorZ: Typo :)

KtorZ: *fixed

denisshevchenko (approved):
LGTM!

Issue #32, opened 2018-11-07
API Endpoint For Addresses à la BIP-44 


KtorZ:
# Context

We want users to be able to retreive new/unused addresses for accounts they own in their wallet. With the (ed25519)BIP-44 scheme, this can be done from the account public keys (as in non-hardened child key derivation) and through address pool api. Address pool contains `gap` of discovered unusused addresses and we can use this pool to inform the user about the unused addresses. Address pool might contain some used addresses as well so having optional (or default) flag to filter only unused might be nice to have.

# Decision

Extend the API to support retreival of new/unused addresses defining a new endpoint for external wallets. The api will return list of addresses from the address pool. There will be optional (or default?) flag to filter unused only from the address pool. Implement this ticket in two interations:
 * implement "address pool endpoint" that will return addresses found in address pool
 * extend the functionality so that endpoint can be given additional flag which will filter address pool and return only unused addresses 

# Acceptance Criterias

1. - [ ] The API `must` support retrieval of new/unused addresses for E.O.S. Wallets
1. - [ ] The API `must` support retrieval of **unused only** addresses for E.O.S. Wallets

---

# Development Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #      | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| ?        | -        |

---

# Retrospective

PR #182, opened 2018-12-27
API Endpoint for Addresses a la BIP-44: API endpoint, first version.

denisshevchenko:
<p align="right">#32</p>

# Overview

- [x] API endpoint for address creation in EOS-wallet.
- [x] Separate types for error during address creation in EOS-wallet.
- [x] Using the _conceptually_ same type hierarchy in `acid-state` for storing these addresses. 
- [x] Additional lenses for work with accounts' public keys.
- [x] `EosWalletId` is now defined in correct place.

# Comments


src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -52,6 +58,9 @@ data CreateAddressError =
     | CreateAddressHdRndAddressSpaceSaturated HdAccountId
       -- ^ The available number of HD addresses in use in such that trying
       -- to find another random index would be too expensive
+    | CreateAddressUnableToDeriveEosAddressPublicKey
+      -- ^ Unable to derive address public key from EOS-wallet's account
+      -- public key.
```
KtorZ: Sounds like we would need a new type instead of extending this one. The `createEosAddress` ain't going to return `CreateAddressKeystoreNotFound` or `CreateAddressHdRndAddressSpaceSaturated`.

src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -178,3 +190,48 @@ newHdAddress nm esk spendingPassword accId hdAddressId =
...
+-- to derive address keys from account public key only.
+createEosAddress
+    :: PublicKey
+    -- ^ Account's public key.
+    -> HdAccountId
+    -> PassiveWallet
+    -> IO (Either CreateAddressError Address)
+createEosAddress accountPublicKey accId pw = do
+    snapshot <- getWalletSnapshot pw
+    let accountAddresses = IxSet.toList $ addressesByAccountId snapshot accId
```
KtorZ: This looks up in `_hdWalletAddresses` from the `HdWallets` in the snapshot. We haven't unified `HdWallets` and `EosHdWallets` yet so this won't work for now, but should work in the future.
Beside, lines 205 to 220 could be replaced with: 

```hs
let newAddressIx = size (snapshot ^. hdWalletsAddresses)
```

src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -178,3 +190,48 @@ newHdAddress nm esk spendingPassword accId hdAddressId =
...
+    -- but we have to check address pool gap for this EOS-wallet because of
+    -- https://github.com/input-output-hk/cardano-wallet/wiki/Address-Derivation-a-la-BIP---44#create-an-address-1
+    let newAddressIx = maximum accountAddressesIxs + 1
+        -- We use non-hardened derivation for EOS-wallet's addresses.
+        mbAddrPublicKey = deriveAddressPublicKey accountPublicKey
+                                                 ExternalChain
+                                                 newAddressIx
+    case mbAddrPublicKey of
+        Nothing ->
+            return $ Left CreateAddressUnableToDeriveEosAddressPublicKey
```
KtorZ: What about `CreateAddressHdSeqGenerationFailed` to make this a bit more aligned with rnd errors.

KtorZ (changes requested):

src/Cardano/Wallet/Kernel/Addresses.hs
denisshevchenko: > This looks up in _hdWalletAddresses from the HdWallets in the snapshot.

Of course! Since we already have `_hdWalletsAddresses` (and addresses for EOS-wallets are the same `Core.Address` as for FOS-wallets) - I thought that we have to use this `(IxSet (Indexed HdAddress))` (and all existing functions/lenses) for working with EOS-wallet's addresses, because you said:

> re-using our existing type hierarchy is going to be much simpler at first.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1159,6 +1220,38 @@ data Account = Account
     } deriving (Show, Ord, Eq, Generic)
 
 
+-- | To create a new address in account in EOS-wallet,
+-- we need account index and EOS-wallet id. Corresponding
+-- account's public key (which will be used for non-hardened
+-- derivation) is already stored in 'acid-state'.
+data NewEosAddress = NewEosAddress
+    { neweosaddrAccountIndex :: !AccountIndex
+    , neweosaddrEosWalletId  :: !EosWalletId
```
KtorZ: No need for a newtype here for `EosWalletId`. We can re-use `WalletId` :+1:

src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -178,3 +216,33 @@ newHdAddress nm esk spendingPassword accId hdAddressId =
...
+    case eosAccountPublicKey snapshot eosAccId of
+        Left _ -> return . Left $ CreateEosAddressUnknownHdAccount eosAccId
+        Right accPublicKey ->
+            case deriveAddressPublicKey accPublicKey ExternalChain newAddrIx of
+                Nothing -> return $ Left $ CreateEosAddressHdSeqGenerationFailed eosAccId
+                Just newAddressPublicKey -> do
+                    let nm = makeNetworkMagic (pw ^. walletProtocolMagic)
+                        newAddress = makePubKeyAddressBoot nm newAddressPublicKey
+                        eosHdAddressId = EosHdAddressId eosAccId (HdAddressIx newAddrIx)
+                        eosHdAddress = EosHdAddress eosHdAddressId (InDb newAddress)
```
KtorZ, line 244: I saw nowhere in your development plan something about re-using our existing hierarchy of HD types (`HdRootId`, `HdAccountId` and `HdAddressId`. 
In the long run, we do want that instead of carrying around new `EosHd` types. Not only have they the same representation, but they also have the same semantic. In the end, they're the same type :+1:

We could move on with this PR, and make another round of cleanup after that to replace all the `EosHd***` types with their `Hd` equivalent but that sounds like a lot of code introduced in this PR that we'll have to remove afterwards.

src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -178,3 +216,33 @@ newHdAddress nm esk spendingPassword accId hdAddressId =
...
+    snapshot <- getWalletSnapshot pw
+    -- Please note that for Ed25519Bip44-scheme address indexes
+    -- are created in a sequence from the last known address.
+    let newAddrIx = fromIntegral . IxSet.size $ eosAddressesByAccountId snapshot eosAccId
+    -- We use non-hardened derivation for EOS-wallet's addresses
+    -- (so we only need account's public key).
+    case eosAccountPublicKey snapshot eosAccId of
+        Left _ -> return . Left $ CreateEosAddressUnknownHdAccount eosAccId
+        Right accPublicKey ->
+            case deriveAddressPublicKey accPublicKey ExternalChain newAddrIx of
```
KtorZ: Not sure we will only derive addresses on the external chain. We don't really make the difference for F.O.S addresses but it will matter here. I'd suggest to move the ChainCode to the function parameters and let the caller decides :+1:

src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -178,3 +216,33 @@ newHdAddress nm esk spendingPassword accId hdAddressId =
...
+                Nothing -> return $ Left $ CreateEosAddressHdSeqGenerationFailed eosAccId
```
KtorZ: Now that I look at it, I think we could be more specific here. The only way we reach this `Nothing` branch would be because the `newAddrIx` is out of the non-hardened bound. Meaning that, we're running out of addresses. 
That's quite unlikely, but it would be great if the error could tell us about it:

`CreateEosAddressNoMoreAddresses` or something like this :+1:

KtorZ (changes requested):
This PR has to be separated in smaller chunks. One candidate for a first PR would be the unification between `EosHd` types and `Hd` types; this is required before we introduce a way for creating new HD addresses. 

Try to keep PR concise and on-the-spot (i.e. about one thing). If several classes of changes have to be introduced to tackle a particular task, make sure to split them across several PRs :+1:

src/Cardano/Wallet/API/V1/Types.hs
denisshevchenko: You're right, it's just a wrapper around `Text`, as well as `WalletId`.

denisshevchenko:
Closed, because of bunch of smaller PRs with corresponding changes.

PR #183, opened 2018-12-28
Unify foreignPending & pending during prefiltering

KtorZ:
<p align="right">#34</p>

# Overview

Still approaching prefiltering refactoring with baby steps, trying to extract what can be extracted to make the final PR easier to review.

- [x] I have removed the distinction between `foreignInputs` and `pendingInputs` for prefilteredBlock
- [x] I have simplified the handling of foreign inputs during prefiltering

# Comments

> NOTE: This would _in theory_require a `SafeCopy` migration. However, as next PRs are going to completely change the `PrefilteredBlock` representation, migrations are going to be required anyway (or made impossible).

We added foreign transaction as part of prefiltered blocks during [CO-421](https://iohk.myjetbrains.com/youtrack/issue/CO-421) and kept them separated initially even though, there's no conceptual differences between foreign inputs and "prefiltered inputs" here. 

Important note also: it suffices to add foreign inputs to the list prefiltered inputs. We don't need to do this mumbo jumbo as part of `prefilterTxForWallets` where we artificially add tx ids to them after re-indexing all of them per transaction inputs. 

We do need foreign transactions ids to figure in the prefiltered `LocalBlockMeta`, however, the only way for a foreign transaction to be relevant for a given account is to have some outputs targeting this account. Therefore, the tx id will be included anyway as part of the "normal" prefiltering on the outputs!

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/Spec/Update.hs
+++ b/src/Cardano/Wallet/Kernel/DB/Spec/Update.hs
@@ -197,7 +197,7 @@ applyBlock (SecurityParameter k) pb = do
         (utxo', balance') = updateUtxo      pb (utxo, balance)
         (pending', rem1)  = updatePending   (pfbInputs pb) (current ^. checkpointPending)
         blockMeta'        = (current ^. checkpointBlockMeta) <> (localBlockMeta $ pfbMeta pb)
-        (foreign', rem2)  = updatePending   (pfbForeignInputs pb) (current ^. checkpointForeign)
+        (foreign', rem2)  = updatePending   (pfbInputs pb) (current ^. checkpointForeign)
     if (pfbContext pb) `blockContextSucceeds` (current ^. checkpointContext . lazy) then do
       put $ Checkpoints . takeNewest k . NewestFirst $ Checkpoint {
           _checkpointUtxo        = InDb utxo'
@@ -230,7 +230,7 @@ applyBlockPartial pb = do
         (utxo', balance') = updateUtxo           pb (utxo, balance)
         (pending', rem1)  = updatePending        (pfbInputs pb) (current ^. pcheckpointPending)
         blockMeta'        = (current ^. pcheckpointBlockMeta) <> pfbMeta pb
-        (foreign', rem2)  = updatePending        (pfbForeignInputs pb) (current ^. pcheckpointForeign)
+        (foreign', rem2)  = updatePending        (pfbInputs pb) (current ^. pcheckpointForeign)
     if (pfbContext pb) `blockContextSucceeds` (current ^. cpContext . lazy) then do
       put $ Checkpoints $ NewestFirst $ PartialCheckpoint {
           _pcheckpointUtxo        = InDb utxo'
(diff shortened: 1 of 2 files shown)


src/Cardano/Wallet/Kernel/DB/Spec/Update.hs
```diff
@@ -197,7 +197,7 @@ applyBlock (SecurityParameter k) pb = do
         (utxo', balance') = updateUtxo      pb (utxo, balance)
         (pending', rem1)  = updatePending   (pfbInputs pb) (current ^. checkpointPending)
         blockMeta'        = (current ^. checkpointBlockMeta) <> (localBlockMeta $ pfbMeta pb)
-        (foreign', rem2)  = updatePending   (pfbForeignInputs pb) (current ^. checkpointForeign)
+        (foreign', rem2)  = updatePending   (pfbInputs pb) (current ^. checkpointForeign)
```
KtorZ, line 200: `updatePending` is basically a set difference, so here, we use a bigger set to remove from but that's okay because `checkpointPending` and `checkpointForeign` are disjoint anyway.

src/Cardano/Wallet/Kernel/PrefilterTx.hs
```diff
@@ -159,57 +155,14 @@ prefilterTx wKey tx = ((prefInps',prefOuts'),metas)
...
-        :: Map TxIn HdAccountId
-        -> Map HdAccountId (Set TxIn)
-    reindexByAccount byTxIn =
-        Map.fromListWith Set.union $ Map.elems $ Map.mapWithKey f byTxIn
-      where
-          f :: TxIn -> HdAccountId -> (HdAccountId, Set TxIn)
-          f txin accId = (accId, Set.singleton txin)
-
+prefilterTxForWallets wKeys tx =
+    mconcat $ map ((flip prefilterTx) tx) wKeys
```
KtorZ, line 165: Back to what it was (cf https://github.com/input-output-hk/cardano-sl/pull/3768/files#diff-58eddd8c17809e069bc881287f7f0620L167)

src/Cardano/Wallet/Kernel/PrefilterTx.hs
```diff
@@ -363,24 +305,15 @@ mkPrefBlock context inps outs accId = (accId, PrefilteredBlock {
...
-        walletInps'           = Set.map fst  $ byAccountId accId Set.empty walletInps
-        foreignInps'          = Set.map fst  $ byAccountId accId Set.empty foreignInps
+        byAccountId k def = fromMaybe def . Map.lookup k
 
-        allInps = (Map.map fst inps)
-        inpsWithtxId = byAccountId accId Set.empty allInps
         -- this Set.map may reduce the number of elements. But this is okey, since we
         -- don't care about repetitions on txIds.
-
+        inps' = Set.map fst (byAccountId accId Set.empty inps) <> (maybe mempty Pending.txIns $ Map.lookup accId foreignInps)
```
KtorZ, line 312: So, this is the major change. Instead of trying to add the pending transactions at a lower level during prefiltering, we add them here, if relevant to the account. (Note that, because foreign inputs are so rare, we could also just add them to all prefiltered block without risking much performance degradation).

akegalj (approved):
LGTM

I was mostly focusing on comparing the diff and checking could there be some major mistake - as I am not fully familiar with modified code bits.

KtorZ:
Thanks Ante. More coming today.

PR #192, opened 2019-01-03
Fix order of evaluation in integration DSL assertion

KtorZ:
<p align="right">#34</p>

# Overview

- [x] I have inverted the arguments for `shouldBe` when asserting errors, such that the error message better reflect the actual assertion.

# Comments

when failed, this looked like:

```
  expected: ClientWalletError AddressNotFound
  but got: ClientWalletError (NotEnoughMoney ErrCannotCoverFee)
```

The first argument passed to `shouldBe` should be what we got, and the second what we actually expected.


KtorZ:
Very small thing. But was kinda bothering me while running tests :sweat_smile:

akegalj (approved):

paweljakubas (commented):
ok!

PR #196, opened 2019-01-03
Delegate to Node for Settings + Info, remove dead code

parsonsmatt:
<p align="right">#9</p>

# Overview

- [x] I have modified the Info and Settings endpoints to point to the node API.
- [x] Removed the old code for these features from the wallet.

# Comments


src/Cardano/Wallet/API/Internal/Handlers.hs
```diff
@@ -48,21 +45,6 @@ applyUpdate nc = do
         Right () ->
             pure NoContent
 
-handleNodeError :: ClientError () -> Handler a
```
parsonsmatt, line 51: Relocated into `NodeProxy` module.

src/Cardano/Wallet/API/V1/Handlers.hs
```diff
@@ -11,16 +11,18 @@ import qualified Cardano.Wallet.API.V1.Handlers.Settings as Settings
...
-         :<|> Settings.handlers     passiveWallet
-         :<|> Info.handlers         w
+handlers :: NodeHttpClient -> ActiveWalletLayer IO -> Server V1.API
+handlers nodeClient w =
+    Addresses.handlers         passiveWallet
+    :<|> Wallets.handlers      passiveWallet
+    :<|> Accounts.handlers     passiveWallet
+    :<|> Transactions.handlers w
+    :<|> Settings.handlers     nodeClient
+    :<|> Info.handlers         nodeClient
```
parsonsmatt, line 26: tempted to bite the bullet and just remove all the vertical alignment here

parsonsmatt (commented):
This PR could be broken up into multiple PRs if you feel that's a better idea.

src/Cardano/Wallet/API/V1/Handlers.hs
KtorZ: I know that feel. Maybe we should allow one day per month called "alignment catharsis day" where we only submit PRs to fix variable-length vertical alignments.

KtorZ (approved):

akegalj (approved):

PR #197, opened 2019-01-04
[DEVOPS-1199] Add cardano-wallet-client CLI and easy API

rvl:
# Overview

Part of https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-1199

- Add `cardano-wallet-client` CLI for using and testing the wallet backend.
- Add a small program to plot timing statistics collected by `wait-for-sync` and `wait-for-restore`
- Add a glue API which makes it easy to set up a wallet client in `ghci`.
- Set up stack/nix integration

# Screenshots

CLI help page:
```
Usage: cardano-wallet-client ([--state-dir DIRECTORY] | ([--cert FILENAME]
                             [--key FILENAME] | [--pem FILENAME])
                             [--cacert FILENAME]) [-k|--insecure] [HOST:PORT]
                             [-p|--path PATH] COMMAND
  Connect to the API of a running Cardano Wallet backend.

Available options:
  --state-dir DIRECTORY    Wallet state directory containing "tls" subdirectory
                           with certificates
  --cert FILENAME          X509 certificate file
  --key FILENAME           Certificate key file
  --pem FILENAME           Combined X509 certificate and key file
  --cacert FILENAME        CA certificate chain for authenticating the server
  -k,--insecure            Skip server certificate authentication
  HOST:PORT                Wallet API host and port to connect to
  -p,--path PATH           Base URL path
  -h,--help                Show this help text

High-level commands
  wait-for-sync            Poll wallet until it has fully synced its chain
  wait-for-restore         Poll wallet until the restore operation is complete

Basic API calls
  node-info                Query node info
  node-settings            Query static settings of wallet

Wallets
  list-wallets             List wallets
  show-wallet              Information about a wallet
  edit-wallet              Update a wallet
  create-wallet            Create a new wallet from mnemonic
  restore-wallet           Restore a wallet from mnemonic
  delete-wallet            Delete a wallet
  get-utxo-statistics      Get UTxO distribution of wallet

Accounts
  list-accounts            Retrieve full list of accounts in wallet
  show-account             Information about an account
  account-balance          Show the balance of an account
  create-account           Add an account to a wallet
  edit-account             Update an account
  delete-account           Delete account

Addresses and Transactions
  list-addresses           List addresses
  show-address             Information about an address
  create-address           Create an address
  list-transactions        Get transaction history
  send-transaction         Send transaction to one or multiple target addresses
  estimate-fees            Estimate fees for a transaction
```

Output of `cardano-wallet-client wait-for-restore` passed into `cardano-wallet-sync-plot`:
restore-stats


KtorZ:
Sweet stuff in there :) 
Haven't yet review any of this but will come to that eventually. 

Just some side cosmetic notes: 

-I had plans to group a bit all executables we have spread across folders into an `app` folder (as the default/standard stack approach) in a similar fashion to what was done in [cardano-sl/cluster](https://github.com/input-output-hk/cardano-sl/tree/develop/cluster)
That'd be the same for this one. Seemingly, let's go for `cardano-wallet-client` rather than tool which is a bit too generic in my opinion.

I like the idea. Will give a more useful feedback ASAP.

rvl:
Cool, thanks. The CLI is clearly an initial version and not complete, but we can decide what needs to be done in this first PR vs what can be done in subsequent PRs.

I have changed the name, good idea.

I have also moved the code into the main `cardano-wallet.cabal` package. However I think it might be better as a separate package. This is because it adds code and dependency bloat to the cardano-wallet library, and also because multi-package projects are well-supported by `stack` and `cabal new-build`.

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+import           Data.Aeson hiding ((.=))
+import           Data.Char (toLower)
+import           Data.List (isSuffixOf)
+import qualified Data.Text as T
+import           Data.Vector ((!))
+import qualified Data.Vector as V
+import           Options.Applicative
+import           Universum hiding ((.~))
+
+import           Graphics.Rendering.Chart.Backend.Cairo
```
KtorZ: Let's use explicit imports for `Options.Applicative` & `Graphis.Rendering.Chart.Backend.Cairo` :+1:

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards   #-}
```
KtorZ, line 3: We don't much like this one anymore I've heard :smile:

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+instance FromJSON DataFile where
+  parseJSON = withObject "DataFile" $ \ob -> DataFile <$> ob .: "data" <*> ob .:? "start_time" .!= ""
+
+instance FromJSON Record where
+  parseJSON = withArray "Record" $ \v ->
+    if V.length v == 2
+      then (Record <$> parseJSON (v ! 0) <*> parseJSON (v ! 1))
+      else fail "Incorrect record array length"
```
KtorZ: Can't we rely on generics for this? Is the `DataFile` format constraints or is a something we define?

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+          layoutlr_right_axis . laxis_title .= "Blocks/second"
+
+plotTitle :: PlotType -> String
+plotTitle PlotSync    = "Blockchain sync"
+plotTitle PlotRestore = "Wallet restore"
+
+loadRecords :: FilePath -> IO DataFile
+loadRecords = eitherDecodeFileStrict >=> handle
+  where
+    handle (Left err) = die err >> pure (DataFile [] "")
```
KtorZ: Doesn't `die` terminates the program? What's the need for `>> pure ...` ?

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+                     | otherwise = SVG
+
+timeAxisLabel :: Text -> String
+timeAxisLabel startTime = ("Time (seconds" <> extra <> ")")
+  where extra | T.null startTime = ""
+              | otherwise = " since started at " <> T.unpack startTime
+
+plotRecords :: PlotType -> Text -> Bool -> Bool -> FilePath -> DataFile -> IO ()
+plotRecords pl title pleft pright outfile (DataFile rs startTime) =
+  toFile (def & fo_format .~ formatFromFileName outfile) outfile $ do
```
KtorZ: Minor thing but this function could be pure and would benefits from a separation between the plot construction and the writing to file. Having run more like:

```hs
run PlotOptions{..} = do
  df <- loadRecords poInputFile
  let plot =  plotRecords poPlot poExtraTitle poPlotLeft poPlotRight df
  toFile (def & fo_format .~ formatFromFileName outfile)
```

app/wallet-client/Main.hs
```diff
@@ -0,0 +1,21 @@
+module Main where
+
+import           Options.Applicative
+import           System.Exit (ExitCode)
+import           Universum
+
+import           Cardano.Wallet.Client.CLI.Options
+import           Cardano.Wallet.Client.CLI.Run
+import           Cardano.Wallet.Client.Easy
```
KtorZ: Same as above, please, favor explicit imports.

app/wallet-client/Main.hs
```diff
@@ -0,0 +1,21 @@
...
+
+main :: IO ()
+main = exitWith =<< uncurry run =<< execParser opts
```
KtorZ, line 13: You don't like `do` notation, do you :trollface: ?

app/wallet-client/README.md
```diff
@@ -0,0 +1,40 @@
+## Easy client functions and CLI for Cardano Wallet API
+
+This provides more util functions for setting up a `WalletClient` so
+that you can quickly make API calls from `ghci`.
+
+There is also a CLI for use in test scripts.
+
+### How to use
+
+Start a wallet using `stack exec cardano-node -- ....`
```
KtorZ: Hopla, this is now `cardano-wallet-server`

app/wallet-client/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+    cardano-wallet-client --state-dir ./state-wallet-testnet localhost:8090 wait-for-sync
+
+
+### Future work
+
+ - implement commands for the entire wallet API
+ - reduce boilerplate
+ - add a `--curl` option which doesn't do anything except print an
+   equivalent `curl` command-line.
```
KtorZ, line 36: I like the idea :o

cardano-wallet.cabal
```diff
@@ -130,8 +139,13 @@ library
     , wai
     , wai-middleware-throttle
     , warp
+    , x509
+    , x509-store
     , zlib
 
+  if !os(windows)
+    build-depends:     unix
+
```
KtorZ: Can't these additions be made to the `wallet-client` executable instead :thinking:  ?

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -103,7 +103,7 @@ module Cardano.Wallet.API.V1.Types (
   , mkLocalTimeDifference
   , EstimatedCompletionTime
   , mkEstimatedCompletionTime
-  , SyncThroughput
+  , SyncThroughput(..)
```
KtorZ, line 104: I think the type was opaque for a reason :sweat_smile:

src/Cardano/Wallet/Client/CLI/Options.hs
```diff
@@ -0,0 +1,251 @@
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE LambdaCase                #-}
+
+module Cardano.Wallet.Client.CLI.Options where
```
KtorZ, line 3: What about having this as a `other-module` of the cardano-wallet-client ?

src/Cardano/Wallet/Client/CLI/Options.hs
```diff
@@ -0,0 +1,251 @@
...
+import           Cardano.Wallet.Client (getAccounts, getAddressIndex,
+                     getTransactionIndex, getWallets)
+import           Cardano.Wallet.Client.Easy
+import           Cardano.Wallet.Client.CLI.ProcessUtil (ProcessID)
+
+----------------------------------------------------------------------------
+-- CLI Types
+
+-- | The command from the user
+data Action m = WaitForSync (Maybe ProcessID) (Maybe FilePath)
```
KtorZ: Minor style recommendation: no more variable-length vertical indent please :pray: Going to the line with a fixed-length vertical alignment plays better with git in the long run.

src/Cardano/Wallet/Client/CLI/ProcessUtil.hs
```diff
@@ -0,0 +1,56 @@
...
+checkProcessExists :: Maybe ProcessID -> IO Bool
+checkProcessExists = const (pure True)
+
+cancelOnExit :: Async a -> IO b -> IO b
+cancelOnExit _ = id
+
+interruptSelf :: IO ()
+interruptSelf = pure ()
+
+#endif
```
KtorZ: So this won't be working at all for windows? Should we even build the CLI on windows?

src/Cardano/Wallet/Client/HttpsSettings.hs
```diff
@@ -0,0 +1,71 @@
...
+        }
+    clientHooks = def
+        { onCertificateRequest = const . return $ credentials
+        , onServerCertificate  = case authServer of
+            AuthenticateServer -> validateDefaultWithIP
+            AllowInsecure      -> \_ _ _ _ -> pure []
+        }
+    clientSupported = def
+        { supportedCiphers = ciphersuite_default
+        }
```
KtorZ, line 71: Looks heavily inspired from: https://github.com/input-output-hk/cardano-sl/blob/develop/node/src/Cardano/Node/Manager.hs

Any reason why not to use it? (The only _true_ difference I could spot was the `AuthenticateServer` addition.

src/Cardano/Wallet/Client/Easy.hs
```diff
@@ -0,0 +1,167 @@
...
+  where
+    check resp = (unfinished, msg, info)
+      where
+        progress = nfoSyncProgress resp
+        unfinished = progress < mkSyncPercentage 100
+        msg = sformat ("Blockchain syncing: "%int%"% "%int%"/"%int%" blocks")
+              (unSyncPercentage progress)
+              (fst info) (fromMaybe 0 (snd info))
+        info = ( unBlockchainHeight $ nfoLocalBlockchainHeight resp
+               , unBlockchainHeight <$> nfoBlockchainHeight resp )
```
KtorZ, line 158: Sounds a lot like: https://github.com/input-output-hk/cardano-sl/blob/develop/cluster/src/Cardano/Cluster.hs#L131

src/Cardano/Wallet/Client/CLI/Options.hs
```diff
@@ -0,0 +1,251 @@
...
+                 <> help "X509 certificate file")
+  <*> strOption (long "key"
+                 <> metavar "FILENAME"
+                 <> help "Certificate key file")
+
+clientAuthPemP :: Parser ClientAuthConfig
+clientAuthPemP = pemConfig <$> strOption (long "pem"
+                                          <> metavar "FILENAME"
+                                          <> help "Combined X509 certificate and key file")
+  where pemConfig pem = ClientAuthConfig pem pem
```
KtorZ: why bother with the `pem` here :thinking: ? This is merely useful when using with `cURL` but I am not sure it buys us anything here.

src/Cardano/Wallet/Client/CLI/Options.hs
```diff
@@ -0,0 +1,251 @@
...
+
+baseUrlP :: Parser BaseUrl
+baseUrlP = baseUrl <$> addrP <*> path
+  where
+    baseUrl (host, port) = BaseUrl Https (B8.unpack host) (fromIntegral port)
+    addrP = argument readAddrM (metavar "HOST:PORT" <> value ("localhost", 8090) <> help "Wallet API host and port to connect to")
+    readAddrM = eitherReader (first show . Parsec.parse addrParserNoWildcard "" . T.pack)
+    path = strOption (long "path" <> short 'p' <> value "" <> metavar "PATH" <> help "Base URL path")
```
KtorZ, line 95: Note, we already have parsers for `NetworkAddress`. I'd advise to piggy-back on them and then, translate `NetworkAddress` to `BaseUrl` whenever needed (as in:)

https://github.com/input-output-hk/cardano-sl/blob/develop/cluster/src/Cardano/Cluster/Util.hs#L379-L381

src/Cardano/Wallet/Client/Wait.hs
```diff
@@ -0,0 +1,152 @@
...
+  build (SyncErrorException e) = build e
+  build SyncErrorInterrupted = build ("Interrupted" :: Text)
+
+instance Eq SyncError where
+  SyncErrorClient a      == SyncErrorClient b      = a == b
+  SyncErrorProcessDied a == SyncErrorProcessDied b = a == b
+  SyncErrorTimedOut a    == SyncErrorTimedOut b    = a == b
+  SyncErrorInterrupted   == SyncErrorInterrupted   = True
+  SyncErrorException _   == SyncErrorException _   = True
+  _ == _ = False
```
KtorZ: Indentation is 2-space all over this file. That is weird. Is your editor picking up the `.editorconfig` at the root of the repository?

src/Cardano/Wallet/Client/Wait.hs
```diff
@@ -0,0 +1,152 @@
...
+
+instance ToJSON r => ToJSON (SyncResult r) where
+  toJSON (SyncResult err st dur rs) =
+    object $ ["data" .= toJSON rs, "start_time" .= st, "duration" .= dur] <> status err
+    where
+      status Nothing  = [ "success" .= True ]
+      status (Just e) = [ "success" .= False, "error" .= String (show e) ]
+
+instance ToJSON SyncError where
+  toJSON e = String (show e)
```
KtorZ: What about using `Generics` here too :) ?

src/Cardano/Wallet/Client/CLI/Options.hs
```diff
@@ -0,0 +1,251 @@
...
+listTransactionsP = (\wid acc addr wc -> getTransactionIndex wc wid acc addr)
+                    <$> optional walletIdP
+                    <*> optional accountIndexP
+                    <*> optional walAddressP
+
+getUtxoStatisticsP :: Parser (WalletCall m UtxoStatistics)
+getUtxoStatisticsP = flip getUtxoStatistics <$> walletIdP
+
+----------------------------------------------------------------------------
+-- Little option parsers
```
KtorZ, line 260: We're on the verge of reviewing our node's CLI parsers and structure. It might make sense to include all those in the list to have them centralized.

KtorZ (changes requested):
A first pass on review. There's quite a lot so I mostly skimmed through it for now. I'll try to play with the content and give more useful feedback a bit later.

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+main = run =<< execParser opts
+  where
+    opts = info (optionsParser <**> helper)
+      ( fullDesc
+        <> progDesc "Plot the results of a wallet sync" )
+
+run :: PlotOptions -> IO ()
+run PlotOptions{..} = do
+  df <- loadRecords poInputFile
+  plotRecords poPlot poExtraTitle poPlotLeft poPlotRight poOutputFile df
```
parsonsmatt: This repo uses 4 space indent. Please enable `editorconfig` for your editor so that it will automatically pick up these things.

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+        <> command "restore" (info (pure PlotRestore) (progDesc "Wallet restore chart"))
+        <> commandGroup "Plot type"
+      )
+    leftP = flag True False (long "no-left" <> help "Don't plot first dataset")
+    rightP = flag True False (long "no-right" <> help "Don't plot second dataset")
+    titleP = strOption (long "extra-title" <> short 't' <> value "" <> help "Append text to chart title")
+    inputP = argument str (metavar "INFILE" <> help "Input JSON")
+    outputP = argument str (metavar "OUTFILE" <> help "Output PNG")
+
+data Record = Record Double [Double] deriving (Show)
```
parsonsmatt: Please include a newline indent between datatypes.

I'd like to see a bang pattern on the `Double` field, or `StrictData` enabled for the module.

I'd prefer to see a record field for these entries to more clearly indicate what they are.

app/sync-plot/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+data DataFile = DataFile
+  { dfRecords   :: [Record]
+  , dfStartTime :: Text
```
parsonsmatt: A start time should not be a `Text` value -- `UTCTime` is more appropriate. Given that the parser below defaults to empty string, `Maybe UTCTime` is likely the right choice.

app/sync-plot/Main.hs
parsonsmatt: We could actually use `DerivingVia` for this one, but we're not on 8.6 yet.

parsonsmatt: Please include diagnostics to make debugging this easier, like:

```suggestion
      else fail 
          $ "Incorrect record array length for Record, expected 2 entries but got: " 
          ++ show v
```

app/wallet-client/README.md
```diff
@@ -0,0 +1,40 @@
+## Easy client functions and CLI for Cardano Wallet API
```
parsonsmatt, line 1: Fantastic documentation, thanks! :smile:

cardano-wallet.cabal
parsonsmatt: I am personally in favor of tiny executables and big libraries, makes it a lot easier to factor stuff out and write tests later.

src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: Yeah, we should either expose an accessor or the relevant client should use `mkSyncThroughput`.

src/Cardano/Wallet/Client/CLI/Options.hs
parsonsmatt: Agree, please newline at the `=` so everything has a constant 4 space indent :smile: 

```haskell
data Action m
    = WaitForSync ...
    | WaitForRestore ....
    | forall a. ....
```

src/Cardano/Wallet/Client/CLI/Options.hs
```diff
@@ -0,0 +1,251 @@
...
+import           Cardano.Wallet.Client.CLI.ProcessUtil (ProcessID)
+
+----------------------------------------------------------------------------
+-- CLI Types
+
+-- | The command from the user
+data Action m = WaitForSync (Maybe ProcessID) (Maybe FilePath)
+              | WaitForRestore (Maybe ProcessID) (Maybe FilePath)
+              | forall a. ToJSON a => WalletEndpoint (WalletCall m a) -- ^ Wallet API call
+              | WalletEndpointVoid (WalletCallVoid m) -- ^ Wallet API call with empty response
```
parsonsmatt: However, since you're using an existential, I'd rather see the `GADTs` syntax used here.

src/Cardano/Wallet/Client/Easy.hs
```diff
@@ -0,0 +1,167 @@
...
+  } deriving (Show)
+
+data ClientAuthConfig = ClientAuthConfig
+  { cfgCertFile    :: FilePath
+  , cfgPrivKeyFile :: FilePath
+  } deriving (Show)
+
+data ConfigError = ConfigCredentialsError Text
+                 | ConfigCertificateError FilePath Text
+                 deriving (Show, Typeable)
```
parsonsmatt: Variable length alignment, please newline before the `=` and align on the 4 space boundary.

test/unit/Test/Spec/Client/Wait.hs
```diff
@@ -0,0 +1,95 @@
+module Test.Spec.Client.Wait (spec) where
```
parsonsmatt, line 1: Great tests :smile:

test/unit/Test/Spec/Client/Wait.hs
```diff
@@ -0,0 +1,95 @@
...
+import           Pos.Util.Servant (single)
+
+import           Cardano.Wallet.API.V1.Types (WalletError (..))
+import           Cardano.Wallet.Client.Easy
+import           Cardano.Wallet.Client.Wait
+import           Cardano.Wallet.Client.CLI.ProcessUtil (interruptSelf)
+
+
+spec :: Spec
+spec = before mkTestWalletClient $ do
```
parsonsmatt: `beforeAll` is probably more appropriate; this will create a new wallet client before each spec. `beforeAll` will share the created wallet client among all specs.

parsonsmatt (changes requested):
The PR overall looks good, but the style is very different from existing code on this repo. Please enable `editorconfig` for your editor, and it will pick up the indentation and line-length settings for the repo.

Additionally, please favor indentation and alignment that is consistent in the face of edits and renaming. This usually means that things that are to be vertically aligned should be aligned by a newline and the appropriate indentation level for that bit. Eg:

```haskell
data Foo = Foo 
         | Bar
```

If I add a type to `Foo` or change the name, then I must also realign the constructors. However,

```haskell
data Foo
    = Foo
    | Bar
```

This layout is more resilient to edits, allowing a diff to be cleaner and easier to read.

rvl:
@KtorZ @parsonsmatt Thanks very much for the review comments. I have addressed most of them but am not finished yet.

app/sync-plot/Main.hs
rvl: Done.

app/sync-plot/Main.hs
rvl: Universum has `die :: String -> IO ()`, so I changed it to be the one from `System.Exit`.

app/sync-plot/Main.hs
rvl: Yes, that is much better, done.

app/wallet-client/README.md
rvl: Updated

src/Cardano/Wallet/Client/CLI/ProcessUtil.hs
rvl: Running the CLI on Windows (as part of tests) is actually one of the reasons I wrote this code. The `wait-for-sync --pid` option won't work yet on Windows so I have marked it as TODO.

src/Cardano/Wallet/Client/HttpsSettings.hs
rvl: Yes, it's the `AuthenticateServer` option. They should be de-duplicated. Will the `mkHttpsManagerSettings` function stay in `cardano-sl` or move elsewhere?

src/Cardano/Wallet/Client/Easy.hs
rvl: Yes it does the same thing, plus logging the timing information, handling Ctrl-C, and giving up if a PID has gone. The wait function uses `Control.Retry`, but I'm not sure if that helped.

src/Cardano/Wallet/Client/CLI/Options.hs
rvl: I want this program to inter-operate well with users' configurations. If they have generated certificates themselves then this option will help. In any case, it's easier to supply just `--pem` instead of both `--cert` and `--key`.

src/Cardano/Wallet/Client/CLI/Options.hs
rvl: I have changed it to use `Pos.Util.OptParse.fromParsec Pos.Core.NetworkAddress.addrParserNoWildcard`. The `BaseUrl` should stay here because then the user can provide an API base URL. This is for interoperability with the interesting/weird setups that users are likely to try -- for example, running the wallet behind a reverse proxy.

src/Cardano/Wallet/Client/Wait.hs
rvl: I have now enabled EditorConfig in Emacs.

app/sync-plot/Main.hs
rvl: Done.

app/sync-plot/Main.hs
rvl: Done.

app/sync-plot/Main.hs
rvl: Good idea, done.

src/Cardano/Wallet/Client/CLI/Options.hs
rvl: Done.

test/unit/Test/Spec/Client/Wait.hs
rvl: Done

src/Cardano/Wallet/API/V1/Types.hs
rvl: What would be the best way to log the throughput value as a number that can be stored in a JSON file and then later plotted?

src/Cardano/Wallet/Client/Wait.hs
rvl: I don't think it's possible because there is no `Generic` instance of `SomeException`.

src/Cardano/Wallet/Client/CLI/Options.hs
rvl: It could, but I have a unit test for a function in this module.

Cabal new-build internal libraries might provide a way of keeping non-core wallet code out of the main library.

app/sync-plot/Main.hs
rvl: :smile: Is `RecordWildcards` a problem in this file of this repo? The data type and usage are in the same file, and the record fields have prefixes.

app/sync-plot/Main.hs
rvl: I intentionally wrote instances for this because they are short and the JSON format should look like a normal JSON data file, not something weird and haskelly. Is it OK? (the alternative is to define `genericParseJSON customOptions`)

app/sync-plot/Main.hs
rvl: Done.

src/Cardano/Wallet/Client/Easy.hs
rvl: Fixed.

rvl:
@ktorz @parsonsmatt Thanks for the reviews, I apologise for the delay (leave, other things, etc). I have rebased on develop and addressed/responded to your comments so far.

I also implemented a few more CLI commands for API endpoints (although the entire API is not covered yet).

About code formatting (everyone's favourite topic I know). I have fixed it, but are half-indented `where`s acceptable? That means the actual bindings are indented 4 spaces instead of 8.

What else should be done for this PR?

KtorZ:
@rvl Apologize for the time it takes but priority has been quite elsewhere at the moment for us. And this is extra code to maintain and test (even though, mind you, it is much appreciated). 
Since it's also a bit out-of-the-loop for us, we haven't really included this in our QA pipeline which should be done at some point. 

Might I can put that on @piotr-iohk radar. Would be a nice idea to play with it and see if things are working smoothly. This is of course of a lower priority vs anything about decoupling or BIP-44 :+1:

rvl:
OK I'll follow up with QA and see how I can assist them.
This tool is something that I need for devops work, because there is currently no easy way to use the wallet API.
Additionally, I have built the Windows chain sync acceptance tests on top of this program. We really need automated sync tests running on Windows.
What's the best way to get this merged? We don't want to have a long-running feature branch.

KtorZ:
I understand and I do agree with you that the situation is not practical here. What I am truly afraid of here is that, as soon as this lands on this repository, it becomes (eventually) the wallet's team responsibility. And that is something that comes a bit "out of nowhere" for us because it wasn't planned nor discussed with any of us :disappointed: 
We can't just afford this extra work at the moment. My recommendation here would be to have that maybe as _another repository_, maintained and managed by devops or, maybe a dedicated team later once we've agreed on spending time and resources on building and improving such tool.

KtorZ:
Closing as per last comment + private conversation.

PR #198, opened 2019-01-04
Move executables in a dedicated location

KtorZ:
<p align="right">-</p>

# Overview

- [x] I have moved executables in a dedicated location
- [x] I have renamed executables following a simple convention 'cardano-wallet-<executable-name>'

# Comments

@rvl This before introducing new executables to the repository. Now we can have `cardano-wallet-client` under `app/` and make things a bit more ordered :smile: 
Doesn't buy us much but, let's do it before we introduce a whole build pipeline on top of that :no_mouth:


rvl (approved):
Looks sensible, thanks.

Issue #191, opened 2019-01-03
Fee sanity check is triggered more than it should


KtorZ:
| Release | Operating System       | Cause |
| ---     | ---                    | ---   |
| 1.4    | Windows & OSX & Linux) | Code  |

# Context
 
 

When solving [CSL-2526](https://github.com/input-output-hk/cardano-sl/pull/3815), we introduced an extra sanity check to make the API fails if it was to generate a transaction with _too many fees_ (to prevent us from creating illed transactions because of a programming error in the coin selection / fee calculation). 
Yet, it turns out this check is a bit too strict / wrong for it compare transaction fees to an absolute value. Therefore, some transactions with many inputs and, actually expensive fees, fail to be created despite being totally valid.  

# Steps to Reproduce

1. Submit a (valid) transaction with _many_ inputs to the API

## Expected behavior

The transaction is successfully created with corresponding fees.

## Actual behavior
 
<A clear and concise description of what you observe instead. If applicable add screenshots to help explain your problem.>

The API fails (500 error) with "fees out of bound".

---

# Resolution Plan

- [x] Extend current test suite with a valid case of valid transaction with out-of-bound fees.
- [x] Make the fee sanity check relative to the transaction's size itself instead of being absolute
- [x] Backport this change to cardano-sl

# PR

| Number   | Base      |
| ---      | ---       |
| #199       | `develop` |
| #207       | `develop` |
| [cardano-sl #3993](https://github.com/input-output-hk/cardano-sl/pull/3993) | `release/2.0.1` |

# QA 

In order to make the current test scenarios fail (and shed a light on the issue), we had to do a few things:

- Make sure the sanity check was included earlier in the coin selection algorithm such that it would occur at the very root of the selection, even if next steps are mocked by test scenarios

- Generated more fragmented UTxO to allow for many inputs to be selected (leading to high fees, out of the absolute bound). See (https://github.com/input-output-hk/cardano-wallet/pull/199/files#diff-82f6927d7e5a4cfec96683f543e2cb58R210) and @akegalj 's comment https://github.com/input-output-hk/cardano-wallet/pull/199/files#r245356862. 

With this, our test suite was able to identify the issue and trigger the wrong sanity check. By then changing the sanity check to a relative one, I haven't been able to make them fail again even with 50K+ more tests. 

Note that the sanity check is rather arbitrary but is now comparing the actual computed fee amount with the estimated one. Cf the corresponding note in the code:

[...]

PR #199, opened 2019-01-04
Review Fee Sanity Check

KtorZ:
<p align="right">#191</p>

# Overview

- [x] I have move the fee sanity check inside the `adjustForFees` method such that it's actually ran for each CoinSelection and is triggered by the appropriate tests
- [x] I have removed some tests introduced in [cardano-sl #3815](https://github.com/input-output-hk/cardano-sl/pull/3815) as they're now redundant with the ones done in the CoinSelection
- [x] I have removed the `feeSanityCheck` option from the `CoinSelectionOption`

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
+++ b/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
@@ -18,7 +18,6 @@ module Cardano.Wallet.Kernel.CoinSelection.FromGeneric (
   , largestFirst
     -- * Estimating fees
   , estimateCardanoFee
-  , checkCardanoFeeSanity
   , boundAddrAttrSize
   , boundTxAttrSize
     -- * Estimating transaction limits
@@ -44,7 +43,6 @@ import           Pos.Chain.Txp as Core (TxIn, TxOutAux, Utxo, toaOut,
 import           Pos.Core as Core (AddrAttributes, Address, Coin (..),
                      TxSizeLinear, addCoin, calculateTxSizeLinear, checkCoin,
                      divCoin, isRedeemAddress, maxCoinVal, mkCoin, subCoin,
-                     sumCoins, txSizeLinearMinValue, unsafeMulCoin,
                      unsafeSubCoin)
 
 import           Pos.Core.Attributes (Attributes)
@@ -148,8 +146,6 @@ data InputGrouping =
 data CoinSelectionOptions = CoinSelectionOptions {
       csoEstimateFee       :: Int -> NonEmpty Core.Coin -> Core.Coin
     -- ^ A function to estimate the fees.
-    , csoFeesSanityCheck   :: Core.Coin -> Bool
-    -- ^ A function we can use to check if fees are not too big or too small.
     , csoInputGrouping     :: InputGrouping
     -- ^ A preference regarding input grouping.
     , csoExpenseRegulation :: ExpenseRegulation
@@ -163,10 +159,9 @@ data CoinSelectionOptions = CoinSelectionOptions {
 -- | Creates new 'CoinSelectionOptions' using 'NoGrouping' as default
 -- 'InputGrouping' and 'SenderPaysFee' as default 'ExpenseRegulation'.
 newOptions :: (Int -> NonEmpty Core.Coin -> Core.Coin)
-           -> (Core.Coin -> Bool) -> CoinSelectionOptions
-newOptions estimateFee check = CoinSelectionOptions {
+           -> CoinSelectionOptions
+newOptions estimateFee = CoinSelectionOptions {
       csoEstimateFee       = estimateFee
-    , csoFeesSanityCheck   = check
     , csoInputGrouping     = IgnoreGrouping
     , csoExpenseRegulation = SenderPaysFee
     , csoDustThreshold     = Core.mkCoin 0
@@ -237,21 +232,7 @@ runCoinSelT opts pickUtxo policy (NE.sortBy (flip (comparing outVal)) -> request
         -- We adjust for fees /after/ potentially dealing with grouping
         -- Since grouping only affects the inputs we select, this makes no
         -- difference.
-        csf <- adjustForFees (feeOptions opts) pickUtxo css
-        let fees = computeFees csf
-        if (csoFeesSanityCheck opts fees)
-        then return csf
-        else error $ "fee out of bound" <> show fees
-
-    computeFees :: CoinSelFinalResult Cardano -> Core.Coin
-    computeFees csf =
-        let
-            inputs  = Core.sumCoins $ fmap (Core.txOutValue . Core.toaOut . snd) $ NE.toList $ csrInputs csf
-            outputs = Core.sumCoins $ fmap (Core.txOutValue . Core.toaOut) $ NE.toList $ csrOutputs csf
-            change  = Core.sumCoins $ csrChange csf
-        in  -- NOTE: _Rather_ safe since we expect fee to be within reasonnable bounds
-            mkCoin $ fromIntegral (inputs - (outputs + change))
-
+        adjustForFees (feeOptions opts) pickUtxo css
 
     intInputGrouping :: InputGrouping
                      -> CoinSelT Core.Utxo CoinSelHardErr m [CoinSelResult Cardano]
@@ -410,15 +391,6 @@ estimateCardanoFee :: TxSizeLinear -> Int -> [Word64] -> Word64
 estimateCardanoFee linearFeePolicy ins outs
     = ceiling $ calculateTxSizeLinear linearFeePolicy
               $ hi $ estimateSize boundAddrAttrSize boundTxAttrSize ins outs
-
-checkCardanoFeeSanity :: TxSizeLinear -> Coin -> Bool
-checkCardanoFeeSanity linearFeePolicy fees =
-    let
-        maxCoeff :: Int = 2
-        minFees = Core.mkCoin $ floor $ txSizeLinearMinValue linearFeePolicy
-    in
-        (fees >= minFees) && (fees <= Core.unsafeMulCoin minFees maxCoeff)
-
 
 -- | Size to use for a value of type @Attributes AddrAttributes@ when estimating
 --   encoded transaction sizes. The minimum possible value is 2.
--- a/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
+++ b/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
@@ -44,7 +44,8 @@ import           Pos.Chain.Txp as Core (TxIn, TxOutAux, Utxo, toaOut,
 import           Pos.Core as Core (AddrAttributes, Address, Coin (..),
                      TxSizeLinear, addCoin, calculateTxSizeLinear, checkCoin,
                      divCoin, isRedeemAddress, maxCoinVal, mkCoin, subCoin,
-                     txSizeLinearMinValue, unsafeMulCoin, unsafeSubCoin)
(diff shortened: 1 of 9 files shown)


src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -396,15 +392,6 @@ estimateCardanoFee linearFeePolicy ins outs
     = ceiling $ calculateTxSizeLinear linearFeePolicy
               $ hi $ estimateSize boundAddrAttrSize boundTxAttrSize ins outs
 
-checkCardanoFeeSanity :: TxSizeLinear -> Coin -> Bool
-checkCardanoFeeSanity linearFeePolicy fees =
-    let
-        maxCoeff :: Int = 2
-        minFees = Core.mkCoin $ floor $ txSizeLinearMinValue linearFeePolicy
-    in
-        (fees >= minFees) && (fees <= Core.unsafeMulCoin minFees maxCoeff)
```
KtorZ, line 405: So, this was the absolute check checking a value against a particular config, regardless of the underlying transaction. Now removed in favor of a relative one.

src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -162,10 +159,9 @@ data CoinSelectionOptions = CoinSelectionOptions {
 -- | Creates new 'CoinSelectionOptions' using 'NoGrouping' as default
 -- 'InputGrouping' and 'SenderPaysFee' as default 'ExpenseRegulation'.
 newOptions :: (Int -> NonEmpty Core.Coin -> Core.Coin)
-           -> (Core.Coin -> Bool) -> CoinSelectionOptions
-newOptions estimateFee check = CoinSelectionOptions {
+           -> CoinSelectionOptions
+newOptions estimateFee = CoinSelectionOptions {
       csoEstimateFee       = estimateFee
-    , csoFeesSanityCheck   = check
```
KtorZ, line 168: The check is now "hardcoded" in the form of an invariant in the generic code.

src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
```diff
@@ -97,15 +97,20 @@ adjustForFees feeOptions pickUtxo css = do
...
             senderPaysFee pickUtxo feeOptions inps outs chgs
 
-    let neInps = case inps' of
-            []   -> error "adjustForFees: empty list of inputs"
-            i:is -> i :| is
+    let estimatedFee = getFee $ feeUpperBound feeOptions inps outs chgs
+    let actualFee = getFee $ computeFee inps' outs' chgs'
+    if actualFee `valueDiv` 2 > estimatedFee then -- Using `valueDiv` instead of `valueMul` to avoid overflow
+        error $ "adjustForFees: fee out of bounds: " <> pretty actualFee <> " while expecting ~" <> pretty estimatedFee
+    else do
```
KtorZ, line 117: So this is the new check that should never be triggered. It's not an exception because that's not something we want to communicate to user in theory. 
Testing ensure that this is never triggered. Instead of comparing to an absolute value, we control that our actual fees are not unrealistically big.

src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -230,13 +228,8 @@ newUnsignedTransaction ActiveWallet{..} options accountId payees = runExceptT $
     -- that it may change in the future.
     let attributes = def :: TxAttributes
     let tx = UnsignedTx inputs outputs attributes coins
-
-    -- STEP 3: Sanity test. Here we check whether our fees are within a reasonable
-    -- range.
     let fees = computeFeesOfUnsignedTx tx
-    if csoFeesSanityCheck options fees
-    then return (snapshot, tx, fees, availableUtxo)
-    else error $ "fees out of bound " <> show fees
```
KtorZ, line 239: So this was move into the generic coin selection algorithm.

test/unit/Test/Spec/CoinSelection.hs
```diff
@@ -63,15 +64,9 @@ import           Test.Spec.CoinSelection.Generators (InitialBalance (..),
...
 
-freeLunchCheck :: Core.Coin -> Bool
-freeLunchCheck c = c == Core.mkCoin 0
-
--- | The smallest fee possible.
-minFee :: Int -> NonEmpty Core.Coin -> Core.Coin
-minFee _ _ = Core.mkCoin 1
-
-minFeeCheck :: Core.Coin -> Bool
-minFeeCheck c = c == Core.mkCoin 1
```
KtorZ, line 74: The checker weren't ran before so we never noticed but assuming that actual fee will be 1 when the fee estimator was also returning 1 was wrong. This works fine when there's only one payee, but because fee gets distributed "proportionnally" over payees, when multiple payees, we end up splitting 1 lovelace (which is the smallest) value we can have across multiple entries. Resulting in each of them rounding up to `1`.
    
So in the end, when fees are at the smallest, our algorithm makes each payees pay for them. Since this is quite an unrealistic scenario (we won't ever have a policy that charges only 1 lovelace per transaction...) I haven't tried to "fix" the fee adjustment algorithm. Instead, falling back to a constant fee that is big enough to allow for fee to be divvied across multiple payees (note that the generator generates usually between 1 and 10 payees).

test/unit/Test/Spec/CoinSelection.hs
```diff
@@ -659,37 +638,37 @@ spec =
             -- like attributes, and trying to setup syntetic experiments with
             -- less than 1ADA (10^6 lovelaces) is probably counter-productive
             prop "one payee, SenderPaysFee, fee = cardano" $ \pm -> forAll (
-                payOne pm cardanoFee cardanoFeeCheck identity (InitialADA 1000) (PayADA 100) random
+                payOne pm cardanoFee identity (InitialADA 1000) (PayADA 100) random
                 ) $ \(utxo, payee, res) ->
                   paymentSucceededWith utxo payee res [feeWasPayed SenderPaysFee]
             prop "multiple payees, SenderPaysFee, fee = cardano" $ \pm -> forAll (
-                payBatch pm cardanoFee cardanoFeeCheck identity (InitialADA 1000) (PayADA 100) random
+                pay pm genFragmentedUtxo genPayees cardanoFee identity (InitialADA 1000) (PayADA 100) random
```
KtorZ, line 645: So here's the main "trick" here to highlight the scenario we had were multiple payees would generate too many fees. By using a different generator (`payBatch` relied on `genUtxoWithAtLeast` which generate around 10 payees at most), we end up generating more payees for each transaction (got up to 500) and I was able to successfully trigger the bug. 
Now that this is now fixed with the relative comparison (estimated fees vs actual fees), even with loads of payees.

test/unit/Test/Spec/CoinSelection/Generators.hs
```diff
@@ -204,6 +205,21 @@ genUtxoWithAtLeast payment = do
...
+    let balance       = toLovelaces payment
+        twoPercentOf  = balance `div` 50
+    genUtxo $ StakeGenOptions {
+                  stakeMaxValue         = Just (Core.mkCoin twoPercentOf)
+                , stakeGenerationTarget = AtLeast
+                , stakeNeeded           = Core.mkCoin (toLovelaces payment)
+                , stakeMaxEntries       = Just 1000
+                , fiddlyAddresses       = False
+                , allowRedeemAddresses  = False
+            }
```
KtorZ, line 220: Making the UTxO more fragmented here forces more inputs to be selected and more payees to be generated which serves our purpose for testing edge cases of fee sanity checking.

test/unit/Test/Spec/GetTransactions.hs
```diff
@@ -146,62 +144,6 @@ prepareFixtures nm initialBalance = do
             , fixturePw = pw
         }
```
KtorZ, line 146: Code and tests below were introduced in https://github.com/input-output-hk/cardano-sl/pull/3815 and are now completely redundant with the one in `Spec/CoinSelection`

src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
akegalj: is it safe to skip minimum bound check? What if fees are too small? (the thing we were testing here https://github.com/input-output-hk/cardano-wallet/pull/199/files#diff-393909f5cc8109550702f2c76a1b4afdL405 ). Is it safe to assume that fees won't be less then estimated fees?

KtorZ: Safe for the users that's for sure. Safe for us, that's debatable. But indeed, the `estimatedFee` are actually a minimal fees so we are quite guaranteed that the actual one are at least equal, if not bigger. 
I'll add the min bound check, oversight from me :+1:

src/Cardano/Wallet/Kernel/Transactions.hs
akegalj: except we don't check minimum bound (https://github.com/input-output-hk/cardano-wallet/pull/199/files#r245325794) which might be ok?

EDIT: already answered https://github.com/input-output-hk/cardano-wallet/pull/199#discussion_r245327717

akegalj (commented):
LGTM

Remaining things to do before approving:
 * https://github.com/input-output-hk/cardano-wallet/pull/199/files#r245327717

test/unit/Test/Spec/CoinSelection/Generators.hs
akegalj: Running a test on https://github.com/input-output-hk/cardano-wallet/pull/199/commits/22129c04fd9d962a8284abd1123c4425614c4330:
```
Coin selection policies unit tests
  random
    multiple payees, SenderPaysFee, fee = cardano FAILED [1]
  Fiddly Addresses
    multiple payees, SenderPaysFee, fee = cardano
      +++ OK, passed 5 tests.

Failures:

  test/unit/Test/Spec/CoinSelection.hs:666:13: 
  1) Coin selection policies unit tests, random, multiple payees, SenderPaysFee, fee = cardano
       uncaught exception: ErrorCall
       fee out of boundCoin {getCoin = 1072341}
       CallStack (from HasCallStack):
         error, called at src/Universum/Debug.hs:60:11 in universum-1.2.0-2E8EL6k5Fv214PFWOWwkIj:Universum.Debug
         error, called at src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs:244:14 in cardano-wallet-2.0.0-H4X2696IYufKTiT2WKvm3c:Cardano.Wallet.Kernel.CoinSelection.FromGeneric
       (after 1 test)
         ProtocolMagic {getProtocolMagicId = ProtocolMagicId {unProtocolMagicId = 1}, getRequiresNetworkMagic = RequiresNoMagic}
         Exception thrown while showing test case:
           fee out of boundCoin {getCoin = 1072341}
           CallStack (from HasCallStack):
             error, called at src/Universum/Debug.hs:60:11 in universum-1.2.0-2E8EL6k5Fv214PFWOWwkIj:Universum.Debug
             error, called at src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs:244:14 in cardano-wallet-2.0.0-H4X2696IYufKTiT2WKvm3c:Cardano.Wallet.Kernel.CoinSelection.FromGeneric
         
```
convinced me its a superset of old property https://github.com/input-output-hk/cardano-wallet/pull/199/files#diff-13c14a11ea8258a566fc79e3b7ed54c9L149

akegalj (approved):

parsonsmatt (approved):

PR #204, opened 2019-01-07
Review Kernel/PrefilterTx

KtorZ:
<p align="right">#34</p>

# Overview

- [x] I have separated computation of TxMeta & LocalBlockMeta from Prefiltering
- [x] I have made Prefiltering stateful (in regards for derivation a la BIP-44 incoming)
- [x] I have changed the type representation of `PrefilteredBlock` to become a list of prefiltered transactions

# Comments

~~This is the big piece. Note that the unit tests aren't (yet) compiling because they relied on old methods for prefiltering UTxO that are not there anymore. I'll fix that eventually; still I wanted to make a PR so that @uroboros could have a look at it.~~


src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -57,23 +58,26 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
   Passive Wallet API implementation
 -------------------------------------------------------------------------------}
 
-type PrefilterResult = ((BlockContext, Map HdAccountId PrefilteredBlock), [TxMeta])
```
KtorZ, line 60: So, the `IsOurs` abstraction works with a list of credentials which do allows us to prefiltering all accounts of a block n one traversal. Then, upon storing each prefiltered block, we re-index them by account id (see `pfbAccounts` below). There's no need to carry around the `Map HdAccountId ...` anymore.

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -57,23 +58,26 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
...
-         xs -> Just $ map (\b -> first (b ^. rbContext,) $ prefilterBlock foreignPendings b xs) bs
+    isForeign <- flip hasPending . foreignPendingByAccount <$> getWalletSnapshot pw
+    case res of
+         [] -> return Nothing
+         xs -> fmap Just $ flip evalStateT xs $ forM bs $ \rb -> do
+            pb <- state (prefilterBlock isForeign rb)
+            metas <- state (resolvedToTxMetas rb) >>= \case
+                Left e  -> throwM e
+                Right x -> return x
+            return ((rb ^. rbContext, pb), metas)
```
KtorZ, line 81: Because transactions metadata are actually computed on the resolved block and has nothing to do with prefiltering, I've separated the two concepts. So, yes, we are traversing the block twice which has the same complexity as traversing it once. If we notice severe performance issues, we may introduce some caching of the address decryption but I don't think it'll be necessary.

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -290,18 +290,18 @@ applyBlock k context restriction blocks = runUpdateDiscardSnapshot $ do
         , accountUpdateNew   = AccountUpdateNewUpToDate Map.empty
         , accountUpdate      =
             matchHdAccountCheckpoints
-              (tryUpdate' $ Spec.applyBlock k      pb)
-              (tryUpdate' $ Spec.applyBlockPartial pb)
+              (tryUpdate' $ Spec.applyBlock k      (context, pb))
+              (tryUpdate' $ Spec.applyBlockPartial (context, pb))
```
KtorZ, line 294: `PrefilteredBlock` no longer contain a `BlockContext`. In many places, we do have prefiltered block without context and only apply a context later. This indicates that the context shouln't be part of a prefiltered block, but part of a higher-level sum type. Going for the poor substitute here: a 2-tuple.

src/Cardano/Wallet/Kernel/DB/Resolved.hs
```diff
@@ -25,72 +29,45 @@ import qualified Formatting.Buildable
...
 -- NOTE: We cannot recover the original transaction from a 'ResolvedTx'.
 -- Any information needed inside the wallet kernel must be explicitly
 -- represented here.
 data ResolvedTx = ResolvedTx {
       -- | Transaction inputs
-      _rtxInputs  :: InDb (NonEmpty (Core.TxIn, ResolvedInput))
+      -- A transaction input @(h, i)@ points to the @i@th output of the transaction
+      -- with hash @h@, which is not particularly informative. The corresponding
+      -- 'ResolvedInput' is obtained by looking up what that output actually is.
+      _rtxInputs  :: InDb (NonEmpty (TxIn, TxOutAux))
```
KtorZ, line 58: I've just removed the type-alias `ResolvedInput`. Debatable but, I am not a big fan of type-aliase. They introduce an extra level of indirection without given us any extra type safety in return. All-in-all, they tend to make code less clear.

src/Cardano/Wallet/Kernel/DB/Resolved.hs
```diff
@@ -110,6 +87,175 @@ data ResolvedBlock = ResolvedBlock {
...
+                { _txMetaId = tx ^. rtxMeta . fromDb . _1
+                , _txMetaAmount = absCoin spent gained
+                , _txMetaInputs = inps
+                , _txMetaOutputs = outs
+                , _txMetaCreationAt = tx ^. rtxMeta . fromDb . _2
+                , _txMetaIsLocal = isLocal
+                , _txMetaIsOutgoing = gained < spent
+                , _txMetaWalletId = view fromDb $ getHdRootId $ accId ^. hdAccountIdParent
+                , _txMetaAccountIx = getHdAccountIx $ accId ^. hdAccountIdIx
+                }
```
KtorZ, line 118: This has been slightly revised and made type-safe with a proper exception when something goes wrong. Note that we do not do much with this exception at the moment and it's just thrown in the wild. But that's slightly better than the "error" we use to have in there.

src/Cardano/Wallet/Kernel/DB/Resolved.hs
```diff
@@ -110,6 +87,175 @@ data ResolvedBlock = ResolvedBlock {
...
+
+
+-- | Compute LocalMetadata from a ResolvedBlock. Any metadata for addresses
+-- that aren't ours are discarded.
+resolvedToLocalBlockMeta
+    :: IsOurs s
+    => ResolvedBlock
+    -> s
+    -> (LocalBlockMeta, s)
+resolvedToLocalBlockMeta block = runState $ fmap LocalBlockMeta $ BlockMeta
```
KtorZ: I am looking forward to remove this one. First of all because, that's another traversal of the block.
And second of all, because it computes things that are: 1/ wrong 2/ probably useless. 
I am currently running a survey across exchanges to know what metadata do they use from the backend. If my sentiment is right, probably none (or only barely a few) of them uses the "changeAddress" metadata on addresses. Our definition of it (as per the specification) is currently erroneous and doesn't reflect correctly what's going on from local transaction (if you send money from one of your account to another one of yours, there's no change address, according to the rules define in the specification... which is not the intuition we have of it).

Anyway, if no one is using that, I'd simply go for the easy way and remove the "changeAddress" metadata entirely which will cut-off this entire function. If the feature is ever needed again, we may re-assess the intial design and specification and come up with a valid definition of a change address.

src/Cardano/Wallet/Kernel/DB/Spec/Pending.hs
```diff
@@ -145,6 +147,11 @@ txIns = Set.fromList
 rawTxOuts :: Pending -> Core.Utxo
 rawTxOuts = Core.utxoUnions . map (Core.txOuts . Core.taTx) . Map.elems . toMap
 
+-- | Check whether a given transaction is part of a pending group.
+hasPending :: Foldable f => Core.TxIn -> f Pending -> Bool
+hasPending txin =
+    any (Set.member txin . txIns) . F.toList
```
KtorZ: Oops. Didn't update that one but this is slightly wrong. `hasPending` should be account-dependent. The integration test about redeeming ADA currently only passes because there's only a single account in the scenario. But we shouldn't account for foreign transactions that aren't related to our account.

TODO.

src/Cardano/Wallet/Kernel/Restore.hs
```diff
@@ -341,6 +355,18 @@ getWalletInitInfo coreConfig creds lock = do
...
+      where
+        toUtxo (txIn,(txOutAux,_))         = (txIn,txOutAux)
+        toAddr (_   ,(txOutAux,addressId)) = HD.initHdAddress addressId (txOutAddress . toaOut $ txOutAux)
+
+        toSummary :: (TxIn,(TxOutAux,HD.HdAddressId))
+                  -> ((TxIn,TxOutAux),HD.HdAddress)
+        toSummary item = (toUtxo item, toAddr item)
+
+        utxoSummary = map toSummary $ M.toList utxoWithAddrs
+        (utxoL, addrs) = unzip utxoSummary
```
KtorZ, line 369: Basically preserving the old structure here until we also review this whole `Restore` module..

src/Cardano/Wallet/Kernel/Util/Core.hs
```diff
@@ -115,25 +111,10 @@ absCoin ca cb
...
-toOutPair :: Core.TxOutAux -> (Core.Address, Core.Coin)
-toOutPair txOutAux = (toAddress txOutAux, toCoin txOutAux)
-
-fromUtxo :: Core.Utxo -> Maybe (NE.NonEmpty (Core.Address, Core.Coin))
-fromUtxo utxo = NE.nonEmpty $ toOutPair <$> Map.elems utxo
-
-derefIn :: Core.TxIn -> Maybe (Core.TxId, Word32)
-derefIn txIn = case txIn of
-   Core.TxInUnknown _ _  -> Nothing
-   Core.TxInUtxo txId ix -> Just (txId, ix)
```
KtorZ, line 136: now dead-code.

src/Cardano/Wallet/Kernel/BListener.hs
uroboros: incidentally, when we still had 'isChangeAddress' the situation was a bit more complicated because 'isChange' required knowledge about prefiltering (e.g., "did _all_ the inputs in this tx belong to me?").

src/Cardano/Wallet/Kernel/DB/Resolved.hs
uroboros: agreed

src/Cardano/Wallet/Kernel/DB/Resolved.hs
uroboros: (see previous comment: in the presence of `isChangeAddr`, metadata becomes more dependent on prefiltering)

src/Cardano/Wallet/Kernel/DB/Spec/Pending.hs
uroboros: the `DB.Spec.*` code tracks the Wallet Spec closely and is ignorant of AccountId. I suggest we layer the introduction of AccountId on top of this (and leave this level unchanged).

src/Cardano/Wallet/Kernel/PrefilterTx.hs
```diff
@@ -1,395 +1,164 @@
...
+    accountEq accId =
+        (== accId) . view (_1 . hdAddressId . hdAddressIdParent)
+    byAccount tx m (HdAddress (HdAddressId accId _) addr, _) =
+        let
+            outs' = Map.filter (accountEq accId) (pftOutputs tx)
+            txs   = [tx { pftOutputs = outs' }]
+            txid  = (pftMeta tx) ^. fromDb . _1
+            slots = (localBlockMeta $ pfbMeta block) ^. blockMetaSlotId . fromDb
+            addrs = (localBlockMeta $ pfbMeta block) ^. blockMetaAddressMeta
+            frgns = pfbForeigns block
```
uroboros: at this point we might have to filter foreigns by account also, which would require `PrefilteredBlock` to have `pfbForeigns` as `(Map HdAccountId (Set TxIn))` instead of `Set TxIn`. But, that messes with the lovely symmetry you have between `PrefilteredBlock` per-account and across-accounts (although you could still decide to tolerate the `Map HdAccountId (Set TxIn)` in a per-account PrefilteredBlock?)

src/Cardano/Wallet/Kernel/BListener.hs
KtorZ: Hum.. well.. this version still contains the `isChangeAddress` metadata :upside_down_face: 
We don't really need the prefiltered block for this actually, we do need to answer the `isOurs` question though.

src/Cardano/Wallet/Kernel/DB/Resolved.hs
KtorZ: Cf comment above, this is still on top of the "old" develop, which contains the `isChangeAddr`. I am currently removing this bit because it's not needed anymore. So this is actually the "complex" version of the story, if we ever need to re-implement it. Though I'd rather re-assess what is means for an address to be a "change address" (`resolvedToLocalBlockMeta` still implements the logic as described in the wallet specifications which does not hold for multi-accounts wallets).

src/Cardano/Wallet/Kernel/BListener.hs
uroboros: yes you're right, you _have_ decoupled meta data from prefiltering in a sense, although one could argue that as long as you're using isOurs, you're prefiltering! With `isChangeAddr` gone, metadata has no more need for IsOurs and we get proper decoupling

src/Cardano/Wallet/Kernel/DB/Spec/Pending.hs
KtorZ: Yes, and that accountId "layer" is already created on top (we do create a `Map HdAccountId Pending`) such that we know which pending goes for which account. However, the `Foldable` approach is wrong because I conflate the structure and completely loose the information we get from the Map partitioning. 

I agree with not introducing the `AccountId` concept at this level.

src/Cardano/Wallet/Kernel/PrefilterTx.hs
KtorZ: Agreed that foreigns should be prefiltered by account as well. Though, we don't need any `Map HdAccoundId (Set TxIn)` because, `PrefilteredBlock` out of this function are already indexed by accounts!
Though, we would need to be able to filter out foreigns that are irrelevant for a particular account (as we do for inputs and outputs).

uroboros (approved):
this is excellent! Prefiltering is so much simpler and more general now

KtorZ:
I've added local workers making random transactions (at the pace of 1 TPS) alongside the integration scenarios. I've added an extra edge node connected to the cluster's relay. This edge node hosts 50 wallets separated from the scenarios. Every second, two wallets are randomly picked and a transaction is submitted between the two. Generating noise and non-empty blocks at almost every slot (some transactions may fail if two "already busy" wallets are selected but, on average, there are always a few transactions in each block). Everything run smoothly :tada: , although the integration tests are a bit slower to run since the relay is now under quite heavy load (+ 20% execution time). 

I'll consider making a PR with the "monkey workers" as soon as I've finalized some details with @KtorZ. 
Meanwhile, this increases my confidence about the new approach to prefiltering :+1:

```
core0------------------------------
|.....listen:        127.0.0.1:3000
|.....api address:   127.0.0.1:8080
|.....doc address:   127.0.0.1:8180
|.....system start:  1547563750
-----------------------------------

core1------------------------------
|.....listen:        127.0.0.1:3001
|.....api address:   127.0.0.1:8081
|.....doc address:   127.0.0.1:8181
|.....system start:  1547563750
-----------------------------------

core2------------------------------
|.....listen:        127.0.0.1:3002
|.....api address:   127.0.0.1:8082
|.....doc address:   127.0.0.1:8182
|.....system start:  1547563750
-----------------------------------

relay------------------------------
|.....listen:        127.0.0.1:3100
|.....api address:   127.0.0.1:8083
|.....doc address:   127.0.0.1:8183
|.....system start:  1547563750
-----------------------------------

edge-------------------------------
|.....api address:   127.0.0.1:8084
|.....doc address:   127.0.0.1:8184
|.....system start:  1547563750
-----------------------------------

noise------------------------------
|.....api address:   127.0.0.1:8085
|.....doc address:   127.0.0.1:8185
|.....system start:  1547563750
-----------------------------------

STARTING MONKEYS

API Documentation
  Fetches the documentation from the API
Accounts
  only account's balance can be retrieved (empty balance)
  only account's balance can be retrieved (non empty balance)
  only account's addresses can be retrieved
Addresses
  address is available after it's been created
Transactions
  successful payment appears in the history
  estimate fees of well-formed transaction returns
  payment fails when wallet has no funds
  payment fails when wallet has insufficient funds
  payment fails when wallet cannot cover for fee
  balance is increased when valid key is redemeed
  Reg#141: metadata become inconsistent after a wallet is deleted
  Reg#141: metadata become inconsistent after an account is deleted
  Reg#141: metadata becomes inconsistent after a wallet is deleted and restored
  coin selection: no extra inputs, no extra change
  coin selection: empties a wallet
  coin selection: needs one extra input
  coin selection: needs many extra inputs
Wallets
  WALLETS_DELETE_01 - deleted wallet is not available
  WALLETS_LIST_01 - One can list all wallets without providing any parameters
  WALLETS_UPDATE_01 - updating a wallet persists the update
  WALLETS_UPDATE_PASS_04 - updating password updates date in spendingPasswordLastUpdate
  WALLETS_UTXO_03 - UTxO statistics reflect wallet's inactivity
  WALLETS_UTXO_04 - UTxO statistics reflect wallet's activity
  CreateWallet
    WALLETS_CREATE_01 - One can create/restore previously deleted wallet
    WALLETS_CREATE_02 - One can create/restore wallet without spending password
    WALLETS_CREATE_03 - One can create/restore wallet with spending password
    WALLETS_CREATE_04 - One cannot create/restore wallet if spending password is not hex-encoded string
      spending password is less than 32 bytes / 64 characters
      spending password is more than 32 bytes / 64 characters
      spending password is a number
      spending password is an empty string
      spending password is an array
      spending password is an arbitrary utf-8 string
    WALLETS_CREATE_05 - One cannot create/restore wallet that already exists
    WALLETS_CREATE_06 - One cannot create/restore wallet using less or more than 12 mnemonics which are valid BIP-39
      less than 12 words
      more than 12 words
      empty mnemonic
    WALLETS_CREATE_07 - One cannot create/restore wallet with invalid BIP-39 mnemonics
      Kanji mnemonics / non-English mnemonics
      Invalid mnemonics
    WALLET_CREATE_08 - One cannot create/restore wallet with mnemonics from the API doc
    WALLETS_CREATE_09 - One cannot create/restore wallet with assurance level other than 'strict' or 'normal'
      Arbitrary String
      Empty String
    WALLETS_CREATE_10 - One cannot create/restore without all required parameters (operation)
      Missing operation
      Missing backupPhrase
      Missing assuranceLevel
      Missing name
    WALLETS_CREATE_11 - One can create/restore wallet with name of any length
      long wallet name
      empty name
  RestoreWallet
    WALLETS_CREATE_01 - One can create/restore previously deleted wallet
    WALLETS_CREATE_02 - One can create/restore wallet without spending password
    WALLETS_CREATE_03 - One can create/restore wallet with spending password
    WALLETS_CREATE_04 - One cannot create/restore wallet if spending password is not hex-encoded string
      spending password is less than 32 bytes / 64 characters
      spending password is more than 32 bytes / 64 characters
      spending password is a number
      spending password is an empty string
      spending password is an array
      spending password is an arbitrary utf-8 string
    WALLETS_CREATE_05 - One cannot create/restore wallet that already exists
    WALLETS_CREATE_06 - One cannot create/restore wallet using less or more than 12 mnemonics which are valid BIP-39
      less than 12 words
      more than 12 words
      empty mnemonic
    WALLETS_CREATE_07 - One cannot create/restore wallet with invalid BIP-39 mnemonics
      Kanji mnemonics / non-English mnemonics
      Invalid mnemonics
    WALLET_CREATE_08 - One cannot create/restore wallet with mnemonics from the API doc
    WALLETS_CREATE_09 - One cannot create/restore wallet with assurance level other than 'strict' or 'normal'
      Arbitrary String
      Empty String
    WALLETS_CREATE_10 - One cannot create/restore without all required parameters (operation)
      Missing operation
      Missing backupPhrase
      Missing assuranceLevel
      Missing name
    WALLETS_CREATE_11 - One can create/restore wallet with name of any length
      long wallet name
      empty name
  WALLETS_CREATE_12 - Cannot perform operation other than 'create' or 'restore'
    Invalid operation name
    Empty operation name
Node
  getNodeSettings
  getNodeInfo ForceNtpCheck
  getNodeInfo NoNtpCheck
  nextUpdate


Finished in 763.7305 seconds
```

PR #207, opened 2019-01-07
Filter out zero coins from the list of inputs to collapse

KtorZ:
<p align="right">#191</p>

# Overview

- [x] I have filtered out dust before collapsing coins to compute actual fee 
- [x] I have also added some useful debugging outputs to the invariants when violated.

# Comments

In case where there's any dust in the change output with null coins, we may end up in a case where
there's no more inputs to consume and still, some coins to collapse. This would trigger the last
case and throw and invariant violation. Filtering out those null coins makes it okay

### DIFF
--- a/src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
+++ b/src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs
@@ -322,13 +322,18 @@ computeFee
     -> [Value dom]
     -> Fee dom
 computeFee inps outs chgs =
-    Fee $ collapse (map utxoEntryVal inps) (map outVal outs <> chgs)
+    Fee $ collapse (map utxoEntryVal inps) (filter (> valueZero) $ map outVal outs <> chgs)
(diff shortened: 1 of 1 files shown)


akegalj (approved):
LGTM, although I don't think filter is necessary here ?

I believe line https://github.com/input-output-hk/cardano-wallet/blob/00de79ae6bc3daa4e5e6ebcfcafd17cc86599b4c/src/Cardano/Wallet/Kernel/CoinSelection/Generic/Fees.hs#L343 will actually act as a filter where all null coins will be removed from outputs. Anyway this doesn't seem wrong - but I can't see how it adds anything on top that we already had (as L343 will act as a filter already).

In my opinion your addition will only change which error is being reported in some corner cases (L336 vs L351). But I might be missing something?

KtorZ:
Well, filter is actually the main addition here ^^"...
The line you pointed out is only executed when there are still inputs to consume. The filter here comes precisely in the case where all inputs would have been consumed but some dust remains.

PR #208, opened 2019-01-07
Remove 'isChangeAddress' metadata

KtorZ:
<p align="right">#34</p>

# Overview

- [ ] I have remove 'isChangeAddress' from the address metadata

# Comments

The rational behind this change is two-fold.

1/ This is not a feature actually used. It has landed in the new
data-layer because it was present in the legacy one. Deadalus don't use
it and exchanges don't (to the best of our knowledge; currently running
a survey to double-check this information).

2/ The computed information is actually wrong. The definition of change
address as it was formalized in the wallet specification paper doesn't
account for inter-wallets transactions (since the concept of
multi-accounts doesn't even exist in the wallet specification). However,
following the existing rule, an address actually used as changed when
making a transaction between two owned accounts wouldn't be recognized
as such.

Rather than trying to fix this, we'll just remove it from the backend
and reconsider the question later if and when it comes back on the
table.

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -69,7 +69,6 @@ module Cardano.Wallet.API.V1.Types (
   , AddressLevel
   , addressLevelToWord32
   , word32ToAddressLevel
-  , IsChangeAddress (..)
   -- * Payments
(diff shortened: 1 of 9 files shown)


uroboros (approved):
looks good

Issue #45, opened 2018-11-07
Review Transaction Signing


KtorZ:
# Context

<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

This can't be achieved without the corresponding address private key. When the private key is owned by the wallet, there's not much change in the process apart from choosing the right ed25519 implementation. This imply that this information should be somewhat available in the wallet or corresponding account and passed (at the moment) onto the `deriveLvl2KeyPair` function. Now, it's a matter of picking the right derivation function for the underlying wallet / addresses.

# Decision

<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

Modify existing code to support both derivation scheme in order to derive the correct address private key from a wallet root key.

# Acceptance Criteria

<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

1. - [x] Transaction signing _must_ be done according to the underlying wallet's scheme.
1. - [ ] Testing _should_ be extended to control that nodes can validate (or not) transactions made using both schemes 

# Development Plan

 - [x] I plan on implementing #41 to provide utilities for Address private keys derivation (ie, mimicing old `deriveLvl2KeyPair` https://github.com/input-output-hk/cardano-sl/blob/e9daaa6c98433fefa0d9cd4dd5578f80353eefa1/core/src/Pos/Core/Common/Address.hs#L387)
 - [x] I plan on testing utilities developed by #41 (will be covered within that issue)
 - [x] I intend to modify https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/Kernel/Transactions.hs#L609 `mkSigner` method to handle both derivation scheme cases (with conditional)
      - addresses following new derivation scheme should have empty address payload
      - addresses following old derivation scheme should have some address payload (list of indices, maybe even empty list for root?)
 - [ ] I plan on testing `mkSigner` dirrectly and/or via creating a new transaction. They should work with Addresses of both derivation schemes
[...]

PR #209, opened 2019-01-07
WIP Support new derivation scheme for mkSigner

akegalj:
Adds support for new derivation scheme: bip44 and ed25519 v1 for
mkSigner method. Old derivation scheme (bip32 ed25519 v0) is still
supported. mkSigner makes a decision which derivation scheme to use
based on address payload: if payload is empty we use new derivation
scheme, if payload contains something we use old derivation scheme
(payload is supposed to be account index and address index in this
case)

<p align="right">#45</p>

# Overview

- [x] Added support for new bip44 ed25519 v1 HD wallet scheme
- [x] Backward compatible with old bip32 ed25519 v0 HD wallet scheme
- [ ] Testing should be extended to control that nodes can validate (or not) transactions made using both schemes

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
+++ b/src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
@@ -19,6 +19,7 @@ module Cardano.Wallet.Kernel.Ed25519Bip44
     , deriveAddressPrivateKey
     , derivePublicKey
     , deriveAccountPrivateKey
+    , deriveAddressKeyPair
 
     -- helpers
     , isInternalChange
@@ -135,3 +136,19 @@ deriveAccountPrivateKey passPhrase masterEncPrvKey@(EncryptedSecretKey masterXPr
         -- lvl3 derivation in bip44 is hardened derivation of account' chain
         accountXPrv = deriveXPrv DerivationScheme2 passPhrase coinTypeXPrv accountIx
     pure $ EncryptedSecretKey accountXPrv passHash
+
+-- | Derives address public/private key pair from the given master private key,
+-- using bip44 and derivation scheme 2
+deriveAddressKeyPair
+    :: PassPhrase                            -- Passphrase used to encrypt Master Private Key
+    -> EncryptedSecretKey                    -- Master Private Key
+    -> Word32                                -- Hardened Account Key Index
+    -> ChangeChain                           -- Change chain
+    -> Word32                                -- Non-hardened Address Key Index
+    -> Maybe (PublicKey, EncryptedSecretKey) -- Address Public Key and Private Key pair
+deriveAddressKeyPair passPhrase masterEncPrvKey accountIx changeChain addressIx = do
+    -- derive account private key from master private key
+    accountPrvKey <- deriveAccountPrivateKey passPhrase masterEncPrvKey accountIx
+    -- derive address private key from account private key
+    addressPrvKey <- deriveAddressPrivateKey passPhrase accountPrvKey changeChain addressIx
+    pure (derivePublicKey addressPrvKey, addressPrvKey)
--- a/src/Cardano/Wallet/Kernel/Transactions.hs
+++ b/src/Cardano/Wallet/Kernel/Transactions.hs
@@ -55,8 +55,7 @@ import           Cardano.Wallet.Kernel.DB.InDb
 import           Cardano.Wallet.Kernel.DB.Read as Getters
 import           Cardano.Wallet.Kernel.DB.TxMeta.Types
 import           Cardano.Wallet.Kernel.Ed25519Bip44
-                     (ChangeChain (InternalChain), deriveAccountPrivateKey,
-                     deriveAddressPrivateKey, derivePublicKey)
+                     (ChangeChain (ExternalChain), deriveAddressKeyPair)
 import           Cardano.Wallet.Kernel.Internal (ActiveWallet (..),
                      PassiveWallet (..), walletNode)
 import qualified Cardano.Wallet.Kernel.Internal as Internal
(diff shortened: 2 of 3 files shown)


src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -609,13 +613,43 @@ mkSigner nm spendingPassword (Just esk) snapshot addr =
...
+                                -- Derive address private key from account private key
+                                >>= \accEsk ->
+                                    deriveAddressPrivateKey
+                                        spendingPassword
+                                        accEsk
+                                        InternalChain
+                                        addressIndex
+                            -- Derive Address type from address private key
+                            deriveAddress =
+                                Core.makePubKeyAddressBoot nm . derivePublicKey
```
KtorZ: Hmmm. Do you think it would be worth defining a helper function that derive a public / private address key pair from the root key, like `deriveLvl2KeyPair` ? 
With the content above basically...

akegalj: yes :+1: 

Actually that's first thing that I did - then I made a fallback to the above

akegalj: Fixed here https://github.com/input-output-hk/cardano-wallet/pull/209/commits/f607c3e3cc672ace959eb47311b9e962c4566612 

Adding tests for helper method in progress

akegalj: Added basic hardened/non-hardened tests here https://github.com/input-output-hk/cardano-wallet/pull/209/commits/1ec7b0387bb8d6ee087c7e1243298835c0369fbd

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -223,6 +224,90 @@ prop_deriveAccountPrivateKeyHardened (InfiniteList seed _) passPhrase@(PassPhras
             masterEncPrvKey
             accountIx
 
+-- | Deriving address public/private key pair should always succeed
+-- if account index is hardened (address index is fixed)
+prop_deriveAddressKeyPairAccountKeyHardened
```
akegalj, line 229: These tests don't actually test much but writing more low-level tests would involve bit manipulation and such tests should be part of `cardano-crypto` anyway.

There is a development plan point in #45 to open an issue to refactor these similary looking methods from this test module

src/Cardano/Wallet/Kernel/Ed25519Bip44.hs
```diff
@@ -135,3 +136,19 @@ deriveAccountPrivateKey passPhrase masterEncPrvKey@(EncryptedSecretKey masterXPr
...
+    -> Word32                                -- Hardened Account Key Index
+    -> ChangeChain                           -- Change chain
+    -> Word32                                -- Non-hardened Address Key Index
+    -> Maybe (PublicKey, EncryptedSecretKey) -- Address Public Key and Private Key pair
+deriveAddressKeyPair passPhrase masterEncPrvKey accountIx changeChain addressIx = do
+    -- derive account private key from master private key
+    accountPrvKey <- deriveAccountPrivateKey passPhrase masterEncPrvKey accountIx
+    -- derive address private key from account private key
+    addressPrvKey <- deriveAddressPrivateKey passPhrase accountPrvKey changeChain addressIx
+    pure (derivePublicKey addressPrvKey, addressPrvKey)
```
KtorZ, line 154: How sweet :heart_eyes: !

src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -609,13 +612,32 @@ mkSigner nm spendingPassword (Just esk) snapshot addr =
...
-                                             spendingPassword
-                                             esk
-                                             accountIndex
-                                             addressIndex
+                mAddressPayload = hdAddr ^. HD.hdAddressAddress
+                                          . fromDb
+                                          . to Core.addrAttributes
+                                          . to attrData
+                                          . to Core.aaPkDerivationPath
+                res = case mAddressPayload of
```
KtorZ, line 620: So okay, looking at the payload is _most probably_ okay for now (as suggested). We may refine this decision later because this _could_ allow for signing transaction with both old and new addresses (since we pick the signer base on the address and not based on the wallet). 
It _shouldn't_ matter in practice as we won't be constructing wallets with both scheme in them. Still, I wonder if it wouldn't be less ad-hoc to later include some flag in our wallet representation to clearly specify the derivation scheme.

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -223,6 +224,90 @@ prop_deriveAccountPrivateKeyHardened (InfiniteList seed _) passPhrase@(PassPhras
...
+    property $ isJust addrPrvKey
+  where
+    masterEncPrvKey = mkEncSecretWithSaltUnsafe emptySalt passPhrase $ generate (BS.pack $ take 32 seed) passBytes
+    addrPrvKey =
+        deriveAddressKeyPair
+            passPhrase
+            masterEncPrvKey
+            accountIx
+            changeChain
+            addressIx
```
KtorZ, line 309: Yeah right. I believe we could just collapse them into one test (the success scenario) with both `addressIx` and `accountIx` as parameters. Or, instead, have them instead of the others that are based on the intermediate functions. 
It's a bit redundant to have them both, I agree.

KtorZ (approved):

src/Cardano/Wallet/Kernel/Transactions.hs
akegalj: > this could allow for signing transaction with both old and new addresses

yes, that's true. If transaction contains mixed old and new addresses transaction will be signed with both signers

> It shouldn't matter in practice as we won't be constructing wallets with both scheme in them. 

yes, that's also true. Old wallets will continue to use old scheme, while new wallets will have whole HD wallet tree constructed with new scheme - so in practice won't matter

> I wonder if it wouldn't be less ad-hoc to later include some flag in our wallet representation to clearly specify the derivation scheme

I think this is a good idea as it removes ambiguity (at least from coders perspective) and makes things more explicit. I will open an issue about this

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: yes - good point :+1:

src/Cardano/Wallet/Kernel/Transactions.hs
akegalj: added as a booling point to #45 development plan

test/unit/Test/Spec/Ed25519Bip44.hs
akegalj: added as a booling point to #45 development plan

Issue #151, opened 2018-12-11
Updates can't actually be queued and postponed indefinitely


KtorZ:
| Release | Operating System       | Cause |
| ---     | ---                    | ---   |
| next    | Windows & OSX & Linux) | Code |

# Context
 
 

Here below is describe how updates are currently handled in `cardano-sl` (in
cunjunction with `cardano-wallet` still used as a middle-man for communicating
between `Daedalus` and its underlying node).

At epoch boundaries, node may extract and process confirmed update proposals
received from the network. A worker running asynchronously regularly checks for 
updates and eventually triggers the `downloadUpdate` function below:

**Pos.Network.Update.Download**  
```hs
downloadUpdate :: forall ctx m . UpdateMode ctx m => ConfirmedProposalState -> m ()
```

In its normal mode, this function does two major things (beside a bunch of checks 
on the update's version and target system):

- It downloads the corresponding update and write it to disk at a given location.
  (the location depends on a CLI option given at runtime). 

- It put the corresponding `ConfirmedProposalState` in an `MVar` shared with the 
  wallet backend. 

**Pos.DB.Update.Context**  
```hs
data UpdateContext = UpdateContext
    {
    -- | A semaphore which is unlocked when update data is downloaded and
    -- ready to apply.
      ucDownloadedUpdate :: !(MVar ConfirmedProposalState)

    -- | A lock which allows only one thread to download an update.
    , ucDownloadLock     :: !(MVar ())

    -- | In-memory state of update-system-as-block-component.
    , ucMemState         :: !MemVar
    }
```

The wallet backend is the only listener on this `ucDownloadedUpdate` `MVar` which it polls
regularly checking for available updates. Pretty much as soon as the `MVar` is filled, it
is taken by the wallet which stores the corresponding update in a queue persisted in acid-state.

From there, the frontend which is regularly polling the wallet backend about the next available 
update get informed that a new update is available in the queue. The user is prompted
with a choice of either:

- Applying the update by shutting down the node immediately with an exit code 20 (which tells the launcher to consequently restart the node) 
- Postponing the shut down, which will apply the update only when the node restarts

Regardless of the choice, the update is popped out of the wallet's queue will be applied when 
the node restarts. 

> NOTE: 
> For 'core' nodes with no wallet running, the behavior is slightly different. An extra worker
> is started which role is to restart the node immediately when an update is available.
> 
> **Pos.Worker.Update**  
> ```hs
[...]

PR #210, opened 2019-01-07
Implement node API tests

parsonsmatt:
<p align="right">#</p>

# Overview

- [x] I have added tests for the reverse proxy endpoints to ensure they work and that the wallet can connect.

# Comments


test/integration/Main.hs
```diff
@@ -39,6 +39,7 @@ main = do
     -- Run tests
     hspec $ do
         describe "API Documentation" $ Documentation.spec dClient
+        describe "Node Endpoints" $ Node.spec wClient
```
KtorZ: What about using the integration test framework and write scenarios using the provided DSL :thinking: ? 
Or are you purposely separating this for now in regards to the decoupling ?

test/integration/Test/Integration/Framework/Cluster.hs
```diff
@@ -43,7 +43,17 @@ import           Pos.Util.Wlog.Compatibility (usingNamedPureLogger)
...
+-- correspond to command line arguments that would be passed to underlying
+-- processes.
+--
+-- As an example, if you wanted to enable the @--wallet-debug@ option for the
+-- underlying node, you would add an entry in this list:
+--
+-- > ("WALLET_DEBUG", "True")
+--
+-- Underscores (@_@) are converted to hyphens (@-@), the text is lowercased, and
+-- a leading @--@ is added.
```
KtorZ, line 56: :+1:

parsonsmatt:
Tests pass locally with https://github.com/input-output-hk/cardano-sl/pull/3995

test/integration/Main.hs
parsonsmatt: I looked at the scenario DSL stuff but since the requests were so simple I just didn't see a need for them

akegalj (commented):
Changes look good to me

Stylish-haskell and weeder are complaining though

test/integration/Main.hs
KtorZ: Well, the DSL also gives control to QA over the testing. That's actually one of its main goal.
Therefore, I'd rather stick with that to allow Piotr to maintain / extend that if necessary.

Plus, if we start having divergent approaches, we'll end up like before where every integration test module has its own way of doing stuff which makes the overall thing hard to review and hard to maintain :/ 

Incidentally, we do want each test to be linked / tagged to some requirements. Make sure to sync with @piotr-iohk on that if they aren't listed in the capabilities and requirements matrix already.

parsonsmatt: These tests are just a smoke test to ensure that the integration framework can communicate with the underlying node through the proxy, and don't really test the wallet behavior at all. When we move to V2, these endpoints won't even be supported anymore.

I can shift them to the integration DSL if you think it's important to do so, but it matches the `Documentation` tests more than the other endpoints IMO.

KtorZ: Indeed, so they make very valid integration tests in the format we currently have them and within the integration DSL :sweat_smile: ! 
Let's have them written in a similar fashion to keep things consistent :pray:

KtorZ: (*Also, there might not be any V2. Or not in way we imagine it.)

_Spoiler: Nothing clearly agreed on yet, but direction is moving towards GraphQL + SQL backend in the long run..._

piotr-iohk: @parsonsmatt, @KtorZ:

I was actually a bit confused by what Node API is :). If you look at the capability matrix -> https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk/edit#gid=0 you'll see that I'm trying to formulate more detailed requirements/capabilities about each endpoint and my map for that is API doc -> https://input-output-hk.github.io/cardano-wallet. So, in other words, I'm trying to get that from end user point of view.

If possible I would like to continue to have a structure in Integration tests that corresponds to the structure in the API doc, i.e.
API category `Wallets`    -> tests in `..\Integration\Scenario\Wallets.hs`
API category `Accounts` -> tests in `..\Integration\Scenario\Accounts.hs`
...
etc.

This way it is clear, readable and less chance something gets omitted :)

So yes, I was confused because there is no `Node` category in the API doc. :) Looking at the added tests here I see that such endpoints are addressed (if I'm not mistaken):


`/api/v1/node-info` - from API doc category `Info`
`/api/v1/node-settings` - from API doc categoty `Settings`
`/api/internal/next-update` - from API doc category `Internal`

So my suggestion would be to put them in their respective files in the Scenario folder (to correspond with the API doc categories). In principle I'd also prefer to have all integration tests in DSL for better/easier maintenance and readability (for me). 

BTW. Also some changes have been done in that area of the API here https://github.com/input-output-hk/cardano-wallet/issues/151 ? (at least that was my understanding from the `Resolution Plan`) :) Good opportunity to add new tests and/or update API doc? (just throwing idea... ;))

parsonsmatt: @piotr-iohk you got it -- the Wallet API included some endpoints that were primarily for communicating to the underlying Node processes. We decided to factor these out and give the Node it's own API for these things. The Wallet API still must support the endpoints, but now instead of doing the information itself, it delegates to the Node's API.

The Node integration tests are therefore about ensuring that the Wallet API can successfully communicate to the Node API for the endpoints that were moved to the Node's API.

parsonsmatt: @KtorZ Alright, reimplemented as scenario :)

KtorZ (approved):

PR #216, opened 2019-01-09
HdAddress type hierarchy for EOS-wallets.

denisshevchenko:
<p align="right">#32</p>

# Overview

- [x] `EosHdWallets` uses the field with the same type hierarchy for storing addresses for EOS-wallets.
- [x] Function `createEosHdAddress` allows to store new address (that will be derived from account's public key)
- [x] Function `addressesByAccountId` allows to get all addresses in account to get the next (sequential) address index.

# Comments

After this PR will be merged, I plan to add new endpoint for creating new address for EOS-wallet.

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/EosHdWallet/Create.hs
+++ b/src/Cardano/Wallet/Kernel/DB/EosHdWallet/Create.hs
@@ -5,9 +5,11 @@ module Cardano.Wallet.Kernel.DB.EosHdWallet.Create (
     -- * Errors
     CreateEosHdRootError(..)
   , CreateEosHdAccountError(..)
+  , CreateEosHdAddressError(..)
     -- * Functions
   , createEosHdRoot
   , createEosHdAccount
+  , createEosHdAddress
   ) where
 
 import           Universum
@@ -21,6 +23,7 @@ import qualified Formatting.Buildable
 import           Pos.Crypto (PublicKey)
 
 import           Cardano.Wallet.Kernel.DB.EosHdWallet
+import           Cardano.Wallet.Kernel.DB.HdWallet
 import           Cardano.Wallet.Kernel.DB.Util.AcidState
 import qualified Cardano.Wallet.Kernel.DB.Util.IxSet as IxSet
 import           Cardano.Wallet.Kernel.EosWalletId (EosWalletId)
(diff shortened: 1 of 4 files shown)


src/Cardano/Wallet/Kernel/DB/EosHdWallet/Create.hs
```diff
@@ -73,6 +84,21 @@ createEosHdAccount eosHdAccount = do
...
+createEosHdAddress :: HdAddress -> Update' CreateEosHdAddressError EosHdWallets ()
+createEosHdAddress hdAddress = do
+    -- Create the new address
+    zoom eosHdWalletsAddresses $ do
+        exists <- gets $ IxSet.member addrId
+        when exists $ throwError $ CreateEosHdAddressExists addrId
+        -- We don't have 'AutoIncrementKey' for 'EosHdAccount', but since we use
+        -- the same type hierarchy as for 'HdAddress', we have to provide fake
+        -- 'AutoIncrementKey' value.
+        let fakeAutoIncrementKey = IxSet.AutoIncrementKey 0
```
KtorZ, line 97: Why don't we exactly :thinking: ? What is this `AutoIncrementKey` used for?

denisshevchenko: > What is this AutoIncrementKey used for?

I have no idea. :-) But since we use the _same_ hierarchy - it requires `AutoIncrementKey`.

KtorZ: Then, could you get an idea :smile:?

denisshevchenko: We get this increment counter here: https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/Kernel/DB/HdWallet/Create.hs#L103. This counter is a part of corresponding `HdAccount`: https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/Kernel/DB/HdWallet.hs#L389-L391. So, during new address creation we take current counter and use it, after that - increment it: https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/Kernel/DB/HdWallet/Create.hs#L111-L115.

AFAIU, we don't need this counter for account in EOS-wallet, because all addresses are sequential, not random. I see in the comment for `HdAccount`:

> A local counter used to generate new 'AutoIncrementKey' for addresses.

But if addresses' indexes in EOS-wallet are sequential - why do we need a counter for them?

Technically, AFAIU, this local counter is needed for efficiency: https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/WalletLayer/Kernel/Addresses.hs#L57-L90. But probably we shouldn't think about efficiency of EOS-wallets, at least for now..

denisshevchenko:
Closed because of inappropriate changes (new PRs will be opened instead).

Issue #213, opened 2019-01-09
Paginated result from the API unexpectedly returns data when page parameter is close to it's boundary value


piotr-iohk:
| Release | Operating System       | Cause |
| ---     | ---                    | ---   |
| next    | Windows & OSX & Linux| Code   |

# Context
The results returned by the API can be paginated in case the API returns a list of items (i.e. wallets, addresses, transactions etc.). One can filter them by `page` and `per_page` parameters. `per_page` is bounded between [1..50] and `page` should be >= 1 (according to api doc -> https://input-output-hk.github.io/cardano-wallet/#tag/Wallets%2Fpaths%2F~1api~1v1~1wallets%2Fget) but in reality it is bounded between [1..9223372036854775807].

When using `page` value that is close to it's boundary value then the data is actually returned, but shouldn't.

One can tell API to return particular page with the results and if there are no results on the page then empty set is returned. I.e. say we have 10 wallets. We can tell API to return page 100 `GET /api/v1/wallets?page=100`. Such request should return `"data": []` as the wallets are available on the `page=1`

The _problem_ is that when you provide value for `page` that is between `[1000000000000000000..9223372036854775807]` it returns all results in such a case. That is unexpected.

Please also see similar #214.

# Steps to Reproduce

0. Have some wallets available.
1. Run, for instance:
```
curl -X GET https://localhost:8090/api/v1/wallets?page=1000000000000000000 \
  -H "Accept: application/json; charset=utf-8" \
  -H "Content-Type: application/json; charset=utf-8" \
  --cert ~/.local/share/Daedalus/testnet/tls/client/client.pem \
  --cacert ~/.local/share/Daedalus/testnet/tls/client/ca.crt
``` 

## Expected behavior

The result should be `"data": []`

## Actual behavior

The API returns all wallets that are available.
 
<A clear and concise description of what you observe instead. If applicable add screenshots to help explain your problem.>

---

# Resolution Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #?       | `develop` |

# QA

PR #217, opened 2019-01-09
Integration tests for WALLETS_LIST

piotr-iohk:
<p align="right">#172</p>

# Overview

- [x] Updated tests for WALLETS_LIST_01
- [x] I have added tests for WALLETS_LIST_02
- [x] I have added tests for WALLETS_LIST_03 (implemented but made partially pending due to #220)
- [x] I have added tests for WALLETS_LIST_04  (implemented but made pending due to #220)

# Comments
This can be addressed in subsequent PR or in this one, but:
 
- I'm missing method like `expectClientHttpError` method for asserting the `ClientHttpError` message -> addressed under #221.

- ~also to move forward I need to get some examples on how to use `Sort` and `Filter` in the safe/typed Client~
After thinking about it I think I'll go for `unsafeRequest` after all for sort and filter in order to have better code coverage... if I'm not mistaken the type-safe client would not reach all cases here -> https://github.com/input-output-hk/cardano-wallet/blob/d28b3f7b24ed3f163d46f8c18a91a117c7e5801e/src/Cardano/Wallet/API/Request/Sort.hs#L208


test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +43,146 @@ spec = do
...
+                        setup $ defaultSetup
+                            & walletName .~ show (name :: Int)
+
+                    response <- request $ Client.getWalletIndexFilterSorts
+                        $- page
+                        $- per_page
+                        $- NoFilters
+                        $- NoSorts
+                    verify response expectations
+            in case (title == "1 wallet; page=9223372036854775807 & per_page=50 => 0 wallets returned") of
```
piotr-iohk: This test fails due to BUG https://github.com/input-output-hk/cardano-wallet/issues/213, hence made this scenario as pending...

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +43,146 @@ spec = do
...
+                True      -> xscenario title $ test
+                False     -> scenario title $ test
+
```
piotr-iohk, line 110: This particular test scenario fails due to BUG https://github.com/input-output-hk/cardano-wallet/issues/213, hence made it pending...

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +43,146 @@ spec = do
...
+            resp <- unsafeRequest ("GET", fromString endpoint) $ Just $ [json|{}|]
+            verify (resp :: Either ClientError Wallet) expectations
+
+    xscenario "WALLETS_LIST_04 - One can sort results by 'balance' and 'created_at'" $ do
+
+        _ <- forM (zip [1,2,3] [3,2,1]) $ \(name, coins) -> do
+            setup $ defaultSetup
+                & walletName .~ show (name :: Int)
+                & initialCoins .~ [coins]
+
```
piotr-iohk, line 250: For now I made this one pending because setting up 3 wallets here was causing:
```
       user error (expectTxStatusEventually: waited too long for statuses.)
```

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +43,146 @@ spec = do
...
+                  , 1
+                  , Just (Client.Page 9223372036854775807)
+                  , Just (Client.PerPage 50)
+                  , [ expectSuccess
+                    , expectDataListSizeEqual 0
+                    ]
+                  )
+                ]
+
+        forM_ matrix $ \(title, wallets_number, page, per_page, expectations) ->
```
parsonsmatt:
```suggestion
        forM_ matrix $ \(title, walletsNumber, page, perPage, expectations) ->
```

Haskell variables should be camel cased.

test/integration/Test/Integration/Scenario/Wallets.hs
parsonsmatt: You can mark a scenario pending with `pendingWith`. It may be necessary to write a lifted version of that function. But it should be:

```haskell
scenario title $ do
  pendingWith "This test fails due to bug #213"
  test
```

That way we get a note in the code that it is pending.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +43,146 @@ spec = do
...
+                        setup $ defaultSetup
+                            & walletName .~ show (name :: Int)
+
+                    response <- request $ Client.getWalletIndexFilterSorts
+                        $- page
+                        $- per_page
+                        $- NoFilters
+                        $- NoSorts
+                    verify response expectations
+            in case (title == "1 wallet; page=9223372036854775807 & per_page=50 => 0 wallets returned") of
```
parsonsmatt: To remove ambiguity, this string should be factored out to a common value and reused rather than copy-pasted.

test/integration/Test/Integration/Scenario/Wallets.hs
parsonsmatt: Rather than `xscenario`, use `scenario`, and `pendingWith` with a note that says *why* it is failing.

parsonsmatt (dismissed):

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -157,6 +161,18 @@ xscenario
...
+-- scenarios from execution with a reason, like:
+--
+--      scenario title $ do
+--          pendingWith "This test fails due to bug #213"
+--          test
+pendingWith
+    :: (MonadIO m, MonadFail m)
+    => String
+    -> m ()
+pendingWith = liftIO . H.pendingWith
```
KtorZ, line 180: :+1:

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -347,6 +363,35 @@ walletName = typed
...
+
+-- | Expects data list returned by the API to be of certain length
+expectDataListSizeEqual
+    :: (MonadIO m, MonadFail m)
+    => Int
+    -> Either ClientError [a]
+    -> m ()
+expectDataListSizeEqual l = \case
+    Left e  -> wantedSuccessButError e
+    Right s -> length s `shouldBe` l
```
KtorZ: I'd recommend making this one slightly more generic such that it works with both `List` and `NonEmpty.List`. 
```suggestion
import qualified Data.Foldable as F

expectListSizeEqual
    :: (MonadIO m, MonadFail m, Foldable xs)
    => Int
    -> Either ClientError (xs a)
    -> m ()
expectListSizeEqual l = \case
    Left e   -> wantedSuccessButError e
    Right xs -> length (F.toList xs) `shouldBe` l
```

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -347,6 +363,35 @@ walletName = typed
...
+expectDataListItemFieldEqual
+    :: (MonadIO m, MonadFail m, Show a, Eq a)
+    => Int
+    -> Lens' s a
+    -> a
+    -> Either ClientError [s]
+    -> m ()
+expectDataListItemFieldEqual i getter a = \case
+    Left e  -> wantedSuccessButError e
+    Right s -> (view getter $ s!!(i - 1)) `shouldBe` a
```
KtorZ: Sweet :smile: !
Yet, this will blow up if the list is too small and also, elements are 0-indexed, it will be better to preserve that in the function arguments to not make it counter-intuitive.

Also, I'd make more re-use of the existing code and not re-write the getter machinery.

```suggestion
expectListItemFieldEqual 
    :: (MonadIO m, MonadFail m, Show a, Eq a)
    => Int
    -> Lens' s a
    -> a
    -> Either ClientError [s]
    -> m ()
expectListItemFieldEqual i getter a = \case
    Left _ -> expectFieldEqual getter a
    Right s 
      | length s > i -> expectFieldEqual getter a (Just (s !! i))
      | otherwise    -> fail $
        "expectListItemFieldEqual: trying to access the #" <> show i <> 
        " element from a list but there's none! Here's the list: " <> 
        show s
```

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +43,133 @@ spec = do
...
+                        $- page
+                        $- perPage
+                        $- NoFilters
+                        $- NoSorts
+                    verify response expectations
+            in case (title == failingScenario) of
+                    True  -> scenario title $ do
+                                pendingWith "Test fails due to bug #213"
+                                test
+                    False -> scenario title $ test
```
KtorZ: `pendingWith` actually "interrupts" the test scenario. So you could just avoid the `let ... in` binding an go with:

```
do
    forM_ ([1..walletsNumber]) $ \name -> do
        when (title == failingScenario) $
            pendingWith ...

        -- rest of the rest
```

KtorZ (changes requested):

piotr-iohk:
I have implemented requested changes and added new tests for WALLETS_LIST_03 and 04(Part of the tests is made pending due to #220), rebased. . Please review @KtorZ and/or @parsonsmatt. Thanks!

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +42,464 @@ spec = do
...
+
+        fixtures <- forM ([1,2]) $ \name -> do
+            setup $ defaultSetup
+                & walletName .~ show (name :: Int)
+
+        let walletIds =
+                [ (fixtures !! 0) ^. wallet . walletId
+                , (fixtures !! 1) ^. wallet . walletId
+                ]
+        let endpoint = "api/v1/wallets?id=" <+> ( fromWalletId $ walletIds !! 0 )
```
KtorZ: I believe you don't need this custom `<+>` operator here :) If you just want to concatenate two `Text`, you can leverage their `Semigroup` instance and go with `<>` --> 

```
let endpoint = "api/v1/wallets?id=" <> fromWalletId (walletIds !! 0)
```

KtorZ: Why are you actually defining those multiple walletIds actually ^^? The previous let binding feels superfluous, and this could just be:

```hs
let wId = (fixtures !! 0) ^. wallet . walletId
let endpoint = "api/v1/wallets?id=" <> fromWalletId wId
```

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +42,464 @@ spec = do
...
+
+        resp <- unsafeRequest ("GET", endpoint) $ Just $ [json|{}|]
```
KtorZ: If there's no body, you can just, not give any :p 

```hs
resp <- unsafeRequest ("GET", endpoint) Nothing
```

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +42,464 @@ spec = do
...
+
+        fixtures <- forM ([1,2]) $ \name -> do
+            setup $ defaultSetup
+                & walletName .~ show (name :: Int)
+
+        let walletIds =
+                [ (fixtures !! 0) ^. wallet . walletId
+                , (fixtures !! 1) ^. wallet . walletId
+                ]
+        let endpoint = "api/v1/wallets?id=EQ%5B" <+> ( fromWalletId $ walletIds !! 0 ) <+> ("%5D" :: Text)
```
KtorZ: Arf :( .. I think might want to have those encoded brackets abstracted over some DSL combinator at some point. I think there's quite a few things that could be written in a nicer way with those pagination / filter / sort queries. 

I'll open a ticket to provide you with the necessary tools for that :+1:

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +42,464 @@ spec = do
...
+                    , expectListItemFieldEqual 0 amount 9
+                    , expectListItemFieldEqual 1 amount 3
+                    , expectListItemFieldEqual 2 amount 0
+                    ]
+                  )
+                ]
+
+        forM_ matrix $ \(endpoint, expectations) -> scenario endpoint $ do
+            pendingWith "Test fails due to bug #220"
+            forM_ ([3,6,6,9,0]) $ \coins -> do
```
KtorZ, line 237: So, as mentioned on slack, the last one here is a `0` coin. So, we can't expect the tx to be resolved at some point. It's a pity though that the backend doesn't reject the request in a first place.

KtorZ: Hence the failure.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +42,464 @@ spec = do
...
+        -- LT[value] : allow resource with attribute less than the value
+
+        fixtures <- forM ([1,2]) $ \name -> do
+            setup $ defaultSetup
+                & walletName .~ show (name :: Int)
+
+        let sortedWalletIds =
+                sort [ fromWalletId $ (fixtures !! 0) ^. wallet . walletId
+                     , fromWalletId $ (fixtures !! 1) ^. wallet . walletId
+                     ]
```
KtorZ, line 293: Hmm.. To be totally fair, I'd have sorted them "by hand". Especially because both tests and the actual implementation are written in Haskell and probably uses the same sorting function under the hood. Therefore, if our `Ord` instance was messed up, this test would catch it because it would just reproduce what's the backend is also doing. 

Yet, that's somewhat tough because we don't know the wallet ids upfront :upside_down_face: Ids are generated deterministically from the wallet secret key, so it's still possible to force some mnemonics and compute their corresponding wallet ids upfront.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +42,464 @@ spec = do
...
+        -- GT[value] : allow objects with an attribute greater than the value
+
+        fixtures <- forM ([1,2]) $ \name -> do
+            setup $ defaultSetup
+                & walletName .~ show (name :: Int)
+
+        let sortedWalletIds =
+                sort [ fromWalletId $ (fixtures !! 0) ^. wallet . walletId
+                     , fromWalletId $ (fixtures !! 1) ^. wallet . walletId
+                     ]
```
KtorZ, line 313: Same remark as before :)

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -43,11 +42,464 @@ spec = do
...
+
+        let walletIds =
+                     [ fromWalletId $ (fixtures !! 0) ^. wallet . walletId
+                     , fromWalletId $ (fixtures !! 1) ^. wallet . walletId
+                     , fromWalletId $ (fixtures !! 2) ^. wallet . walletId
+                     ]
+        let sortedWalletIds = sort walletIds
+
+        let endpoint = "api/v1/wallets?sort_by=created_at&id=RANGE%5B" <+> ( sortedWalletIds !! 0 )
+                        <+> "," <+> ( sortedWalletIds !! 2 )  <+> ("%5D" :: Text)
```
KtorZ: Why sorting the wallets by ids here? The sort given in the query is by `created_at` ?

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -405,6 +856,10 @@ spec = do
             }|]
             verify (response :: Either ClientError Wallet) expectations
   where
+
+    fromWalletId :: Client.WalletId -> Text
+    fromWalletId (Client.WalletId a) = a
```
KtorZ, line 861: Looks like we could use a `Lens` here to unwrap things like this. This would play nicely with `fixture ^. wallet . walletId . asText` 

(not asking you to implement that, more like leaving a note for myself :p )

KtorZ (changes requested):

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: OK, however, I have experienced the issue from #220 (also pretty much repeatedly) when I had no `0` coin with something like:
```
            forM_ (zip [1,2,3] [3,2,1]) $ \(name, coins) -> do
                setup $ defaultSetup
                    & walletName .~ show (name :: Int)
                    & initialCoins .~ [coins]
```

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Yes. Sorting the ids because I want to use `RANGE[smallest, biggest]` and then I sort the results by `created_at` to make sure all the expected wallet id's are present in the returned results (because I know they will be presented in the descending order)...

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Not sure what you mean by sorting "by hand" :) Pls explain.

Also, I suppose `sort` is a _native_ Haskell method, so I trust it works fine. :)

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: >Why are you actually defining those multiple walletIds actually ^^?

Well, that's part of the test actually. Let's assume that filtering by id stopped working and returns all wallet's even if I specify particular one. If I have only one wallet in this test, it will not catch that :)

test/integration/Test/Integration/Scenario/Wallets.hs
KtorZ: Well, `sort` is native yes, but it relies on the `Ord` instance of type you're trying to sort. 

```hs
sort :: Ord a => [a] -> [a]
```

Therefore:

```hs
data WalletId = WalletId String deriving (Show, Eq)

instance Ord WalletId where
  (WalletId a) <= (WalletId b) = a >= b

main :: IO ()
main = do
  print $ sort $ [WalletId "a", WalletId "b", WalletId "c"]
  -- [WalletId "c",WalletId "b",WalletId "a"]
```

I suppose a comment would be enough here since we're _probably_ (I hope) using the stock `Ord` instance for the `WalletId` and that we aren't defining it by hand.

KtorZ (approved):

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Ok, got it.
However I was not sorting `WalletId` but actually a `Text` since: `fromWalletId :: Client.WalletId -> Text`.

Maybe the name of the list should have been more descriptive then, like: `sortedWalletIdsText` instead of `sortedWalletIds`...

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: OK, read your code snippet now again, and... indeed, the additional `walletIds` list here is superfluous... :)

Issue #87, opened 2018-11-22
Remove protocol parameters from `NodeStateAdaptor`


paweljakubas:
# Context

The wallet backend requires to know a few of the underlying blockchain protocol
parameters. Those parameters may change over time through updates, at epoch
boundaries. At the moment, nodes are keeping track of those parameters, and the
wallet reads them using some shared context via the `NodeStateAdaptor`.

# Decision

- Extend the monitoring API of cardano node to expose protocol parameters (This
  can be done by extending the `/api/v1/node-settings` API endpoint.

- Use the extended monitoring API instead of the current `NodeStateAdaptor` to
  fetch protocol parameters and removes them from the current
  `NodeStateAdaptor`.

- Add a thin caching layer on top of our http client to only fetch parameters
  if the epoch has changed. This cache doesn't need to be persisted and be
  stored in plain memory and its usage should be transparent for the wallet
  backend.

# Acceptance Criterias

1. Protocol parameters available through the `NodeStateAdaptor` _must_ be
   identified and made available through the node monitoring API
2. The wallet backend _must_ resolve those parameters from the node monitoring
   API its connected to
3. The wallet backend _should_ cache the results from the API and only refresh
   them at epoch boundaries.

---

# Development Plan

## Proof of concept

Quick hack to see if everything works when protocol parameters are sent via the API.

- [x] List all fields of NodeStateAdaptor / uses (and later: motivate which ones we care about)
- [x] Serve some NodeStateAdaptor field via the node API
- [x] Call Node API from Wallet as proof of concept
- [x] Have integration tests compiling
- [x] Have integration not failing because of TLS settings
- [x] Have an idea for caching

## Wire it up in a proper way
-  ~Revisit TLS Settings and ensure no paths are hard-coded~ Matt did this
-  ~Ensure no localhost/port etc are hardcoded~ Matt did this
- [x] Ensure JSON instances are properly decoupled (they are handwritten, so…)
- [x] Merge protocol-parameters endpoint with node-settings. I'm delaying this to have more options for caching.
- [x] Deal with the first easy protocol parameters
- [x] Deal with `getCoreConfig` Not needed
-  ~Deal with `withNodeState`~ Separate issue
- [x] Actually start the server from the wallet itself (this was missed initially)

## Finalizing
-  ~Caching and invalidation~ Separate issue
- [x] Error handling — think so, and retries
[...]

PR #218, opened 2019-01-11
Make NodeStateAdaptor use the Node API

Anviking:
<p align="right">#87</p>

# Overview

Retrieve the following protocol parameters through the node API:

- [x] `getTipSlotId`
- [x] `getMaxTxSize`
- [x] `getFeePolicy`
- [x] `getSecurityParameter`
- [x] `getSlotCount`

# Comments

- Tests pass, so I guess we could get this into develop?
- This is the second take on this
- Caching will be addressed in another PR

### DIFF
--- a/src/Cardano/Wallet/Action.hs
+++ b/src/Cardano/Wallet/Action.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 module Cardano.Wallet.Action (actionWithWallet) where
 
 import           Universum
@@ -68,7 +69,7 @@ actionWithWallet
             genesisConfig
             nodeRes
             ntpStatus
-            eta
+            NodeStateAdaptor.toMonadIO
             nodeClient
 
     liftIO $ Keystore.bracketLegacyKeystore userSecret $ \keystore -> do
@@ -93,14 +94,6 @@ actionWithWallet
                 walletLayer
                 (runNode genesisConfig txpConfig nodeRes plugs)
   where
-
-    -- TODO(@anviking #87): proper error handling
-    eta :: MonadIO m => Show e => ExceptT e m a -> m a
-    eta e = do
-        x <- runExceptT e
-        case x of
-            Right a   -> return a
-            Left  err -> error $ "Error talking to the Node: " <> (show err)
 
     plugins :: (PassiveWalletLayer IO, PassiveWallet)
             -> NodeHttpClient
@@ -140,3 +133,4 @@ actionWithWallet
       where
         loggerName :: LoggerName
         loggerName = lpDefaultName . bpLoggingParams . npBaseParams $ nodeParams
+
--- a/src/Cardano/Wallet/Action.hs
+++ b/src/Cardano/Wallet/Action.hs
@@ -2,6 +2,7 @@ module Cardano.Wallet.Action (actionWithWallet) where
 
 import           Universum
 
+import qualified Data.ByteString.Char8 as BS8
 import           Ntp.Client (NtpConfiguration, ntpClientSettings, withNtpClient)
 
 import           Pos.Chain.Genesis as Genesis (Config (..))
@@ -17,6 +18,7 @@ import           Pos.Util.Wlog (LoggerName, Severity (..), logInfo, logMessage,
                      usingLoggerName)
 import           Pos.WorkMode (EmptyMempoolExt)
 
+import           Cardano.Node.Client (NodeHttpClient)
 import qualified Cardano.Wallet.API.V1.Headers as Headers
 import           Cardano.Wallet.Kernel (PassiveWallet)
 import qualified Cardano.Wallet.Kernel as Kernel
@@ -25,8 +27,9 @@ import qualified Cardano.Wallet.Kernel.Keystore as Keystore
 import           Cardano.Wallet.Kernel.Migration (migrateLegacyDataLayer)
 import qualified Cardano.Wallet.Kernel.Mode as Kernel.Mode
 import qualified Cardano.Wallet.Kernel.NodeStateAdaptor as NodeStateAdaptor
-import           Cardano.Wallet.Server.CLI (NewWalletBackendParams,
-                     getFullMigrationFlag, getWalletDbOptions, walletDbPath,
+import           Cardano.Wallet.Server.CLI (NewWalletBackendParams (..),
+                     WalletBackendParams (..), getFullMigrationFlag,
+                     getWalletDbOptions, walletDbPath, walletNodeAddress,
                      walletRebuildDb)
 import           Cardano.Wallet.Server.Middlewares
                      (faultInjectionHandleIgnoreAPI, throttleMiddleware,
@@ -48,14 +51,27 @@ actionWithWallet
     -> SscParams
     -> NodeResources EmptyMempoolExt
     -> IO ()
-actionWithWallet params genesisConfig walletConfig txpConfig ntpConfig nodeParams _ nodeRes = do
+actionWithWallet
+    params@(NewWalletBackendParams (WalletBackendParams{..}))
+    genesisConfig walletConfig txpConfig ntpConfig nodeParams _ nodeRes = do
     logInfo "[Attention] Software is built with the wallet backend"
     ntpStatus <- withNtpClient (ntpClientSettings ntpConfig)
     userSecret <- readTVarIO (ncUserSecret $ nrContext nodeRes)
+
+    let (nodeIp, nodePort) = walletNodeAddress
+    nodeClient <- Plugins.setupNodeClient
+         (BS8.unpack nodeIp, fromIntegral nodePort)
+         walletNodeTlsClientCert
+         walletNodeTlsCaCertPath
+         walletNodeTlsPrivKey
+
     let nodeState = NodeStateAdaptor.newNodeStateAdaptor
             genesisConfig
             nodeRes
             ntpStatus
+            eta
+            nodeClient
+
     liftIO $ Keystore.bracketLegacyKeystore userSecret $ \keystore -> do
         let dbOptions = getWalletDbOptions params
         let dbPath = walletDbPath dbOptions
@@ -69,7 +85,7 @@ actionWithWallet params genesisConfig walletConfig txpConfig ntpConfig nodeParam
         WalletLayer.Kernel.bracketPassiveWallet pm dbMode logMessage' keystore nodeState (npFInjects nodeParams) $ \walletLayer passiveWallet -> do
             migrateLegacyDataLayer passiveWallet dbPath (getFullMigrationFlag params)
 
-            let plugs = plugins (walletLayer, passiveWallet) dbMode
+            let plugs = plugins (walletLayer, passiveWallet) nodeClient dbMode
 
             Kernel.Mode.runWalletMode
                 genesisConfig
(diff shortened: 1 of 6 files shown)


src/Cardano/Wallet/Action.hs
```diff
@@ -78,13 +94,23 @@ actionWithWallet params genesisConfig walletConfig txpConfig ntpConfig nodeParam
...
                 (runNode genesisConfig txpConfig nodeRes plugs)
   where
+
+    -- TODO(@anviking #87): proper error handling
+    eta :: MonadIO m => Show e => ExceptT e m a -> m a
+    eta e = do
+        x <- runExceptT e
+        case x of
+            Right a   -> return a
+            Left  err -> error $ "Error talking to the Node: " <> (show err)
```
Anviking: So this just crashes now…

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
```diff
@@ -374,25 +379,27 @@ newNodeStateAdaptor :: forall m ext. (NodeConstraints, MonadIO m, MonadMask m)
...
+                    -> (forall e a. (Show e, Show a) => ExceptT e IO a -> m a)
+                    -> NodeHttpClient
                     -> NodeStateAdaptor m
-newNodeStateAdaptor genesisConfig nr ntpStatus = Adaptor
+newNodeStateAdaptor genesisConfig nr ntpStatus eta client = Adaptor
     { withNodeState            =            run
-    , getTipSlotId             =            run $ \_lock -> defaultGetTipSlotId genesisHash
-    , getMaxTxSize             =            run $ \_lock -> defaultGetMaxTxSize
-    , getFeePolicy             =            run $ \_lock -> defaultGetFeePolicy
+    , getTipSlotId             = eta $ API.unV1 . API.setSlotId <$> getNodeSettings client
```
Anviking, line 393: Wait, why are we ever asking the node about the current `SlotId` if we know it ourselves in the wallet?

src/Cardano/Wallet/Action.hs
```diff
@@ -48,14 +51,27 @@ actionWithWallet
     -> SscParams
     -> NodeResources EmptyMempoolExt
     -> IO ()
-actionWithWallet params genesisConfig walletConfig txpConfig ntpConfig nodeParams _ nodeRes = do
+actionWithWallet
+    params@(NewWalletBackendParams (WalletBackendParams{..}))
```
KtorZ: Good reminder that we need to get rid of this wrapper now :)

src/Cardano/Wallet/Action.hs
```diff
@@ -48,14 +51,27 @@ actionWithWallet
...
     logInfo "[Attention] Software is built with the wallet backend"
     ntpStatus <- withNtpClient (ntpClientSettings ntpConfig)
     userSecret <- readTVarIO (ncUserSecret $ nrContext nodeRes)
+
+    let (nodeIp, nodePort) = walletNodeAddress
+    nodeClient <- Plugins.setupNodeClient
+         (BS8.unpack nodeIp, fromIntegral nodePort)
+         walletNodeTlsClientCert
+         walletNodeTlsCaCertPath
+         walletNodeTlsPrivKey
```
KtorZ, line 67: ~This is already done as part of:~

https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/Server/Plugins.hs#L96-L100

Ok, this was moved up there.

src/Cardano/Wallet/Action.hs
```diff
@@ -48,14 +51,27 @@ actionWithWallet
...
+
     let nodeState = NodeStateAdaptor.newNodeStateAdaptor
             genesisConfig
```
KtorZ, line 70: Minor styling advice: use qualified import to disambiguate ambiguous methods like `fromList` or `encode` which need contexts to get fully grasped when read. Here, it just makes think needlessly verbose: `newNodeStateAdaptor` is clear in itself :+1:

src/Cardano/Wallet/Action.hs
KtorZ: Yeah... This is not a _valid_ use-case for `error`. Don't use raw exception for things that are actually expected to fail. We may use them seldom for invariants, but not as error handling. 

If there's no particular way to recover from a failure here, logging it would be the very least we could do :) 



Also `eta` doesn't strike me as an acronym for anything :thinking: ?

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
KtorZ: Do we ?

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
```diff
@@ -374,25 +379,27 @@ newNodeStateAdaptor :: forall m ext. (NodeConstraints, MonadIO m, MonadMask m)
...
+    , getMaxTxSize             = eta $ fromIntegral . unMaxTxSize . API.setMaxTxSize <$> getNodeSettings client
```
KtorZ, line 394: I'd rather have this type be a `MaxTxSize` and carried as such.  This would remove the need for comments such as:

```hs
estimateMaxTxInputs
  :: Byte -- ^ Maximum size of a transaction
  -> Word64
estimateMaxTxInputs = estimateMaxTxInputsExplicitBounds
                        boundAddrAttrSize
                        boundTxAttrSize
```

We may unwrap it only when we truly need it!

src/Cardano/Wallet/WalletLayer/Kernel/Transactions.hs
```diff
@@ -71,17 +71,14 @@ getTransactions wallet mbWalletId mbAccountIndex mbAddress params fop sop = lift
...
-            currentSlot <- liftIO $ Node.getTipSlotId (wallet ^. Kernel.walletNode)
             if null metas then
                 -- A bit artificial, but we force the termination and make sure
                 -- in the meantime that the algorithm only exits by one and only
                 -- one branch.
                 go cp (min pp $ length acc) (acc, total <|> mbTotalEntries)
             else do
                 txs <- catMaybes <$> forM metas (\meta -> do
-                    runExceptT (metaToTx db sc currentSlot meta) >>= \case
+                    toTransaction wallet meta >>= \case
```
KtorZ, line 81: Any plan to update `toTransaction` in the future :no_mouth: ?

KtorZ (changes requested):

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
Anviking: The only use-case:

https://github.com/input-output-hk/cardano-wallet/blob/8bbb1630bc903c230dc591cee182964e5119191e/src/Cardano/Wallet/WalletLayer/Kernel/Transactions.hs#L100

src/Cardano/Wallet/WalletLayer/Kernel/Transactions.hs
Anviking: What do you mean?

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
Anviking: Hm. Ok. Doesn't this look inconsistent though?
```haskell
estimateMaxTxInputsExplicitBounds
  :: Byte -- ^ Size of @Attributes AddrAttributes@
  -> Byte -- ^ Size of @Attributes ()@
  -> MaxTxSize
```

Furthermore, `MaxTxSize` is uncomfortable to unwrap now since it is really the "intermediate API type" from the node API: 

```haskell
newtype MaxTxSize = MaxTxSize (MeasuredIn 'Bytes Word)
```

What about creating new newtypes in either:
1. `NodeStateAdaptor`
2. `API.V1.Types`
3. `Pos.Node.API` 

, like this:
```haskell
newtype MaxTxSize = MaxTxSize Byte
```
and convert between the two?

(For now irrelevant side note: don't we get into DerivingVia territory then? If we only use MeasuredIn for JSON instances…)

src/Cardano/Wallet/WalletLayer/Kernel/Transactions.hs
KtorZ: Nevermind :upside_down_face:

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
KtorZ: > What about creating new newtypes

Sounds reasonable. But let's not bother too much for now :(  Perhaps a ticket with low-priority.

> don't we get into DerivingVia territory then

We could, but there's no problem that DerivingVia help us to solve here that generics can't already solve for us. Plus, we don't have DerivingVia in 8.4 :sob:

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
KtorZ: What do you mean by "we know it ourselves" ?

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
Anviking: >  But let's not bother too much for now

:+1::+1::+1:

Keeping the types in `NodeStateAdaptor` unchanged then, and creating a ticket for the future.

src/Cardano/Wallet/Action.hs
Anviking: Done. We get more cluttered imports though :(

KtorZ (approved):
LGTM.
Will merge after #237 (this also bumps cardano-sl, to latest develop).

Anviking:
> KtorZ approved these changes 

@KtorZ Though, FYI, when artificially failing in `toMonadIO` I was confused by seeing the following in the test output:

```
       user error (expected a successful response but got an error: ClientHttpError (ConnectionError "HttpExceptionRequest Request {\n  host                 = \"127.0.0.1\"\n  port                 = 8090\n  secure               = True\n  requestHeaders       = []\n  path                 = \"/api/internal/reset-wallet-state\"\n  queryString          = \"\"\n  method               = \"DELETE\"\n  proxy                = Nothing\n  rawBody              = False\n  redirectCount        = 10\n  responseTimeout      = ResponseTimeoutDefault\n  requestVersion       = HTTP/1.1\n}\n (InternalException (HostCannotConnect \"127.0.0.1\" [Network.Socket.connect: <socket: 101>: does not exist (Connection refused)]))"))
```

Not sure if this is as expected or not.

(In the log output I saw the `NodeCommunicationError` as expected though)

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
```diff
@@ -687,3 +687,41 @@ instance Buildable MissingBlock where
 instance Buildable UnknownEpoch where
   build (UnknownEpoch slotId) =
     bprint ("UnknownEpoch " % build) slotId
+
+
+{-------------------------------------------------------------------------------
+  Errors
+-------------------------------------------------------------------------------}
+
+newtype NodeCommunicationError = NodeCommunicationError String
```
Anviking, line 696: Was a bit unsure about if we should have done `NodeCommunication e` instead and done buildable stuff

PR #219, opened 2019-01-11
Only use a faucet when one is needed

KtorZ:
<p align="right">#172</p>

# Overview

- [x] I have changed the `setupWallet` internal function from the DSL to only use get and use a faucet wallet when there's a need for one (i.e. when the fixture requires some initial coins).

# Comments

During integration test scenarios, a faucet wallet is needed only when
we require a wallet to have some initial coins. Quite a few scenarios
don't require this and therefore, we could save some faucet wallets (and
make the overall a bit faster) by just not getting one from the pool
when there's no need

This just stroke me as I was running the tests. We have a lot of faucet wallets but since we are extending the test suite quite a lot, we may hit that boundary sooner than we think.

### DIFF
--- a/test/integration/Test/Integration/Framework/DSL.hs
+++ b/test/integration/Test/Integration/Framework/DSL.hs
@@ -449,7 +449,7 @@ expectTxStatusEventually
 expectTxStatusEventually statuses = \case
     Left e    -> wantedSuccessButError e
     Right txn -> do
-        result <- ask >>= \ctx -> timeout (60 * second) (waitForTxStatus ctx statuses txn)
+        result <- ask >>= \ctx -> timeout (120 * second) (waitForTxStatus ctx statuses txn)
         case result of
             Nothing -> fail "expectTxStatusEventually: waited too long for statuses."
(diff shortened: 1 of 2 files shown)


piotr-iohk (approved):
Ok for the change!
 
BTW. I see that CI for this fails on one of the tests with the same issue I have observed:

```
  1) Transactions coin selection: needs many extra inputs
       user error (expectTxStatusEventually: waited too long for statuses.)
```

KtorZ:
Ah!
That is very interesting! I am not going to merge without having investigated this then. That is not normal.

piotr-iohk:
Actually, I was about to create a bug for that putting all my observations there. Locally I observed that the issue is more likely to occur when I create few wallets with some coins in the scenario setup (e.g. with 3 it fails every once in a while, with 4 wallets it fails for me almost every time).

piotr-iohk:
#220

KtorZ:
Will investigate the failure as part of #220. Merging this meanwhile.

PR #223, opened 2019-01-14
WIP Modify fixture in newPayment for new address scheme

akegalj:
This PR modifies fixtures in NewPayment spec so
that fixture now contains either old address scheme
or new address scheme.

<p align="right">#45</p>

# Overview

- [x] I have added new type `AddressSchemeVersion` to distinguish between old scheme and new scheme
    - https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-30d40db9a639deefc7b1db60e7053813R600
- [x] I have partially extended how we can generate keys in cardano-sl crypto
   - https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-fafd0cdcd559a7b124cc61c29413fb54R7
   - https://github.com/input-output-hk/cardano-sl/commit/4ecd69bebd98c27d1d20d68c64293eb303a6cf8c
   - https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-849da1a7f8deb2932cb50ae0cbaa4d35R133
- [x] I have added passphrase to fixtures as it is required for new derivation scheme
   - https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-849da1a7f8deb2932cb50ae0cbaa4d35R83
 - [x] I have modified `prepareFixtures` so that it can mix old and new derivation/address scheme
   - https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-849da1a7f8deb2932cb50ae0cbaa4d35R93

# Comments

Tests are passing currently passing if we choose to test old address scheme path only https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-849da1a7f8deb2932cb50ae0cbaa4d35R94 . When new derivation scheme is choosen tests still fail and I have difficulties tracking this failure down. 

When `Kernel.AddressSchemeV1` is sellected, we are going to generate root/master key with https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-849da1a7f8deb2932cb50ae0cbaa4d35R101 and we are going to generate addresses for our utxo with https://github.com/input-output-hk/cardano-wallet/pull/223/files#diff-849da1a7f8deb2932cb50ae0cbaa4d35R125 . 

With these changes to fixture, and running with `AddressSchemeV1` hardcoded tests are going to fail with `UnknownHdAccount`. When I tried to modify:

```haskell
        let accounts    = Kernel.prefilterUtxo newRootId esk utxo'
            hdAccountId = Kernel.defaultHdAccountId newRootId
            hdAddress   = Kernel.defaultHdAddress nm esk passPhrase newRootId
```

into 
```haskell
        let accounts    = Kernel.prefilterUtxo newRootId esk utxo'
            hdAccountId = newAccountId
            hdAddress   = Kernel.defaultHdAddress nm esk passPhrase newRootId
```
Which looks good from my perspective (am I right?) I will be start getting failure `UnknownHdAccountRoot`. If I try with:

```haskell
        let accounts    = Kernel.prefilterUtxo newRootId esk utxo'
            hdAccountId = newAccountId
            hdAddress   = Nothing
```
it will fail with ` CoinSelHardErrUtxoExhausted`. Long version is:

```
 NewPayment.EstimateFees, Estimating fees (wallet layer), estimating fees works (SenderPaysFee)
       user error (EstimateFeesError EstFeesTxCreationFailed NewTransactionErrorCoinSelectionFailed CoinSelHardErrUtxoExhausted{ balance: 0 coin(s), value:   10 coin(s)})
```

And this is where I am going deeper into the problem but I am not sure should I - if not all bits are in place. Maybe its expected everything will still fail with this fixture update if other bits of bip44 feature is not complete. This way I am going through big chunks of code to get understanding and maybe that's wasting time. If this is a good dirrection - I am going to proceed with this method.

### DIFF
--- a/src/Cardano/Wallet/Kernel/Transactions.hs
+++ b/src/Cardano/Wallet/Kernel/Transactions.hs
@@ -10,6 +10,8 @@ module Cardano.Wallet.Kernel.Transactions (
     , PaymentError(..)
     , EstimateFeesError(..)
     , RedeemAdaError(..)
+    , AddressSchemeVersion (..)
+    , addressSchemeVersion
     , cardanoFee
     , mkStdTx
     , prepareUnsignedTxWithSources
@@ -34,7 +36,7 @@ import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map.Strict as Map
 import qualified Data.Vector as V
 import qualified System.Random.MWC (GenIO, asGenIO, initialize, uniformVector)
-import           Test.QuickCheck (Arbitrary (..), oneof)
+import           Test.QuickCheck (Arbitrary (..), elements, oneof)
 
 import           Formatting (bprint, build, sformat, (%))
 import qualified Formatting.Buildable
@@ -594,6 +596,27 @@ instance Arbitrary SignTransactionError where
         , SignTransactionErrorNotOwned <$> arbitrary
         ]
 
+-- Which address scheme is being used
+data AddressSchemeVersion
+    -- Address scheme 0 (initial address scheme) is bip32 HD wallet derived
+    -- address. It is derived with ed25519v0 curve and contains address payload: account index and address index
+    = AddressSchemeV0
+    -- Address scheme 1 is bip44 HD wallet derived
+    -- address. It is derived with ed25519v1 curve and contains no address payload
+    | AddressSchemeV1
+    deriving (Show)
+
+instance Arbitrary AddressSchemeVersion where
+    arbitrary = elements [AddressSchemeV0, AddressSchemeV1]
+
+addressSchemeVersion :: Address -> AddressSchemeVersion
+addressSchemeVersion addr =
+    if isJust mPayload
+        then AddressSchemeV0
+        else AddressSchemeV1
+  where
+    mPayload = Core.aaPkDerivationPath $ attrData $ Core.addrAttributes addr
+
 mkSigner :: NetworkMagic
          -> PassPhrase
          -> Maybe EncryptedSecretKey
@@ -612,16 +635,13 @@ mkSigner nm spendingPassword (Just esk) snapshot addr =
                                        . HD.hdAddressIdParent
                                        . HD.hdAccountIdIx
                                        . to HD.getHdAccountIx
-                mAddressPayload = hdAddr ^. HD.hdAddressAddress
+                internalAddress = hdAddr ^. HD.hdAddressAddress
                                           . fromDb
-                                          . to Core.addrAttributes
-                                          . to attrData
-                                          . to Core.aaPkDerivationPath
-                res = case mAddressPayload of
+                res = case addressSchemeVersion internalAddress of
                     -- If there is some payload we expect this payload to be addressIx and accountIx
                     -- used for old address scheme so we continue with using
                     -- old HD address derivation scheme: simple bip32 with ed25519 v0
-                    Just _ -> Core.deriveLvl2KeyPair nm
+                    AddressSchemeV0 -> Core.deriveLvl2KeyPair nm
                                     (Core.IsBootstrapEraAddr True)
                                     (ShouldCheckPassphrase False)
                                     spendingPassword
@@ -630,7 +650,7 @@ mkSigner nm spendingPassword (Just esk) snapshot addr =
                                     addressIndex
                     -- If there is no payload we assume it is a new address scheme (which doesn't have payload)
                     -- New HD address derivation scheme: bip44 with ed25519 v1
-                    Nothing -> first (Core.makePubKeyAddressBoot nm) <$>
+                    AddressSchemeV1 -> first (Core.makePubKeyAddressBoot nm) <$>
                         deriveAddressKeyPair
                             spendingPassword
                             esk
--- a/src/Cardano/Wallet/Kernel/Transactions.hs
+++ b/src/Cardano/Wallet/Kernel/Transactions.hs
@@ -598,22 +598,22 @@ instance Arbitrary SignTransactionError where
 
 -- Which address scheme is being used
 data AddressSchemeVersion
-    -- Address scheme 0 (initial address scheme) is bip32 HD wallet derived
+    -- Address scheme random (initial address scheme) is bip32 HD wallet derived
     -- address. It is derived with ed25519v0 curve and contains address payload: account index and address index
-    = AddressSchemeV0
-    -- Address scheme 1 is bip44 HD wallet derived
+    = AddressSchemeRnd
+    -- Address scheme sequential is bip44 HD wallet derived
     -- address. It is derived with ed25519v1 curve and contains no address payload
-    | AddressSchemeV1
+    | AddressSchemeSeq
     deriving (Show)
 
 instance Arbitrary AddressSchemeVersion where
-    arbitrary = elements [AddressSchemeV0, AddressSchemeV1]
+    arbitrary = elements [AddressSchemeRnd, AddressSchemeSeq]
 
 addressSchemeVersion :: Address -> AddressSchemeVersion
 addressSchemeVersion addr =
     if isJust mPayload
-        then AddressSchemeV0
-        else AddressSchemeV1
+        then AddressSchemeRnd
+        else AddressSchemeSeq
   where
     mPayload = Core.aaPkDerivationPath $ attrData $ Core.addrAttributes addr
 
@@ -639,18 +639,18 @@ mkSigner nm spendingPassword (Just esk) snapshot addr =
                                           . fromDb
                 res = case addressSchemeVersion internalAddress of
                     -- If there is some payload we expect this payload to be addressIx and accountIx
-                    -- used for old address scheme so we continue with using
-                    -- old HD address derivation scheme: simple bip32 with ed25519 v0
-                    AddressSchemeV0 -> Core.deriveLvl2KeyPair nm
+                    -- used for random address scheme so we continue with using
+                    -- random HD address derivation scheme: simple bip32 with ed25519 v0
+                    AddressSchemeRnd -> Core.deriveLvl2KeyPair nm
                                     (Core.IsBootstrapEraAddr True)
                                     (ShouldCheckPassphrase False)
                                     spendingPassword
                                     esk
                                     accountIndex
                                     addressIndex
-                    -- If there is no payload we assume it is a new address scheme (which doesn't have payload)
-                    -- New HD address derivation scheme: bip44 with ed25519 v1
-                    AddressSchemeV1 -> first (Core.makePubKeyAddressBoot nm) <$>
+                    -- If there is no payload we assume it is sequential address scheme (which doesn't have payload)
+                    -- Sequential HD address derivation scheme: bip44 with ed25519 v1
+                    AddressSchemeSeq -> first (Core.makePubKeyAddressBoot nm) <$>
                         deriveAddressKeyPair
                             spendingPassword
                             esk
--- a/stack.yaml
+++ b/stack.yaml
@@ -4,7 +4,7 @@ packages:
 
 extra-deps:
 - git: https://github.com/input-output-hk/cardano-sl
-  commit: cdc77284bde949d4b1b9ac7be81b76b906a71fc7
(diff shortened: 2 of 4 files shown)


KtorZ:
@akegalj Any hope with that one :) ?

src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -594,6 +596,27 @@ instance Arbitrary SignTransactionError where
         , SignTransactionErrorNotOwned <$> arbitrary
         ]
 
+-- Which address scheme is being used
+data AddressSchemeVersion
```
akegalj, line 600: special type created to help distinguish between old and new derivation scheme

stack.yaml
```diff
@@ -4,7 +4,7 @@ packages:
 
 extra-deps:
 - git: https://github.com/input-output-hk/cardano-sl
-  commit: cdc77284bde949d4b1b9ac7be81b76b906a71fc7
+  commit: 4ecd69bebd98c27d1d20d68c64293eb303a6cf8c
```
akegalj, line 7: partially extended key generation with new derivation scheme https://github.com/input-output-hk/cardano-sl/commit/4ecd69bebd98c27d1d20d68c64293eb303a6cf8c

test/unit/Test/Spec/Ed25519Bip44.hs
```diff
@@ -23,6 +23,10 @@ import           Test.QuickCheck (Arbitrary (..), InfiniteList (..), Property,
                      arbitraryBoundedIntegral, arbitrarySizedBoundedIntegral,
                      property, shrinkIntegral, (.&&.), (===), (==>))
 
+-- TODO (akegalj): seems like there already is a notion of
```
akegalj, line 26: there is a development point about this

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -72,11 +75,12 @@ import           Servant.Server
...
-    , fixtureESK       :: EncryptedSecretKey
-    , fixtureAccountId :: HdAccountId
-    , fixturePw        :: PassiveWallet
-    , fixturePayees    :: NonEmpty (Address, Coin)
+      fixtureHdRootId   :: HdRootId
+    , fixtureESK        :: EncryptedSecretKey
+    , fixtureAccountId  :: HdAccountId
+    , fixturePw         :: PassiveWallet
+    , fixturePayees     :: NonEmpty (Address, Coin)
+    , fixturePassPhrase :: PassPhrase
```
akegalj, line 82: passphrase is required in ed25519bip44

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
                 -> Pay
                 -> Fixture.GenActiveWalletFixture Fixture
 prepareFixtures nm initialBalance toPay = do
-    let (_, esk) = safeDeterministicKeyGen (B.pack $ replicate 32 0x42) mempty
+    addressVersion <- return Kernel.AddressSchemeV1 -- pick arbitrary
```
akegalj: hardcoded in order to test V0 branch first and then V1

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
+    passPhrase <- pick $ arbitrary @PassPhrase
```
akegalj, line 94: passphrase is required for ed25519bip44

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
+    mnemonic <- pick $ arbitrary @(Mnemonic 12)
+    let seed = mnemonicToSeed mnemonic
+    let esk =
+            case addressVersion of
```
akegalj, line 98: root key depends on derivation scheme version

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
-                          <*> pure (NoSpendingPassword $ InDb now)
+                          <*> pure (HasSpendingPassword $ InDb now)
                           <*> pure AssuranceLevelNormal
                           <*> (InDb <$> pick arbitrary)
     newAccountId <- HdAccountId newRootId <$> deriveIndex (pick . choose) HdAccountIx HardDerivation
     utxo   <- pick (genUtxoWithAtLeast initialBalance)
     -- Override all the addresses of the random Utxo with something meaningful,
     -- i.e. with 'Address'(es) generated in a principled way, and not random.
     utxo' <- foldlM (\acc (txIn, (TxOutAux (TxOut _ coin))) -> do
-                        newIndex <- deriveIndex (pick . choose) HdAddressIx HardDerivation
```
akegalj, line 103: I am not sure why address derivation was `HardDerivation` before?

V0 works fine with `SoftDerivation` and I think it should be `SoftDerivation` in the first place

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
-
-                        let Just (addr, _) = deriveLvl2KeyPair nm
-                                                               (IsBootstrapEraAddr True)
-                                                               (ShouldCheckPassphrase True)
-                                                               mempty
-                                                               esk
-                                                               (newAccountId ^. hdAccountIdIx . to getHdAccountIx)
-                                                               (getHdAddressIx newIndex)
+                        newIndex <- deriveIndex (pick . choose) HdAddressIx SoftDerivation
```
akegalj: ed25519bip44 requires address derivation to be soft ferivation (non-hardened)

ps - the discussion about wording "hardened/non-hardened index" "hard/soft index" ?

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
+
+                        let Just (addr, _) =
+                                case addressVersion of
```
akegalj, line 119: depending on derivation scheme picked - we are going to use old derivation method or ed25519bip44 (new one)

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
+                                    Kernel.AddressSchemeV1 -> first (makePubKeyAddressBoot nm) <$> deriveAddressKeyPair
+                                                            passPhrase
+                                                            esk
+                                                            (newAccountId ^. hdAccountIdIx . to getHdAccountIx)
+                                                            ExternalChain
+                                                            (getHdAddressIx newIndex)
                         return $ M.insert txIn (TxOutAux (TxOut addr coin)) acc
                     ) M.empty (M.toList utxo)
-    payees <- fmap (\(TxOut addr coin) -> (addr, coin)) <$> pick (genPayeeWithNM nm utxo toPay)
+    payees <- fmap (\(TxOut addr coin) -> (addr, coin)) <$> pick (genPayeeWithNM nm utxo' toPay) -- FIX this in the morning - generate keys for new addresses
```
akegalj: comment will be removed.

`genPayeeWithNM` will now generate payees (outputs) with both old and new address https://github.com/input-output-hk/cardano-sl/commit/4ecd69bebd98c27d1d20d68c64293eb303a6cf8c#diff-87ba451be2edb49d60fa08f373618434R104

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
 
     return $ \keystore aw -> do
         liftIO $ Keystore.insert newRootId esk keystore
         let pw = Kernel.walletPassive aw
 
         let accounts    = Kernel.prefilterUtxo newRootId esk utxo'
-            hdAccountId = Kernel.defaultHdAccountId newRootId
-            hdAddress   = Kernel.defaultHdAddress nm esk emptyPassphrase newRootId
+            hdAccountId = newAccountId -- Kernel.defaultHdAccountId newRootId
```
akegalj, line 142: I am not sure why default account has been created and inserted into db when the right thing to insert is new account created in this fixture (with all the addresses from above)?

This is really confusing to me

Also see comments from Comments section in description https://github.com/input-output-hk/cardano-wallet/pull/223

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +91,54 @@ prepareFixtures :: NetworkMagic
...
+            hdAddress   = Nothing -- Kernel.defaultHdAddress nm esk passPhrase newRootId
```
akegalj, line 143: I am also not clear what is the goal of default hd address - this bit is really confusing to me

akegalj:
ps - maybe it makes sense to keep this PR hanging a few more days - until we merge more bip44 PRs and `develop` branch will be more feature complete (regarding bip44).

To prioritze feature completion over test completition (althought both are important)

stack.yaml
KtorZ: Hmmm. Please don't. We want to ditch `cardano-sl` in the long run / near future. Hence why we implemented the logic in our repository.

src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -594,6 +596,27 @@ instance Arbitrary SignTransactionError where
...
+    -- Address scheme 0 (initial address scheme) is bip32 HD wallet derived
+    -- address. It is derived with ed25519v0 curve and contains address payload: account index and address index
+    = AddressSchemeV0
+    -- Address scheme 1 is bip44 HD wallet derived
+    -- address. It is derived with ed25519v1 curve and contains no address payload
+    | AddressSchemeV1
```
KtorZ: I'd however desambiguate the names here and go for `AddressSchemeRnd` and `AddressSchemeSeq`.

stack.yaml
akegalj: Ok - I will start moving the logic to cardano-wallet within this PR

test/unit/Test/Spec/NewPayment.hs
KtorZ: That is a bit wrong here. Because the point of the new derivation is to be sequential, not random. So, it's likely a bit more complex than that. We need to generate addresses within a certain gap if we want the fixture to be meaningful. Otherwise, we won't be able to discover those addresses.

test/unit/Test/Spec/NewPayment.hs
KtorZ: To the best of my knowledge, the default address is there to allow users to perform transactions while the wallet is restoring (and therefore, hasn't discovered its addresses yet).
I am not sure how this is relevant in the context of unit testing...

test/unit/Test/Spec/NewPayment.hs
akegalj: Yes - that's true.

I thought that generating in random won't harm for tests - but in fact it might - if rest of the code expects it to be sequential.

src/Cardano/Wallet/Kernel/Transactions.hs
akegalj: Done in https://github.com/input-output-hk/cardano-wallet/pull/223/commits/47f521beb310e23ef8a4739e12503d9a33a7189f

test/unit/Test/Spec/NewPayment.hs
akegalj: This commit https://github.com/input-output-hk/cardano-wallet/pull/223/commits/cf95ddb11a234c3f54e403d70c0f215aae23b9e5 should address this problem: creates simple sequential account and corresponding sequential addresses

akegalj: Its a simplest use case and can be improved upon though

KtorZ:
Closing this one until we are ready for this to get in.

PR #225, opened 2019-01-15
HdRootId as an opaque type

denisshevchenko:
<p align="right">#231</p>

# Overview

- [x] `HdRootId` is an opaque type with `Text` inside.

# Comments

`HdRootId` now is a unified identifier for wallets (both FO and EO ones).
Technically it contains a text, but actually it contains one from two options:

1. for FO-wallets - `Core.Address` made from wallet's root public key, in Base58-format.
2. for EO-wallets - `UUID`, in Base58-format.

Please note that using `Core.Address` as wallet's identifier isn't a good decision, but since exchanges and users may have wallets' ids stored on their side we have to keep backward-compatibility for existing ids.

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/HdRootId.hs
+++ b/src/Cardano/Wallet/Kernel/DB/HdRootId.hs
@@ -4,96 +4,100 @@
 module Cardano.Wallet.Kernel.DB.HdRootId (
       HdRootId
     , decodeHdRootId
-    , mkHdRootIdForEOWallet
-    , mkHdRootIdForFOWallet
+    , genHdRootId
+    , eskToHdRootId
     ) where
 
 import           Universum
 
+import qualified Data.ByteString as BS
 import           Data.ByteString.Base58 (bitcoinAlphabet, decodeBase58,
                      encodeBase58)
 import           Data.SafeCopy (base, deriveSafeCopy)
-import qualified Data.UUID as Uuid
-import           Data.UUID.V4 (nextRandom)
-
-import           Test.QuickCheck (Arbitrary (..))
-import           Test.QuickCheck.Gen (oneof)
-
+import qualified Data.UUID as UUID
+import qualified Data.UUID.V4 as UUID
 import           Formatting (bprint, build)
 import qualified Formatting.Buildable
+import           Test.QuickCheck (Arbitrary (..))
+import           Test.QuickCheck.Gen (chooseAny, oneof, vectorOf)
 
 import qualified Pos.Binary.Class as Bi
-import qualified Pos.Core as Core
+import           Pos.Core (Address, addrToBase58, makePubKeyAddressBoot)
 import           Pos.Core.NetworkMagic (NetworkMagic (..))
-import           Pos.Crypto (EncryptedSecretKey, encToPublic)
+import           Pos.Crypto (EncryptedSecretKey, encToPublic,
+                     safeDeterministicKeyGen)
+import           Test.Pos.Core.Arbitrary ()
+
 
 -- | Unified identifier for wallets (both FO and EO ones).
 -- Technically it contains a text, but actually it contains one from two options:
 --
--- 1. for FO-wallets - 'Core.Address' made from wallet's root public key, in Base58-format.
+-- 1. for FO-wallets - 'Address' made from wallet's root public key, in Base58-format.
 -- 2. for EO-wallets - 'UUID', in Base58-format.
 --
--- Please note that using 'Core.Address' as wallet's identifier is a bad decision,
--- but since exchanges and users may have wallets' ids stored on their side we
--- have to keep backward-compatibility for existing ids.
+-- Using 'Address' for HdRootId was an arbitrary decision which turns out to
+-- yield confusion and interrogations. There's no particular relationship
+-- between an HdRootId and an actual Address. At that time, we needed an ID that
+-- can be unique and determinically computed from a root encrypted secret key.
+-- Addresses have that property, and therefore, were picked as root ids.
+--
+-- Since users of the API may have wallets' ids stored on their side, we can't
+-- really change that for existing wallets in order to keep backward
+-- compatibility.
 newtype HdRootId = HdRootId { getHdRootId :: Text }
     deriving (Eq, Ord, Show, Generic)
+deriveSafeCopy 1 'base ''HdRootId
 
 instance NFData HdRootId
 
--- | TODO: It is not very arbitrary, maybe we should improve it.
 instance Arbitrary HdRootId where
-    arbitrary = oneof
-        [ pure $ HdRootId "Ae2tdPwUPEZ18ZjTLnLVr9CEvUEUX4eW1LBHbxxxJgxdAYHrDeSCSbCxrvx"
-        , pure $ HdRootId "J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg"
-        , pure $ HdRootId "kKyjp8hUCYDK1UhU8Wg5sBu3Ud2TqFDCbTWCo9ySDGjaLLk73"
-        , pure $ HdRootId "QxUZbMEK9Vg6V9UWQ1BRLwrBoQQuYt2RjnEW4vRmXBtsT9WKk"
-        , pure $ HdRootId "maKVJgoL6c65NfUJDGeyQhcK6b9rynpEdwKe4T1wn3jfEZZRv"
-        , pure $ HdRootId "SC6jKYB9gFnNhvZNuQW6e7MnWM59pwcyNp1ayp9n9iEiucYWr"
-        , pure $ HdRootId "RqTmjdzemFRaZzu4TZH9qwS5NefKgTZ6fp2bELoX8hwJPSuBw"
-        ]
+    arbitrary = oneof [ arbitraryFO , arbitraryEO ]
+      where
+        arbitraryFO = do
+            (_, esk) <- safeDeterministicKeyGen
+                <$> (BS.pack <$> vectorOf 32 arbitrary)
+                <*> pure mempty
+            nm <- arbitrary
+            pure (eskToHdRootId nm esk)
+
+        arbitraryEO =
+            HdRootId
+            . decodeUtf8
+            . encodeBase58 bitcoinAlphabet
+            . UUID.toASCIIBytes
+            <$> chooseAny
 
 instance Buildable HdRootId where
     build (HdRootId uniqueId) = bprint build uniqueId
 
-deriveSafeCopy 1 'base ''HdRootId
 
--- | Decodes 'HdRootId' from the text.
-decodeHdRootId :: Text -> Either Text HdRootId
-decodeHdRootId rawId = decodeFromBS . encodeUtf8 $ rawId
+decodeHdRootId :: Text -> Maybe HdRootId
+decodeHdRootId txt = do
+    bs <- decodeBase58 bitcoinAlphabet (encodeUtf8 txt)
+    decodeAddr bs <|> decodeUUID bs
   where
-    decodeFromBS :: ByteString -> Either Text HdRootId
-    decodeFromBS bs = do
-        let base58Err = "Invalid Base58 representation of HdRootId"
-        decodedBS <- maybeToRight base58Err $ decodeBase58 bitcoinAlphabet bs
-        let probablyUUID = decodeUUID decodedBS
-            probablyAddress = decodeAddress decodedBS
-        if isRight probablyUUID
-            then probablyUUID
-            else probablyAddress
+    decodeAddr :: ByteString -> Maybe HdRootId
+    decodeAddr bs = do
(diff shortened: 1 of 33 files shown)


denisshevchenko:
@KtorZ Yes, I'm fixing it.

denisshevchenko:
All tests passed, checks have failed because of Nix.

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,83 @@
...
+
+-- | Unified identifier for wallets (both FO and EO ones).
+-- Technically it contains a text, but actually it contains one from two options:
+--
+-- 1. for FO-wallets - 'Core.Address' made from wallet's root public key, in Base58-format.
+-- 2. for EO-wallets - 'UUID', in Base58-format.
+--
+-- Please note that using 'Core.Address' as wallet's identifier is a bad decision,
+-- but since exchanges and users may have wallets' ids stored on their side we
+-- have to keep backward-compatibility for existing ids.
```
KtorZ: Sweet. 
Minor thing: when saying something is _bad_, it's better to explain _why_ because it may not be obvious to the reader.

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,83 @@
...
+newtype HdRootId = HdRootId { getHdRootId :: Text }
+    deriving (Eq, Ord, Show, Generic)
+
+instance NFData HdRootId
+
+instance Arbitrary HdRootId where
+    arbitrary = pure $
+        HdRootId "Ae2tdPwUPEZ18ZjTLnLVr9CEvUEUX4eW1LBHbxxxJgxdAYHrDeSCSbCxrvx"
```
KtorZ: This doesn't look very `Arbitrary` to me ;)

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,83 @@
...
+
+instance Buildable HdRootId where
+    build (HdRootId uniqueId) = bprint build uniqueId
+
+deriveSafeCopy 1 'base ''HdRootId
+
+-- | Specialized formatter for 'HdRootId'.
+hdRootIdF :: Format r (HdRootId -> r)
+hdRootIdF = build
```
KtorZ: We don't need this. This is just `build`, so let's just rely on `build`. That's why type-classes are for in the end, they give you a common-interface between many types. Here, that's the `Buildable` interface.

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,83 @@
...
+    decodeFromBS :: ByteString -> Either Text HdRootId
+    decodeFromBS bs = do
+        let base58Err = "Invalid Base58 representation of HdRootId"
+        decodedBS <- maybeToRight base58Err $ decodeBase58 bitcoinAlphabet bs
+        -- Check what text contains: UUID or Core.Address.
+        case Uuid.fromASCIIBytes decodedBS of
+            Just _ -> Right $ HdRootId rawId
+            Nothing -> case Bi.decodeFull' decodedBS of
+                Left _ -> Left "Invalid HdRootId (it's not an Address nor UUID)"
+                Right (_addr :: Core.Address) -> Right $ HdRootId rawId
```
KtorZ: Can't we leverage the `Alternative` abstraction here to make the code clearer in what it does :thinking: ? There's no need for a monadic approach here. We try one decoder or the other, in an alternative way.

`decodeHdRootId = decodeAddress <|> decodeUUID`

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,83 @@
...
+
+-- | Generates a random 'HdRootId' (based on UUID) for new EO-wallets.
+-- Because of unification we decode result to Base58, as a result we
+-- have Base58-encoded text for all identifiers.
+genRandomHdRootId :: IO HdRootId
+genRandomHdRootId = HdRootId
+    . decodeUtf8
+    . encodeBase58 bitcoinAlphabet
+    . Uuid.toASCIIBytes
+    <$> nextRandom
```
KtorZ: We'll want to have that seeded (with an `EncryptedSecretKey`) which will also remove the need for `IO` here. Ideally, we want to parameterize it over the wallet's type (which isn't something we have now, but any `data WalletType = ExternallyOwned | FullyOwned` can do for now).

In one case, one wants to derive a "random" but deterministic UUID from it, in the other, derive an address from it. 
Therefore, I'll suggest to replace both this and `addressToHdRootId` with something more opaque: 

```hs
mkRandomHdRootId :: WalletType -> EncryptedSecretKey -> HdRootId
```

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -254,51 +252,15 @@ decryptHdLvl2DerivationPath hdPass addr = do
...
-newtype HdRootId = HdRootId { getHdRootId :: InDb Core.Address }
-  deriving (Eq, Ord, Show, Generic)
-
-instance NFData HdRootId
-
-
-instance Arbitrary HdRootId where
-  arbitrary = do
-      (_, esk) <- Core.safeDeterministicKeyGen <$> (BS.pack <$> vectorOf 32 arbitrary)
-                                               <*> pure mempty
```
KtorZ, line 298: This looked definitely more `Arbitrary` to me :p

src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -112,7 +114,7 @@ data TxMetaT f = TxMeta {
     , _txMetaTableCreatedAt  :: Beam.Columnar f Core.Timestamp
     , _txMetaTableIsLocal    :: Beam.Columnar f Bool
     , _txMetaTableIsOutgoing :: Beam.Columnar f Bool
-    , _txMetaTableWalletId   :: Beam.Columnar f Core.Address
+    , _txMetaTableWalletId   :: Beam.Columnar f HdRootId
```
KtorZ, line 116: How does types get serialized to SQLite :thinking: ? It seems that anything that is `Typeable` can be serialized to the DB... Later, we do have those `FromField` instances that actually don't come from `beam` but come from another package `sqlite-simple`. So we use two diffferent sqlite driver o.O, that's fantastic. 

Anyway. If it works, it works; I'll comment on our `beam` rework ticket about it..

KtorZ (changes requested):
Some changes in the `HdRootId` definitions thinking about what's coming next. 
Rest looks good :+1:

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
denisshevchenko: Ok, two questions:

1. Will we receive `EncryptedSecretKey` for EOS-wallets?
2. If so - how can we convert `EncryptedSecretKey` to `Int`? This is for "deterministically random" value of `UUID`: we need `Int`-value for `mkStdGen`. https://github.com/input-output-hk/cardano-wallet/blob/1b18e6181b7371623ce740c07cbb2b97833b779d/src/Cardano/Wallet/Kernel/DB/HdRootId.hs#L108-L110

KtorZ: Pffffiooou. I guess my mind wasn't really clear yesterday :no_mouth: ... We actually won't need to have deterministic generation of UUID for E.O. wallets because we won't need to peek into the keystore for them ANYWAY. 

Incidentally, obviously, we won't have any secret key for E.O. wallets (by definition). Discard my comment please and sorry about that.

denisshevchenko:
@KtorZ Fixed.

KtorZ (approved):
Wrapped up the remaining bits.

denisshevchenko:
@KtorZ You are awesome!

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,99 @@
...
+--
+-- Please note that using 'Core.Address' as wallet's identifier is a bad decision,
+-- but since exchanges and users may have wallets' ids stored on their side we
+-- have to keep backward-compatibility for existing ids.
+newtype HdRootId = HdRootId { getHdRootId :: Text }
+    deriving (Eq, Ord, Show, Generic)
+
+instance NFData HdRootId
+
+-- | TODO: It is not very arbitrary, maybe we should improve it.
```
akegalj: Lately I add all my TODO's as a development point to parrent issue like:
```
- [ ] Imrove Arbitrary instance of HdRootId
```

Later if I figure I am too lazy to work on that development point or find it less urgent I update the wording to: 
```
- [ ] Open an issue to improve Arbitrary instance of HdRootId
```

and I open a new ticket, mark the above development point as done.

That way its very visible how you can improve the parrent issue and usually I use this in my retrospective to list how we can improve things.

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,99 @@
...
+decodeHdRootId :: Text -> Either Text HdRootId
+decodeHdRootId rawId = decodeFromBS . encodeUtf8 $ rawId
+  where
+    decodeFromBS :: ByteString -> Either Text HdRootId
+    decodeFromBS bs = do
+        let base58Err = "Invalid Base58 representation of HdRootId"
+        decodedBS <- maybeToRight base58Err $ decodeBase58 bitcoinAlphabet bs
+        let probablyUUID = decodeUUID decodedBS
+            probablyAddress = decodeAddress decodedBS
+        if isRight probablyUUID
```
akegalj: Why not using alternative here:
```haskell
decodeUUID decodeBS <|> decodeAddress decodeBS
```
?

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
```diff
@@ -0,0 +1,99 @@
...
+
+mkHdRootIdForFOWallet
+    :: NetworkMagic
+    -> EncryptedSecretKey
+    -> HdRootId
+mkHdRootIdForFOWallet nm esk = HdRootId
+    . decodeUtf8
+    . Core.addrToBase58
+    . (Core.makePubKeyAddressBoot nm)
+    . encToPublic $ esk
```
akegalj, line 103: hm - there is a very small likelihood that `mkHdRootIdForEOWallet` and `mkHdRootIdForFOWallet` might collide, right ? Could `nextRandom` from `mkHdRootIdForEOWallet` produce already existing `HDRootId`  (using old `Core.Address`) ?

We enforce a rule in decoding `decodeHdRootId` by first trying EO id and later FO id but the result is dependant only on our ordering...

akegalj (commented):
This is partial review - will finish it in few hours.

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
akegalj: I see this has been fixed though https://github.com/input-output-hk/cardano-wallet/pull/225/commits/3ec6f10e6d83933eb54bcdc4bb1ae33f1e2f079a#diff-3c97127e54449021987ec2d29990f40dR54

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
akegalj: I see this has been fixed as well https://github.com/input-output-hk/cardano-wallet/pull/225/commits/3ec6f10e6d83933eb54bcdc4bb1ae33f1e2f079a#diff-3c97127e54449021987ec2d29990f40dR77

KtorZ: Done already in https://github.com/input-output-hk/cardano-wallet/pull/225/commits/3ec6f10e6d83933eb54bcdc4bb1ae33f1e2f079a

Happy to see we concurred on that :)

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
KtorZ: There's a chance yes. But the same chance that we may generate two identical addresses from two different ESK I believe. This won't matter much since we enforce uniqueness of indexes in the DB. So, at the very least, we'll get an exception from the DB when trying to save a model with "duplicate" HdRootId.

denisshevchenko: `nextRandom` generates UUID v4, so it's not just "very small likelihood", it's absolutely insignificant probability. :-)

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
denisshevchenko: Already done.

src/Cardano/Wallet/Kernel/DB/HdRootId.hs
akegalj: :+1:

akegalj (commented):
Finished review - the rest of PR looks good

Issue #195, opened 2019-01-03
Add golden tests for bip44 derivations


akegalj:
# Context

In #41 we have implemented bip44 derivations and added property tests. Nicolas advised to use `cardano-cli` to generate golden tests for bip32/bip44 and add module for bip44ed255915 golden tests:
 * Nicolas advised to use rust's cardano-cli to generate golden tests https://github.com/input-output-hk/cardano-cli, https://cardanorust.iohkdev.io/cli-wallet/
 * use cardano-cli instructions https://github.com/input-output-hk/cardano-wallet/wiki/Testing-Yoroi-and-Cardano-wallet

# Decision

Add golden tests for bip44 derivations defined in #41. 

# Acceptance Criterias

1. Tests _must_ contain golden test for deriving public key from private key
2. Tests _must_ contain golden test for deriving account private key from master private key
3. Tests _must_ contain golden test for deriving address private key from account private key
4. Tests _must_ contain golden test for deriving address public key from account public key
5. Issue _should_ add tutorial about how to use `cardano-cli` to add golden tests as `cardano-wallet` wiki page

---

# Development Plan

- [x] I intend to extend `cardano-sl/mnemonic` package to support new seed generation as done by Icarus/Yoroi.

- [x] I intend to add support for passphrase 

- [x] I intend to add golden test in cardano-wallet to compare addresses generated from `cardano-cli` and the one obtained from `cardano-wallet` using the same set of mnemonic and password

- [x] I intend to extend the wiki documentation to explain how to generate new test vectors

# PR

| Number   | Base      |
| ---      | ---       |
| #227        | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| 1, 2, 3, 4 | I've taken a slightly different approach here and went for [golden test comparing the whole address](https://github.com/input-output-hk/cardano-wallet/blob/417853293e618c5bc09b0b2036c83b10afe616a3/test/unit/Golden/Ed25519Addresses.hs#L40-L68). Reasons a plural: <br/> 1/ We can compare them against addresses generated via `cardano-cli`, <br/> 2/ We don't have access to intermediate byte representations on `cardano-cli` (without instrumenting the rust code via debug and println and recompiling it locally, which is sort-of impractical) <br/> 3/ It achieves the same end-goal for if intermediate representations don't match, then the final address won't either. <br/> 4/ Tests are also quite easy to read and extend this way. 
[...]

PR #227, opened 2019-01-16
Golden Tests for Addresses à la BIP-44

KtorZ:
<p align="right">#195</p>

# Overview

- [x] I have implemented a method to generate `XPrv` / `EncryptedSecretKey` compatible with the new derivation scheme à la BIP-44 (& Icarus/Yoroi wallets)
- [x] I have added golden tests which illustrates and compare this generation and derivation of child keys with Yoroi / Icarus wallets.

# Comments

That was a bit confusing initially but after a chat with the Rust team, we discovered that root seed in `cardano-cli` (and therefore, Yoroi, Icarus) were generated differently from what we use to do. So, having the right derivation scheme and approach isn't sufficient; if, given the same mnemonics and passphrase, we do not generate the same root seed, all the rest is kinda pointless.

This PR fixes that :+1:

### DIFF
--- a/cardano-sl.yaml
+++ b/cardano-sl.yaml
@@ -33,7 +33,7 @@ packages:
   commit: 93f2246c54436e7f98cc363b4e0f8f1cb5e78717
 
 - git: https://github.com/input-output-hk/cardano-crypto
-  commit: 59eb60b3a227c615736db9bb1609fd5c20109098
+  commit: 45e1a0eafac774c109704be44ca06fd8cae193ba
 
 - git: https://github.com/andrewthad/haskell-ip
   commit: 9bb453139aa82cc973125091800422a523e1eb8f
--- a/test/unit/Golden/Ed25519Addresses.hs
+++ b/test/unit/Golden/Ed25519Addresses.hs
@@ -1,0 +1,197 @@
+module Golden.Ed25519Addresses
+    ( spec
+    ) where
+
+import           Universum hiding (pretty)
+
+import           Crypto.Encoding.BIP39 (ConsistentEntropy, EntropySize)
+import qualified Data.Aeson as Aeson
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as BL
+import qualified Data.Text as T
+import qualified Data.Text.Encoding as T
+import           Test.Hspec (Spec, describe, it, shouldBe)
+
+import           Cardano.Mnemonic (Mnemonic, mkMnemonic)
+import           Cardano.Wallet.Kernel.Ed25519Bip44 (ChangeChain (..),
+                     deriveAddressKeyPair, genEncryptedSecretKey,
+                     isInternalChange)
+import           Pos.Core (Address)
+import           Pos.Core.Common (IsBootstrapEraAddr (..), decodeTextAddress,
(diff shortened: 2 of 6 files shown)


cardano-sl.yaml
```diff
@@ -33,7 +33,7 @@ packages:
   commit: 93f2246c54436e7f98cc363b4e0f8f1cb5e78717
 
 - git: https://github.com/input-output-hk/cardano-crypto
-  commit: 59eb60b3a227c615736db9bb1609fd5c20109098
+  commit: 45e1a0eafac774c109704be44ca06fd8cae193ba
```
akegalj, line 36: :+1:

test/unit/Golden/Ed25519Addresses.hs
```diff
@@ -0,0 +1,197 @@
...
+
+spec :: Spec
+spec = do
+    describe "Ed25519 Addresses - Golden Tests" $ forM_ tests
+        $ \vec@(TestVector addr accIx change addrIx mw pw) -> it (titleize vec) $ do
+            let rootPassphrase = mempty
+            let esk = genEncryptedSecretKey (mw, pw) rootPassphrase
+            let pubKey = fst <$> deriveAddressKeyPair rootPassphrase esk accIx change addrIx
+            let addr' = makePubKeyAddress NetworkMainOrStage (IsBootstrapEraAddr True) <$> pubKey
+            addr' `shouldBe` (Just addr)
```
akegalj, line 48: :+1:

test/unit/Golden/Ed25519Addresses.hs
```diff
@@ -0,0 +1,197 @@
...
+  where
+    tests :: [TestVector]
+    tests =
+        [ TestVector
+            { _address = unsafeMkAddress "Ae2tdPwUPEZ1DYmhvpJWtVkMUbypPVkCVjQLNJeKRRG4LJ64dqTSRpWqzLH"
+            , _accountIx = 0x80000000
+            , _changeChain = ExternalChain
+            , _addressIx = 0x00000000
+            , _mnemonic = unsafeMkMnemonic @24
```
akegalj, line 57: should we test with less then 15 and more then 24 mnemonics ? (I expect it should be forbidden and that it should fail.

I guess we are only testing happy paths with golden tests. Maybe that's enough and expected though. Adding a possibility that address is `Nothing` https://github.com/input-output-hk/cardano-wallet/pull/227/files#diff-35fada5f498722b3f0bb76a4db32b5c4R27 to test vector and testing unhappy paths might be nice addition (or not?)

akegalj (approved):
LGTM

there is a small comment about testing but other than that it looks good!

test/unit/Golden/Ed25519Addresses.hs
KtorZ: We could indeed, and it was maybe a wrong choice to pick 15 and 24. Those weren't pick because they're plausible boundaries, but just... because they are plausible values.  

Also, there's no real point in testing failure cases in Golden tests, that's really not their intent. We do test failure paths for derivation in our property test suites already. Here, we do really want to control that our outputs match the one from cardano-cli, and that this doesn't change over time.

Issue #235, opened 2019-01-18
Remove WIP Paths and expose them in the API


KtorZ:
# Context

We marked #100 as done though, related endpoints for externally owned wallets are still under a WIP path and not quite ready to be used by anyone. 
In order to really be able to wrap this up, we do need to expose those endpoints through the API first.

# Decision

We'll expose those WIP paths through "normal" API sections and make them part of the existing public API.

# Acceptance Criterias

1. The API _must_ expose endpoints to deal with externally owned wallets (CRUD operations)
1. WIP modules _must_ disappear 

---

# Development Plan

- [x] I intend to remove WIP modules
- [x] I intend to expose new endpoints for to E.O. wallets in the API behind a proper prefix

# PR

| Number   | Base      |
| ---      | ---       |
| #235 | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| 1.  & 2.      | Endpoints have been removed, or more or less, moved into relevant places in the API: [here](https://github.com/input-output-hk/cardano-wallet/blob/a0da59848750e28762c2ac713242eba47613a61b/src/Cardano/Wallet/API/V1/Transactions.hs#L35-L43) and [here](https://github.com/input-output-hk/cardano-wallet/blob/a0da59848750e28762c2ac713242eba47613a61b/src/Cardano/Wallet/API/V1/Wallets.hs#L45-L85). The result can be viewed in the API documentation [here](https://input-output-hk.github.io/cardano-wallet/#tag/Externally-Owned-Wallets) and [here](https://input-output-hk.github.io/cardano-wallet/#tag/Transactions%2Fpaths%2F~1api~1v1~1transactions~1unsigned%2Fpost) and [here](https://input-output-hk.github.io/cardano-wallet/#tag/Transactions%2Fpaths%2F~1api~1v1~1transactions~1externally-signed%2Fpost)

---

# Retrospective

[...]

PR #237, opened 2019-01-18
Remove and integrate WIP APIs with rest of the code

KtorZ:
<p align="right">#235</p>

# Overview

- [x] I have remove WIP modules
- [x] I have moved the API handlers and declarations for WIP modules to relevant places
- [x] I have completed the CRUD operation for the API declaration of externally owned wallets
- [x] I have removed a bit of dead-code
- [x] I have removed few methods relative to externally owned wallets from the haskell client until we finalize the implementation.

# Comments


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -628,6 +624,41 @@ instance BuildableSafeGen NewEosWallet where
         neweoswalAssuranceLevel
         neweoswalName
 
+
+data UpdateEosWallet = UpdateEosWallet
+    { ueowalAssuranceLevel :: !AssuranceLevel
+    , ueowalName           :: !Text
+    , ueowalAddressPoolGap :: !AddressPoolGap
+    } deriving (Eq, Show, Generic)
```
KtorZ, line 629: cc @akegalj

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1952,35 +1949,6 @@ instance BuildableSafeGen SignedTransaction where
...
-
-deriveSafeBuildable ''WalletAndTxHistory
-instance BuildableSafeGen WalletAndTxHistory where
-    buildSafeGen sl WalletAndTxHistory{..} = bprint ("{"
-        %" wallet="%buildSafe sl
-        %" transactions="%buildSafe sl
-        %" }")
-        waltxsWallet
-        waltxsTransactions
-
```
KtorZ, line 1983: dead code is dead.

src/Cardano/Wallet/Client.hs
```diff
@@ -250,10 +246,6 @@ natMapClient phi f wc = WalletClient
         \x -> f . phi . updateWallet wc x
     , getUtxoStatistics =
         f . phi . getUtxoStatistics wc
-    , postEosWallet =
-        f . phi . postEosWallet wc
-    , deleteEosWallet =
-        f . phi . deleteEosWallet wc
```
KtorZ, line 256: removed until we finalize the E.O. implementation. We'll re-introduce when writing the integration tests.

src/Cardano/Wallet/API/V1/Handlers/Wallets.hs
```diff
@@ -102,3 +116,49 @@ getUtxoStatistics pwl wid = do
...
+    res <- liftIO $ WalletLayer.createEosWallet pwl req
+    case res of
+        Left e  -> throwM e
+        Right w -> return $ single w
+
+readEosWallets
+    :: PassiveWalletLayer IO
+    -> WalletId
+    -> Handler (APIResponse EosWallet)
+readEosWallets _ _ = do
```
akegalj, line 139: I find this name "readEosWallets" a bit missleading, no? 

Would we prefer "readEosWallet" (singular)?

src/Cardano/Wallet/API/V1/Handlers/Wallets.hs
```diff
@@ -102,3 +116,49 @@ getUtxoStatistics pwl wid = do
...
+
+--------------------------------------------------------------------------------
+-- EXTERNALLY OWNED WALLETS
+--------------------------------------------------------------------------------
+
+createEosWallets
+    :: PassiveWalletLayer IO
+    -> NewEosWallet
+    -> Handler (APIResponse EosWallet)
+createEosWallets pwl req = do
```
akegalj, line 129: maybe singular "createEosWallet" instead?

src/Cardano/Wallet/API/V1/Handlers/Wallets.hs
```diff
@@ -102,3 +116,49 @@ getUtxoStatistics pwl wid = do
...
+    -> Handler (APIResponse EosWallet)
+readEosWallets _ _ = do
+    throwM $ err501 { errBody = "Not Implemented" }
+
+updateEosWallets
+    :: PassiveWalletLayer IO
+    -> WalletId
+    -> UpdateEosWallet
+    -> Handler (APIResponse EosWallet)
+updateEosWallets _ _ _ = do
```
akegalj, line 147: maybe singular "updateEosWallet" instead?

src/Cardano/Wallet/API/V1/Handlers/Wallets.hs
```diff
@@ -102,3 +116,49 @@ getUtxoStatistics pwl wid = do
...
+    throwM $ err501 { errBody = "Not Implemented" }
+
+deleteEosWallets
+    :: PassiveWalletLayer IO
+    -> WalletId
+    -> Handler NoContent
+deleteEosWallets _ _ = do
```
akegalj, line 154: maybe singular "deleteEosWallet" instead

src/Cardano/Wallet/API/V1/Types.hs
akegalj: :goat: :+1:

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -877,17 +882,8 @@ data Wallet = Wallet {
     , walCreatedAt                  :: !WalletTimestamp
     , walAssuranceLevel             :: !AssuranceLevel
     , walSyncState                  :: !SyncState
-    , walType                       :: !WalletType
```
akegalj, line 880: :+1:

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -948,11 +939,12 @@ instance ToSchema PublicKey where
 
 -- | Externally-owned sequential (EOS) wallet (mobile client or hardware wallet).
 data EosWallet = EosWallet
-    { eoswalId             :: !EosWalletId
+    { eoswalId             :: !WalletId
```
akegalj, line 939: :+1:

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -174,7 +171,6 @@ import           Cardano.Wallet.API.V1.Generic (jsendErrorGenericParseJSON,
                      jsendErrorGenericToJSON)
 import           Cardano.Wallet.API.V1.Swagger.Example (Example, example)
 import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
-import           Cardano.Wallet.Kernel.EosWalletId (EosWalletId)
```
akegalj, line 177: I wonder can `Cardano.Wallet.Kernel.EosWalletId` module be deleted as well?

akegalj: Aha -I think its being used by `src/Cardano/Wallet/Kernel/DB/EosHdWallet.hs` . Maybe we can use `WalletId` instead of `EosWalletId` in the same way we did in this module

akegalj: and it is used in other places - as it differs with its UUID id - so now I am not sure how this alligns with https://github.com/input-output-hk/cardano-wallet/pull/237/files#diff-a61d46b2ed81b8f7899f4ac04a2e9a6eR942 . Do we really want to use `WalletId` in `EosWallet` or would we be better with new type here?

My understanding from https://github.com/input-output-hk/cardano-wallet/pull/225 : 
> Please note that using Core.Address as wallet's identifier isn't a good decision, but since exchanges and users may have wallets' ids stored on their side we have to keep backward-compatibility for existing ids.

was that current `WalletId` doesn't have the best choice for id's (not sure why is that though) and that UUID is a better approach (again, not sure what's the reasoning behind the claim)

@denisshevchenko maybe you can jump in to clear this for me

src/Cardano/Wallet/API/V1/Types.hs
akegalj: or not - see confusion https://github.com/input-output-hk/cardano-wallet/pull/237/files#diff-a61d46b2ed81b8f7899f4ac04a2e9a6eL177

src/Cardano/Wallet/API/V1/Types.hs
akegalj: can't grasp what was the use for the above code - so will skip on this one

src/Cardano/Wallet/API/V1/Wallets.hs
```diff
@@ -41,3 +41,45 @@ type API = Tags '["Wallets"] :>
...
+    :> ( "wallets"
+        :> "externally-owned"
+        :> Summary "Creates a new or restores an existing Wallet."
+        :> ReqBody '[ValidJSON] (NewEosWallet)
+        :> PostCreated '[ValidJSON] (APIResponse EosWallet)
+
+    :<|> "wallets"
+        :> "externally-owned"
+        :> Summary "Returns the Wallet identified by the given walletId."
+        :> CaptureWalletId
```
akegalj, line 55: aha - maybe the idea is to abstract and use the same walletId (in the same way we use `HDRootId`) . The same confusion about `WalletId` vs `EosWalletId` here

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -347,8 +348,7 @@ walletName :: HasType Text s => Lens' s Text
 walletName = typed
 
 spendingPasswordLastUpdate :: Lens' Wallet WalletTimestamp
-spendingPasswordLastUpdate f (Wallet v1 v2 v3 v4 spLU v6 v7 v8 v9) =
-    (\spLU' -> Wallet v1 v2 v3 v4 spLU' v6 v7 v8 v9) <$> f spLU
+spendingPasswordLastUpdate = field @"walSpendingPasswordLastUpdate"
```
akegalj, line 369: interesting - wasn't aware of this

akegalj (approved):
LGTM

PR looks ok - there are few stuff that are confusing to me - probably due to me not having enough context about EO wallets yet

There are minor namings issue (should be singular probably). Other then that there is a confusion between `EosWalletId` and `WalletId` use case in my mind

src/Cardano/Wallet/API/V1/Types.hs
KtorZ: This was added some time ago and never used. Just noticed while adjusting types for exposing the endpoints, so I removed it.

src/Cardano/Wallet/API/V1/Handlers/Wallets.hs
KtorZ: Actually had a good thought on that one. And went for the plural on purpose:

- We already use the plural form in several places
    - In API paths: e.g. `/api/v1/wallets`
    - In API groups tags: e.g. "Wallets"
    - In module names: e.g. `Handlers.Wallets`, `WalletLayer.Wallets` 

- It makes the overall a bit more consistent because we don't have to "think" whether to use the plural or singular because it returns one or more result (so we'd have `listWallets` or `getWallet`). 

- This reads as: 'read' action on 'EosWallets'. 


Naming is hard and... it's usually even harder to find an agreement. Note that I also refrained myself to rename all 'Eos' into just 'EO' because the sequential aspect has nothing to do with the reason being of the "external-related" methods. That's something we could do before release (and that goes also for the methods in this module). 

Renaming things is a valid refactoring technique if it helps making things clearer.

src/Cardano/Wallet/API/V1/Types.hs
KtorZ: `EosWalletId` will eventually disappear from the codebase. We'll stick with `WalletId` for API-level representation and convert `WalletId` to `HdRootId` in the lower levels. 
We can't really change existing wallet ids because this may give a hard time to our users, yet, we're free to support new types of IDs. Therefore, E.O. wallets will use a UUID, but encoded as Base58 as well for better consistency. 

#225 basically deprecates the current `EosWalletId` module.

test/integration/Test/Integration/Framework/DSL.hs
KtorZ: Was just a lens and because I removed the `WalletType` from the representation, it didn't compiled... Changed it using generic lenses and a field accessor to make it slightly more resilient to changes (beside fixing the compile error).

denisshevchenko (approved):

src/Cardano/Wallet/Client/Http.hs
```diff
@@ -178,6 +168,7 @@ mkHttpClient baseUrl manager = WalletClient
 
     addressesAPI
         :<|> walletsAPI
+        :<|> _
```
denisshevchenko, line 171: @KtorZ What does this `_` mean?

src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
```diff
@@ -167,10 +167,6 @@ toWallet db hdRoot = V1.Wallet {
     , walCreatedAt                  = V1.WalletTimestamp createdAt
     , walAssuranceLevel             = v1AssuranceLevel
     , walSyncState                  = V1.Synced
-    -- FIXME: Now we have 2 types of wallet: regular and external.
-    -- Currently there's only regular wallets, it will be changed in
-    -- next CHW-related PRs.
-    , walType                       = V1.WalletRegular
```
denisshevchenko, line 173: FINALLY! :+1: :+1:

src/Cardano/Wallet/Client/Http.hs
KtorZ: It's a type hole. As said above, I've removed the methods related to external wallets from the Client. Likely, those methods will still go over some refactoring before being ready, so this will spare us some unnecessary maintenance work. We'll re-introduce when ready for testing :)

PR #238, opened 2019-01-21
[BIP44-103] Track EOS wallets

uroboros:

# Overview

When the wallet applies blocks it prefilters for "our" addresses and in this way tracks the utxo and balance of a wallet. 

# Development plan

To enable tracking of EOS wallets

- [x] Implement ISOurs prefiltering for Hd Sequential wallets
- [ ] Persist EOS wallet Account PKs and AccountPools
- [ ] On wallet startup, recover EOS wallet AccountPKs and use track the EOS accounts during applyBlock
- [ ] test EOS tracking
 
# Comments

- This does not include extending Restoration of EOS wallet

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
+++ b/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
@@ -98,8 +98,9 @@ module Cardano.Wallet.Kernel.DB.HdWallet (
 
 import           Universum hiding ((:|))
 
-import           Control.Lens (Getter, at, lazy, to, (+~), _Wrapped)
+import           Control.Lens (Getter, at, lazy, to, (+~), (?~), _Wrapped)
 import           Control.Lens.TH (makeLenses)
+import qualified Data.Map as Map
 import qualified Data.IxSet.Typed as IxSet (Indexable (..))
 import           Data.SafeCopy (base, deriveSafeCopy)
 
@@ -116,6 +117,7 @@ import           Pos.Crypto (HDPassphrase)
 import qualified Pos.Crypto as Core
 
 import           Cardano.Wallet.API.V1.Types (WalAddress (..))
+import           Cardano.Wallet.Kernel.AddressPool (AddressPool, lookupAddressPool)
 import           Cardano.Wallet.Kernel.DB.BlockContext
 import           Cardano.Wallet.Kernel.DB.HdRootId (HdRootId)
 import           Cardano.Wallet.Kernel.DB.InDb
@@ -209,6 +211,10 @@ deriveSafeCopy 1 'base ''HasSpendingPassword
 class IsOurs s where
     isOurs :: Core.Address -> s -> (Maybe HdAddress, s)
 
+{-------------------------------------------------------------------------------
+  isOurs for Hd Random wallets
+-------------------------------------------------------------------------------}
+
 -- | NOTE: We could modify the given state here to actually store decrypted
 -- addresses in a `Map` to trade a decryption against a map lookup for already
 -- decrypted addresses.
@@ -219,6 +225,7 @@ instance IsOurs [(HdRootId, Core.EncryptedSecretKey)] where
         let addrId = HdAddressId accId addressIx
         return $ HdAddress addrId (InDb addr)
 
+-- TODO @uroboros Delete or generalise?
 -- | Extract our addresses from block outputs.
 ourAddresses
     :: [TxOutAux]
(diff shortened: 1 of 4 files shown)


src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -238,6 +245,36 @@ decryptHdLvl2DerivationPath hdPass addr = do
...
+--  Otherwise if we don't find the address, we return the state unchanged.
+instance IsOurs (Map HdAccountId (AddressPool Core.Address)) where
+    isOurs addr pools = case lookupAddressInPools of
+        (Just (hdAddr, pool')) ->
+            (Just hdAddr, pools & at (accountId hdAddr) ?~ pool')
+        Nothing ->
+            (Nothing,pools)
+
+        where
+            accountId a = a ^. to _hdAddressId . to _hdAddressIdParent
```
KtorZ: can't use `hdAddressId` and `hdAddressIdParent` directly :thinking: ? Is it because of TH and the location of this instance?

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -238,6 +245,36 @@ decryptHdLvl2DerivationPath hdPass addr = do
...
+                = foldl' (<|>) Nothing $ map lookupAddressInPool' (Map.toList pools)
+
+            lookupAddressInPool' (accId, pool)
+                = case lookupAddressPool addr pool of
+                    (Nothing, _)          ->  Nothing
+                    (Just (_, ix), pool') -> (Just (mkHdAddress accId ix, pool'))
+
+            mkHdAddress accId_ ix_
+                = HdAddress (HdAddressId accId_ (HdAddressIx (fromIntegral ix_)))
+                            (InDb addr)
```
KtorZ: No more variable-length vertical indent please :pray:

uroboros:
@KtorZ in this implementation of EOS isOurs: 

We prefilter each transaction/block for a wallet of EOS accounts `Map AccountId AccountPool`, this means that when we do prefiltering at the call sites, we'll need to do a call for each "type of state", `HdSeq`, `HdEOS` (and perhaps `HdFOS` also). Is this how you imagined it?

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -238,6 +245,36 @@ decryptHdLvl2DerivationPath hdPass addr = do
...
+    isOurs addr pools = case lookupAddressInPools of
+        (Just (hdAddr, pool')) ->
+            (Just hdAddr, pools & at (accountId hdAddr) ?~ pool')
+        Nothing ->
+            (Nothing,pools)
+
+        where
+            accountId a = a ^. to _hdAddressId . to _hdAddressIdParent
+
+            lookupAddressInPools
```
KtorZ: Minor things but the name is slightly misleading here. It suggests that this is a function while it's actually a result `Maybe (HdAddress, AddressPool Address)`

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -238,6 +245,36 @@ decryptHdLvl2DerivationPath hdPass addr = do
...
+--  (since the address pool may have been extended in response to the address discovery)
+--  Otherwise if we don't find the address, we return the state unchanged.
+instance IsOurs (Map HdAccountId (AddressPool Core.Address)) where
+    isOurs addr pools = case lookupAddressInPools of
+        (Just (hdAddr, pool')) ->
+            (Just hdAddr, pools & at (accountId hdAddr) ?~ pool')
+        Nothing ->
+            (Nothing,pools)
+
+        where
```
KtorZ: I'd suggest adding type-signatures to functions below. This serves as free-doc checked by the compiler :smile:

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
KtorZ: Also, have a look at `initHdAddress` :+1:

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: yes, the Lens is defined below this definition, we could move this to later in the file?

KtorZ: Yeah, let's pull down all the `IsOurs` stuff later in the file to keep things grouped :+1:

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -542,6 +485,80 @@ hdAccountRestorationState a = case a ^. hdAccountState of
...
+-- addresses in a `Map` to trade a decryption against a map lookup for already
+-- decrypted addresses.
+instance IsOurs [(HdRootId, Core.EncryptedSecretKey)] where
+    isOurs addr s = (,s) $ foldl' (<|>) Nothing $ flip map s $ \(rootId, esk) -> do
+        (accountIx, addressIx) <- decryptHdLvl2DerivationPath (eskToHdPassphrase esk) addr
+        let accId = HdAccountId rootId accountIx
+        let addrId = HdAddressId accId addressIx
+        return $ HdAddress addrId (InDb addr)
+
+-- TODO @uroboros Delete or generalise?
```
KtorZ: I believe this can now be deleted indeed :+1:
It's not used anywhere.

KtorZ (approved):

KtorZ:
@uroboros 

> We prefilter each transaction/block for a wallet of EOS accounts Map AccountId AccountPool, this means that when we do prefiltering at the call sites, we'll need to do a call for each "type of state", HdSeq, HdEOS (and perhaps HdFOS also). Is this how you imagined it?

We'll have to pick one or the other yes. That'll depends on the type of wallet being used.

Issue #231, opened 2019-01-17
Finalize HD hierarchy for EOS-wallets


denisshevchenko:
# Context

We want to unify `Hd` type hierarchy for EOS-wallets. The main goal is to use existing `Hd***`-types as much as possible. Furthermore, we want to unify `HdRootId` both for FO-wallets and EOS-wallets, with backward-compatibility.

# Decision

# Acceptance Criterias

---

# Development Plan

- [x] Update `HdRootId` type to something more opaque.
- [x] Introduce a new branch in `HdAccount` to have room for EO stuff (without implementing anything on E.O side yet, but just making sure that the existing F.O. stuff keeps on working fine).
- [x] Add EO-specific stuff in EO-branch in `HdAccountBase`.
- [x] Replace `EosHd***` types by their equivalent `Hd***` types and leverage the `HdAccount` branch to store and retrieve the relevant information.
- [x] Change API-request for EOS-wallet creation: user must provide not only list of accounts' public keys, but list of their indexes as well.
- [x] `getAddressPool` function is added.

Separate PRs will be opened for each of steps.

# PR

| Number   | Base      |
| ---      | ---       |
| #225       | `develop` |
| #243       | `develop` |
| #247       | `develop` |
| #250       | `develop` |
| #255       | `develop` |
| #256       | `develop` |
| #273        | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| ?        | -        |

Technically, this issue is (mostly) about code refactoring/restructuring, not about new features. The only exception is:

> Change API-request for EOS-wallet creation: user must provide not only list of accounts' public keys, but list of their indexes as well.

But this change is _already_ covered by tests in PR #334.

---

# Retrospective

KtorZ:
@denisshevchenko What's missing to send this one in QA :thinking: ?

piotr-iohk:
Well, there is QA section missing ^^ ;)
@denisshevchenko, can you fill it in please?

piotr-iohk:
lgtm :+1:

PR #242, opened 2019-01-21
Additional fields for EOS-stuff in 'HdWallets'.

denisshevchenko:
<p align="right">#231</p>

# Overview

- [x] I have added new `IxSet` for `AddressPoolGap`.
- [x] I have added new `IxSet` for accounts' public keys.

# Comments

We have an opaque `HdRootId` now, so it's possible to use existing type hierarchy (`HdRoot`, `HdAccount` and `HdAddress`) to store EOS-wallets' stuff (common for both types of wallets). But since we have a _specific_ EOS-stuff, we introduce two new fields in `HdWallets` to store it:

1. `IxSet` for `HdRootAddressPoolGap` (one value for each EOS-wallet).
2. `IxSet` for `HdAccountPublicKey` (one value for each account in EOS-wallet).

SInce it's `IxSet`, we'll be able to introduce (in the next PR) getters we need:

```
addressPoolGap :: HdRootId -> Maybe AddressPoolGap

accountPublicKey :: HdAccountId -> Maybe PublicKey
```

### DIFF
--- a/src/Cardano/Wallet/Kernel/AddressPoolGap.hs
+++ b/src/Cardano/Wallet/Kernel/AddressPoolGap.hs
@@ -14,6 +14,7 @@ import           Universum
 import qualified Data.Aeson.Options as Aeson
 import           Data.Aeson.TH
 import           Data.Default (Default (..))
+import           Data.SafeCopy (base, deriveSafeCopy)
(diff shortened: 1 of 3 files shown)


denisshevchenko:
After discussion I've realized that this design isn't good. Closed in favor of next PR. Branch will be deleted later.

PR #243, opened 2019-01-22
Introduce new branch for EOS-stuff.

denisshevchenko:
<p align="right">#231</p>

# Overview

- [x] New type `HdRootBase` is added, with two branches for FO-stuff and EOS-stuff.
- [x] `HdRoot` is changed: `HdRootBase` instead of `HdRootId`.
- [x] Corresponding lenses and functions are changed.

# Comments

Introduce a new branch in `HdRoot` to have room for EO-stuff (without implementing anything on E.O side yet, but just making sure that the existing F.O. stuff keeps on working fine).

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
+++ b/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
@@ -19,6 +19,7 @@ module Cardano.Wallet.Kernel.DB.HdWallet (
   , HdWallets(..)
   , HdAccountId(..)
   , HdAddressId(..)
+  , HdRootBase(..)
   , HdRoot(..)
   , HdAccount(..)
   , HdAddress(..)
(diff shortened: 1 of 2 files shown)


src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -250,6 +252,23 @@ instance Ord HdAddressId where
...
+      -- Branch for EOS-wallets.
+      -- TODO: Add new EOS-related stuff in the next PR,
+      -- please see issue #231.
+    | HdRootBaseEO !HdRootId
+    deriving (Eq, Show)
+
+instance Arbitrary HdRootBase where
+    arbitrary = oneof
+        [ HdRootBaseFO <$> arbitrary
+        , HdRootBaseFO <$> arbitrary
```
KtorZ: Second should be `EO` I suppose. Maybe we should go for slightly longer name to improve readability before we commit to having a lot of `EO` and `FO` throughout the codebase :thinking:

denisshevchenko: Fixed.

uroboros (commented):
LGTM

KtorZ (approved):

PR #245, opened 2019-01-23
Add EO-specific stuff in HdRootBase.

denisshevchenko:
<p align="right">#231</p>

# Overview

- [x] I have added a `Map` for address pools per account.
- [x] I have added a `Map` for accounts' public keys.

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/AddressPool.hs
+++ b/src/Cardano/Wallet/Kernel/AddressPool.hs
@@ -26,9 +26,12 @@ import           Universum
 
 import           Control.Lens (at, makeLenses)
 import qualified Data.Map as Map
+import           Data.SafeCopy (SafeCopy (..), contain, safeGet, safePut)
 import           Formatting (bprint, build, int, (%))
 import qualified Formatting.Buildable
 import           Serokell.Util (listJson)
+
+import           Test.QuickCheck (Arbitrary (..))
 
 import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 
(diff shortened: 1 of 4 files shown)


src/Cardano/Wallet/Kernel/AddressPool.hs
```diff
@@ -58,6 +61,23 @@ instance Buildable address => Buildable (AddressPool address) where
...
+    putCopy pool = contain $ do
+        safePut $ Map.toList (pool ^. addresses)
+        safePut (pool ^. gap)
+        -- | Please note that 'nextAddresses' is missing, because it'a a function
+        -- and we cannot serialize it.
+    getCopy = contain $ AddressPool
+        <$> (Map.fromList <$> safeGet)
+        <*> safeGet
+        -- Dummy function, because we didn't store a real function.
+        <*> pure (\_ -> Map.empty)
```
denisshevchenko, line 74: I'm not sure if such a `SafeCopy` instance is what we _actually_ need.

src/Cardano/Wallet/Kernel/DB/EosHdWallet.hs
```diff
@@ -93,7 +93,6 @@ makeLenses ''EosHdRoot
 makeLenses ''EosHdAccount
 makeLenses ''EosHdWallets
 
-deriveSafeCopy 1 'base ''AddressPoolGap
```
denisshevchenko, line 96: I moved it in `AddressPoolGap`'s module.

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -294,7 +309,7 @@ data HdRoot = HdRoot {
 
       -- | When was this wallet created?
     , _hdRootCreatedAt   :: !(InDb Core.Timestamp)
-    } deriving (Eq, Show)
```
denisshevchenko, line 297: I removed it because otherwise I have to define `Show` instance for `AddressPool` as well, but I cannot do it without _custom_ `Show` instance which are forbidden in our code.

denisshevchenko:
Closed in favor of #247

PR #249, opened 2019-01-23
Akegalj/45/remove extra argument

akegalj:
<p align="right">#45</p>

# Overview

- [x] I have removed extra argument from `pay` method from `WalletLayer`

# Comments

### DIFF
--- a/src/Cardano/Wallet/API/V1/Handlers/Transactions.hs
+++ b/src/Cardano/Wallet/API/V1/Handlers/Transactions.hs
@@ -14,8 +14,6 @@ import           Universum
 
 import           Servant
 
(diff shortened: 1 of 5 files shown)


src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -594,6 +596,27 @@ instance Arbitrary SignTransactionError where
         , SignTransactionErrorNotOwned <$> arbitrary
         ]
 
+-- Which address scheme is being used
+data AddressSchemeVersion
```
akegalj: ah - this is extra - will force push

KtorZ (approved):

PR #250, opened 2019-01-23
Allow AddressPool to be created from an initial list of addresses

KtorZ:
<p align="right">#231</p>

# Overview

- [x] I have extended the constructor of `AddressPool` to also accept an initial list of addresses + indexes
- [x] I have extended the test suite to reflect that change and tests the behavior when various initial (valid) list of addresses are given
- [x] I have provided an accessor to retrieve all known addresses from the pool gap (and allow future serialization / persistence of those addresses)

# Comments

This is for allowing serialization of the underlying pool. In theory,
we should never hit the 'Left' side of the constructor because given
addresses are generated only using the pool and nothing more.

### DIFF
--- a/src/Cardano/Wallet/Kernel/AddressPool.hs
+++ b/src/Cardano/Wallet/Kernel/AddressPool.hs
@@ -15,11 +15,23 @@
 -------------------------------------------------------------------------------}
 
 module Cardano.Wallet.Kernel.AddressPool
-    ( AddressPool
+    (
+    -- * Types
+      AddressPool
+    , ErrAddressPoolInvalid(..)
+
+    -- * Construction
+    , emptyAddressPool
     , initAddressPool
+
+    -- * Manipulation
     , lookupAddressPool
+
+    -- * Inspection
     , getAddressPoolSize
     , getAddressPoolGap
+    , getKnownAddresses
+    , verifyPool
     ) where
 
 import           Universum
@@ -29,6 +41,7 @@ import qualified Data.Map as Map
 import           Formatting (bprint, build, int, (%))
 import qualified Formatting.Buildable
 import           Serokell.Util (listJson)
+import           Test.QuickCheck (Arbitrary (..), elements)
 
 import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 
@@ -41,13 +54,13 @@ import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 -- varying wallet types
 data AddressPool address = AddressPool
     { _addresses -- all addresses currently in the pool, discovered or not
-        :: Map address Word
+        :: !(Map address Word32)
 
     , _gap -- the "gap" to maintain after the last discovered address in a pool
-        :: AddressPoolGap
+        :: !AddressPoolGap
 
     , _nextAddresses -- compute next 'gap' addresses from an index
-        :: Word -> Map address Word
+        :: Word32 -> Map address Word32
     }
 makeLenses ''AddressPool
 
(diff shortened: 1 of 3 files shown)


src/Cardano/Wallet/Kernel/AddressPool.hs
```diff
@@ -41,13 +44,13 @@ import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
...
     { _addresses -- all addresses currently in the pool, discovered or not
-        :: Map address Word
+        :: Map address Word32
 
     , _gap -- the "gap" to maintain after the last discovered address in a pool
         :: AddressPoolGap
 
     , _nextAddresses -- compute next 'gap' addresses from an index
-        :: Word -> Map address Word
+        :: Word32 -> Map address Word32
```
KtorZ, line 63: Moving to `Word32` to save us some `fromIntegral` here and there. Ideally, We could also just parameterize over `ix` which has `Num ix, Ord ix, Enum ix`, that should work the same.

src/Cardano/Wallet/Kernel/AddressPool.hs
```diff
@@ -66,14 +69,16 @@ instance Buildable address => Buildable (AddressPool address) where
...
     :: (Ord address)
     => AddressPoolGap
-    -> (Word -> address)
-    -> AddressPool address
-initAddressPool gap_ newAddress = AddressPool
-    { _addresses = nextAddresses_ 0
+    -> (Word32 -> address)
+    -> [(address, Word32)]
+    -> Either ErrAddressPoolInvalid (AddressPool address)
+initAddressPool gap_ newAddress addrs0 = verifyPool $ AddressPool
```
KtorZ: So, we try to construct a pool from the given indexes, and, at the end, we verify that all indexes we keep track of are indeed sequential. We should never hit that `Left` side as long as we only extend pools using the address pool module. Yet, we never know.

src/Cardano/Wallet/Kernel/AddressPool.hs
```diff
@@ -66,14 +69,16 @@ instance Buildable address => Buildable (AddressPool address) where
...
+    { _addresses = addresses_
     , _gap = gap_
     , _nextAddresses = nextAddresses_
     }
   where
+    addresses_ = Map.fromList addrs0 <> nextAddresses_ (fromIntegral $ length addrs0)
```
KtorZ: So, to initialize the pool, we also always need to look ahead for some addresses (the next gap).

src/Cardano/Wallet/Kernel/AddressPool.hs
```diff
@@ -114,6 +119,35 @@ getAddressPoolGap
     :: AddressPool address -> AddressPoolGap
 getAddressPoolGap pool = pool ^. gap
 
+getKnownAddresses
+    :: AddressPool address
+    -> [(address, Word32)]
+getKnownAddresses pool =
+    let xs = sortOn snd $ Map.toList (pool ^. addresses)
+    in take (length xs - fromIntegral (pool ^. gap))  xs
```
KtorZ, line 173: Debatable. But if think it makes sense to only store addresses we have discovered up until now; that is, we have to remove `gap` addresses from the end.

test/unit/Test/Spec/AddressPool.hs
```diff
@@ -3,36 +3,82 @@ module Test.Spec.AddressPool (spec) where
...
+-------------------------------------------------------------------------------}
+
+newtype InitialAddresses address
+    = InitialAddresses [(address, Word32)]
+    deriving (Show)
+
+instance Arbitrary (InitialAddresses Word32) where
+    arbitrary = do
+        n <- choose (0, 100)
+        return $ InitialAddresses $ zip [0..n] [0..]
```
KtorZ: Nothing fancy here, we still model addresses with `Word32`, but we have to make sure they're unique. So, for generating valid list of addresses, we just take `n` unique natural numbers.

src/Cardano/Wallet/Kernel/AddressPool.hs
uroboros: `initAddressPool` is used when we create a new EOS wallet, in which case there are no initial addresses. I think this must remain unchanged and then we need a separate "create existing AddressPool", which uses the HdAddrs in AcidState for the given account, and creates a pool with purely those addrs. This way, if an addr is in AcidState, we consider it "known". In this way, the addr pool is only extended on "create new EOS with addr pool" and on BIP 44-style addr discovery

src/Cardano/Wallet/Kernel/AddressPool.hs
uroboros: `getKnownAddrs` becomes trivial if we separate initAPool/readAPool (just the list of HdAddrs in AcidState for the given account)

uroboros (changes requested):

src/Cardano/Wallet/Kernel/AddressPool.hs
KtorZ: Agreed. We can provide both constructors.

src/Cardano/Wallet/Kernel/AddressPool.hs
KtorZ: That is only true provided we persist the list of undiscovered addresses. I was reluctant in doing so yesterday because we probably don't want undiscovered addresses in the DB.

On the other hand, we'll need to allow user to view addresses from the edge as their available addresses. And therefore, we might as well just persist that.

uroboros: keep in mind for HdRandom wallets, we persist undiscovered addrs (and mark them as "Used" when we actually discover)

KtorZ:
@uroboros ^^^

uroboros (approved):
I wonder about the naming of `emptyAPool` (which is not empty) vs `initAPool`. How about `newAPool` + `initAPool`?

Issue #252, opened 2019-01-24
Documentation isn't build nor uploaded from the CI anymore 


KtorZ:
| Release | Operating System       | Cause |
| ---     | ---                    | ---   |
| next    | Windows & OSX & Linux) | Configuration |

# Context
 
 

The CI has a special "documentation" job when things get merged into develop. This builds the documentation and publish it to the right `gh-pages` branch (then hosted here: https://input-output-hk.github.io/cardano-wallet/). However, this job is "allowed to fail" since the script will abort when there's actually nothing to push. 

The failure happened after we renamed the executables, but forgot to do it in the travis configuration.

# Steps to Reproduce

1. Merge something into develop
1. Wait for the "Documentation" job to fail

## Expected behavior

When there's an API change, the documentation is built and published to the repository's gh-pages

## Actual behavior
 
<A clear and concise description of what you observe instead. If applicable add screenshots to help explain your problem.>

https://travis-ci.org/input-output-hk/cardano-wallet/jobs/483553527

---

# Resolution Plan

- [x] Provide the right executable in the travis configuration
- [x] (bonus) Ideally, make this job "required" and find a work-around when there's nothing to push

# PR

| Number   | Base      |
| ---      | ---       |
| #253        | `develop` |

# QA 

The documentation job is back https://travis-ci.org/input-output-hk/cardano-wallet/jobs/483806387 and works correctly. It should now run fine even if there's nothing to upload.

piotr-iohk:
LGTM

PR #253, opened 2019-01-24
Fix 'Documentation' job in .travis.yml

KtorZ:
<p align="right">#252</p>

# Overview

- [ ] I have made the documentation job required. So we will now be notified of failures sooner!
- [ ] I have allowed empty commits to be published to gh-pages (this will prevent the job from failing if there's nothing to publish)
- [ ] I have modified the executable for creating the documentation to its new name
- [ ] (bonus) I have fixed the display of integration-test logs 

# Comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -186,7 +186,8 @@ jobs:
     # Run all tests with code coverage
     script:
     - tar xzf $HOME/.local/stack-work.tar.gz # Cached extra-dependencies
-    - travis_wait 42 stack --no-terminal test cardano-wallet:test:integration --fast || cat ./state-integation/logs/*.lob.pub
+    - travis_wait 42 stack --no-terminal test cardano-wallet:test:integration --fast
+    - cat ./state-integration/logs/*.log.pub
 
   # FULL COVERAGE
   - stage: tests
@@ -221,12 +222,11 @@ jobs:
     # Upload API doc automatically on each build against `develop` to `gh-pages`
     script:
     - tar xzf $HOME/.local/stack-work.tar.gz # Cached extra-dependencies
-    - travis_wait 42 stack --no-terminal build cardano-wallet:exe:cardano-generate-swagger-file --fast
-    - stack exec -- cardano-generate-swagger-file --target wallet@v1
+    - travis_wait 42 stack --no-terminal build cardano-wallet:exe:cardano-wallet-generate-swagger --fast
+    - stack exec -- cardano-wallet-generate-swagger --target wallet@v1
     - git add swagger.json && git commit -m $TRAVIS_COMMIT
-    - git checkout gh-pages && git cherry-pick -X theirs -
+    - git checkout gh-pages && git cherry-pick -X theirs -n - && git commit --allow-empty --no-edit
     - git push -f -q https://KtorZ:$GITHUB_API_KEY@github.com/input-output-hk/cardano-wallet gh-pages &>/dev/null # TODO: Create a specific CI user with its own key
 
   allow_failures:
-    - env: ACTION=documentation
     - env: ACTION=weeder


.travis.yml
```diff
@@ -186,7 +186,8 @@ jobs:
     # Run all tests with code coverage
     script:
     - tar xzf $HOME/.local/stack-work.tar.gz # Cached extra-dependencies
-    - travis_wait 42 stack --no-terminal test cardano-wallet:test:integration --fast || cat ./state-integation/logs/*.lob.pub
+    - travis_wait 42 stack --no-terminal test cardano-wallet:test:integration --fast
+    - cat ./state-integration/logs/*.log.pub
```
KtorZ, line 190: There were 2 typos : 
- state-inte**ga**tion -> state-inte**gra**tion
- .l**ob**.pub -> .l**og**.pub

+, I've moved the line to the next one. Travis will keep on executing this even in case of failures. So we'll have logs always, even in case of success. That could be useful also for comparing stuff.

.travis.yml
```diff
@@ -217,16 +218,15 @@ jobs:
   # DOCUMENTATION
   - stage: documentation
     env: ACTION=documentation
-    if: type = push AND branch = develop
+    # if: type = push AND branch = develop TEMPORARY TO TEST THE PR
```
KtorZ: This condition prevent the documentation from being built and published on each PR. However, we do want to test that out before merging it.
Will remove the comment once we're convinced the changes below run as expected.

.travis.yml
```diff
@@ -217,16 +218,15 @@ jobs:
...
     # Upload API doc automatically on each build against `develop` to `gh-pages`
     script:
     - tar xzf $HOME/.local/stack-work.tar.gz # Cached extra-dependencies
-    - travis_wait 42 stack --no-terminal build cardano-wallet:exe:cardano-generate-swagger-file --fast
-    - stack exec -- cardano-generate-swagger-file --target wallet@v1
+    - travis_wait 42 stack --no-terminal build cardano-wallet:exe:cardano-wallet-generate-swagger --fast
+    - stack exec -- cardano-wallet-generate-swagger --target wallet@v1
```
KtorZ, line 226: This was the actual reason of the failure.

.travis.yml
```diff
@@ -217,16 +218,15 @@ jobs:
...
     - git add swagger.json && git commit -m $TRAVIS_COMMIT
-    - git checkout gh-pages && git cherry-pick -X theirs -
+    - git checkout gh-pages && git cherry-pick -X theirs -n - && git commit --allow-empty --no-edit
```
KtorZ, line 228: Changing this line a bit. 

`-n` for `git cherry-pick` will stop the cherry-picking before the content it actually committed. This allow us to commit manually using `--allow-empty` in case there's indeed nothing to commit (because the documentation has not changed compared to before). 

This is slightly artificial, but at least, builds for documentation will now always succeed and we can remove them from the `allowed_failures` below. Therefore, if they fail again, we'll be notified by email and will know that something wrong is going on.

.travis.yml
piotr-iohk: Looking at the CI now -> https://travis-ci.org/input-output-hk/cardano-wallet/jobs/483806387

`cat ./state-integration/logs/*.log.pub` shows now a lot of logs... to the point that Travis won't show all of them :) :

>This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log.

Perhaps it would be more convenient to zip and attach the log directory `./state-integration/logs/` as a downloadable artifact to the job. It would certainly made the log with the test results more readable and clear. I suppose that analyzing this long log file directly in Travis when the test fails would not be convenient as well?

KtorZ: Hmmm.. that'll require a bit more work to instrument actually. Especially since Travis doesn't support build artifacts on PRs (only on merges or cron jobs). 
What about: aborting on first failure and show last `n` lines of logs.

That should be sufficient to help debbuging. What do you think @piotr-iohk ?

KtorZ: Well, the raw log is actually already available :tada:

https://api.travis-ci.org/v3/job/483806387/log.txt

KtorZ: image

piotr-iohk: I would be actually against aborting at first failure since the first failure would cover all the subsequent potential failures and we would have no knowledge about them until we fix the first failure...

I mean I can live with all the logs being shown. It is more _noisy_  and doesn't seem to be handy to analyse (to me), but that's fine if you think it makes sense. (Thing is, probably I won't be doing thorough log analysis in case of failure, so if it works for you [or developers in general] that's fine by me :) )

Yes, the log is available, but, I suppose that you have the contents of the `*.log.pub` there only because you are doing `cat ./state-integration/logs/*.log.pub` in the `.travis.yml` file.

KtorZ: Unfortunately right now, I don't think we can spend time in doing this _the right way_. In the end, what we'll get from an artifact is one copy/paste away from the raw logs. So that'll be quite a lot of efforts for little gains in the end. 

It's noisy, that's is true, but not because of the approach in itself, but because our logs are noisy (especially since the log file contains logs of all 5 nodes in the cluster thanks to a bug in the logging mechanism >.>)

piotr-iohk (approved):
overall, sounds good :)

PR #255, opened 2019-01-24
Replace EosHd-types by their equivalent Hd-types

denisshevchenko:
<p align="right">#231</p>

# Overview

- [x] I have removed `EosHd*`-stuff.
- [x] Roots and accounts for EOS-wallets are stored as for EO-wallet.

# Comments

Changes in API (accounts' indexes with public keys) will be added in the next PR.


src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -548,26 +541,27 @@ createHdWallet newRoot defaultHdAccountId defaultHdAddress utxoByAccount =
...
-        , eosAccountUpdate       = return () -- just need to create it, no more
+    mkUpdate :: AddressPoolGap
+             -> (PublicKey, HdAccountIx)
+             -> AccountUpdate HD.CreateHdRootError ()
+    mkUpdate addrPoolGap (accountPublicKey, accountIx) = AccountUpdate {
+          accountUpdateBase  = AccountUpdateBaseEO (HdAccountId (newRoot ^. hdRootId) accountIx)
+                                                   accountPublicKey
+                                                   addrPoolGap
+        , accountUpdateNew   = AccountUpdateNewUpToDate mempty
+        , accountUpdateAddrs = [] -- No addresses yet.
```
uroboros, line 564: we need to add the new address pool addresses here (a new EOS Wallet-account must has gap number of HdAddrs)

uroboros (approved):

denisshevchenko:
@KtorZ Conflicts (after merging of #256) are fixed.

src/Cardano/Wallet/Kernel/DB/HdWallet/Create.hs
```diff
@@ -147,18 +147,20 @@ initHdRoot rootId name hasPass assurance created = HdRoot {
...
     , _hdAccountState = st
     , _hdAccountAutoPkCounter = AutoIncrementKey 0
     }
   where
-    defName = AccountName $ sformat ("Account: " % build)
-                                    (accountId ^. hdAccountIdIx)
+    defName = AccountName $ sformat ("Account: " % build) accountIx
+    HdAccountId _ accountIx = case accountBase of
+        HdAccountBaseFO accId     -> accId
+        HdAccountBaseEO accId _ _ -> accId
```
KtorZ, line 163: Not sure we do want this (this = taking a HdAccountBase as an argument and doing a pattern matching inside the function). It'd make more sense to have separated functions for `initFOAccont` and `initEOAccount`. They'll likely take different parameters and, we don't want to mix them up in single functions. 

Thus, I'd keep the argument as it is right now, and only rename the function.

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -707,14 +699,18 @@ accountUpdateCreate accId (AccountUpdateNewIncomplete curUtxo genUtxo ctx) =
...
-             ((accountUpdateId, ) <$> accountUpdate)
+             (accountUpdateCreate accountUpdateBase accountUpdateNew)
+             accountId
+             ((accountId, ) <$> accountUpdate)
     mapM_ createAddress accountUpdateAddrs
     return res
   where
+    getAccountId updateBase = case updateBase of
+        AccountUpdateBaseFO accId     -> accId
+        AccountUpdateBaseEO accId _ _ -> accId
```
KtorZ, line 713: This (and all other occurrences of this) sounds like a good use-case for a lens. Actually, there's even already a function in `Kernel/DB/HdWallet` which does exactly that:

```hs
getHdAccountId :: HdAccountBase -> HdAccountId
getHdAccountId (HdAccountBaseFO accountId)     = accountId
getHdAccountId (HdAccountBaseEO accountId _ _) = accountId
```

This is likely something we'll keep on doing. So maybe having a lens here would be nice too (more consistent with the existing code-base). We already have:

```hs
hdAccountId :: Lens' HdAccount HdAccountId
hdAccountId = hdAccountBase . lens getHdAccountId setHdAccountId
```

So, maybe a `HdAccountBaseId` would do :+1:

KtorZ (approved):

PR #256, opened 2019-01-24
API changes for EOS-wallets creation: accounts public keys with indexes.

denisshevchenko:
<p align="right">#231</p>

# Overview

- [x] During EOS-wallet creation not only account's public key is provided, but its index as well.

# Comments

Since we receive accounts' public keys as they are, we need their indexes as well.

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -57,6 +57,7 @@ module Cardano.Wallet.API.V1.Types (
   , AccountIndex
   , AccountAddresses (..)
   , AccountBalance (..)
+  , AccountPublicKeyWithIx (..)
   , getAccIndex
   , mkAccountIndex
   , mkAccountIndexM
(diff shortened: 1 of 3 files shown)


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1138,6 +1101,73 @@ instance ToHttpApiData AccountIndex where
...
+        accpubkeywithixPublicKey
+        accpubkeywithixIndex
+
+instance Buildable [AccountPublicKeyWithIx] where
+    build = bprint listJson
+
+-- | A type modelling the request for a new 'EosWallet',
+-- on the mobile client or hardware wallet.
+data NewEosWallet = NewEosWallet
+    { neweoswalAccountsPublicKeysWithIxs :: ![AccountPublicKeyWithIx]
```
KtorZ: Remember that we are making an API for human to use :upside_down_face:  .. `accountPublicKeys` was fine, or simply, "accounts".

denisshevchenko: Agree. Fixed.

KtorZ (approved):

KtorZ:
@denisshevchenko hlint isn't happy with this one ^^^^

denisshevchenko:
@KtorZ Thanks, fixed.

akegalj (approved):
LGTM

account indexes are not used for anything in EOS wallet so far, am I right? What is a plan on when to use them?

EDIT: I see usage in this PR https://github.com/input-output-hk/cardano-wallet/pull/255/files

denisshevchenko:
> account indexes are not used for anything in EOS wallet so far, am I right?

@akegalj We're storing accounts for EOS-wallets in the same hierarchy, I mean `HdAccount`, and it contains `HdAccountBase` which includes `HdAccountId` which includes `HdAccountIx`. :-) So we _have_ to provide an index.

Issue #258, opened 2019-01-25
Support batch import of addresses through the API


KtorZ:
# Context

Exchanges typically use our wallet as a giant bucket of addresses, using a single account containing many addresses. Usually, they keep a mapping of 1 address per user, such that, they can advertise a "deposit address" to their users. 
This means that, when they create a new random address from the API, and show it to their users, this address should remain available and known until used. If they were to restart their wallet with an empty DB, they'll loose all unused address and there's no guarantee they'll ever generate those again.

# Decision

In order to allow exchange to migrate from versions prior to 1.4 to 1.4+ using the new data-layer, we need to provide them with an extra endpoint to give them the ability to import addresses into an existing wallet / account.

# Acceptance Criterias

1. The API `must` support an extra endpoint to perform batch import of addresses
1. The API handler `mustn't` fail if one or several addresses can't be imported (but rather, go through with the rest and report the failures back, if any).
1. The API handler `mustn't` override existing addresses in the DB
1. The API `may` disallow batch import during a restoration (unless we can correctly manage the fact that addresses may be imported prior to being discovered during restoration)
1. The API `may` define a hard-limit for the batch's size

---

# Development Plan

- [x] I intend to add an extra endpoint `POST /api/v1/wallets/{wid}/accounts/{accix}/addresses` to define batch import
- [x] I intend to respond with a number of success and, a list of addresses that has failed to import (if any)
- [x] I intend to implement the logic "naively", using existing handles to create one address and iterate over each address in the batch to add them one by one.
- [x] I intend to add integration tests to verify various scenarios of address import (used address, unused address, during restoration...)

# PR

| Number   | Base      |
| ---      | ---       |
| #259       | `develop` |
| #268 | `develop` |
| input-output-hk/cardano-sl#4041 | `release/2.0.2` |
| input-output-hk/cardano-sl#4040 | `develop` |
| #286 | `develop` |
| #308 | `develop` |
| #309 | `develop` |
| #319 | `develop` | 
| #320 | `develop` |
| input-output-hk/cardano-sl#4069 | `develop` |
| input-output-hk/cardano-sl#4070 | `release/2.0.2` |
| #327  | `develop` |
| input-output-hk/cardano-sl#4084 | `develop` |

# QA 

Note, this has also been backported to `develop` 

| Criteria | Coverage |
| ---      | ---      |
[...]

PR #259, opened 2019-01-25
Batch Import for Addresses

KtorZ:
<p align="right">#258</p>

# Overview

- [ ] I have added an extra endpoint to import batch of addresses (used or unused)
- [ ] I have defined a `BatchImportResult` type to aggregate imported results (we care about failures, but not much about successes)
- [ ] I have implemented the import in a rather naive fashion adding addresses one-by-one.  

# Comments

The approach taken to implement the import may be rather inefficient. However, this endpoint will likely be used only once per exchange and need to get shipped ASAP. So performance is our least concern here. If it later becomes _a thing_; we may consider doing batch updates and provide the necessary tools out of our acid-state layer. I thought `AccountUpdate` were doing batch updates for addresses (since they do have a `addrs` field); Though, it boils down to a quick-n-dirty `mapM_ createAddress accountUpdateAddrs` :disappointed: 

Nothing has been tested yet. Integration tests are coming in a second PR later and may likely fix a bunch of stuff in the implementation. Who knows. At least, this can serve as a skeleton / foundation.

image

### DIFF
--- a/src/Cardano/Wallet/API/V1/Addresses.hs
+++ b/src/Cardano/Wallet/API/V1/Addresses.hs
@@ -19,4 +19,8 @@ type API = Tags '["Addresses"] :>
       :<|> "addresses" :> Capture "address" Text
                        :> Summary "Returns interesting information about an address, if available and valid."
                        :> Get '[ValidJSON] (APIResponse WalletAddress)
+      :<|> "wallets" :> CaptureWalletId :> "accounts" :> CaptureAccountId :> "addresses"
+        :> Summary "Batch import existing addresses"
+        :> ReqBody '[ValidJSON] [WalAddress]
+        :> Post '[ValidJSON] (APIResponse (BatchImportResult WalAddress))
       )
--- a/src/Cardano/Wallet/API/V1/Handlers/Addresses.hs
+++ b/src/Cardano/Wallet/API/V1/Handlers/Addresses.hs
@@ -20,6 +20,7 @@ handlers :: PassiveWalletLayer IO -> ServerT Addresses.API Handler
 handlers w =  listAddresses w
          :<|> newAddress w
          :<|> getAddress w
+         :<|> importAddresses w
 
 listAddresses :: PassiveWalletLayer IO
               -> RequestParams -> Handler (APIResponse [WalletAddress])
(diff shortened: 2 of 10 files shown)


uroboros (approved):
LGTM - will we later extend this same API endpoint to add addrs for Hd sequential wallets? Or will that be another endpoint?

KtorZ:
@uroboros We likely won't. This is really just to facilitate migration of wallets from previous versions. The problem doesn't really occur (or is actually completely different) for sequential derivation because addresses are actually derived .. in sequence. So, as a user, one'll bump into the same addresses on a fresh new wallet. 

The big issue with the sequential derivation for exchanges at the moment is the number of unused addresses they need to have in supply. Having a `gap` of 100 may not be enough for big exchanges. But having a gap of 1000+ may be extremely inefficient to manipulate in the wallet backend. We are investigating at the moment how they deal with wallets in other blockchain that also follow a sequential derivation scheme.

deepfire:
@KtorZ, it would be really nice to get some test coverage for this.

Thinking about use cases an exchange will have (primarily collisions with existing ones?).

deepfire:
@KtorZ, another thing is that we'll need this in 1.4 as well.

KtorZ:
@deepfire  
> @KtorZ, it would be really nice to get some test coverage for this.

Ho, yes, the ticket isn't closed yet ^^" ... This is just the first step. 

> @KtorZ, another thing is that we'll need this in 1.4 as well.

You meant 1.5 right?

deepfire:
@KtorZ, no, 1.4, because customers who are currently still on 1.3, and whom we shouldn't keep waiting until 1.5..

src/Cardano/Wallet/API/V1/Addresses.hs
```diff
@@ -19,4 +19,8 @@ type API = Tags '["Addresses"] :>
       :<|> "addresses" :> Capture "address" Text
                        :> Summary "Returns interesting information about an address, if available and valid."
                        :> Get '[ValidJSON] (APIResponse WalletAddress)
+      :<|> "wallets" :> CaptureWalletId :> "accounts" :> CaptureAccountId :> "addresses"
+        :> Summary "Batch import existing addresses"
+        :> ReqBody '[ValidJSON] [WalAddress]
```
parsonsmatt, line 24: So this *is* going to perform poorly on large accounts. The "right" way to do this is with a file upload (ie [Multipart](https://haskell-servant.readthedocs.io/en/stable/cookbook/file-upload/FileUpload.html)) to a `tmp` directory we can stream from. Then we can process this in batches/chunks in constant memory.

How many addresses are exchanges wanting to backup and restore? We should find this out and determine if it's a realistic bottleneck -- 100k addresses is 160MB of memory just in list overhead, but 1k is only 1.6MB list overhead.

deepfire: @parsonsmatt, let's assume 50000 for now -- I'll ask our key client, but the answer is unlikely to come this week..

KtorZ:
@matt.parsons agreed, hence the warning in the documentation and the comments in the PR / corresponding ticket.
I went for the naive and fastest approach here because:

- We basically need this for Friday
- That's most likely something exchanges will use one time only when they move from 1.3 to 1.4
- This came up as an unexpected and unplanned requirement, so we deliver whatever we can in the allowed time frame.

Performances were the least concern here and if we intend to keep on supporting this feature, I 100% agree with what you just wrote.

Issue #239, opened 2019-01-21
Implement IsOurs prefiltering for Hd Sequential wallets


uroboros:
# Context

When the wallet applies blocks it prefilters for "our" addresses and in this way tracks the utxo and balance of a wallet. We need to enable tracking of EOS wallets

# Decision

# Acceptance Criterias

- [ ] Creation of E.O.S Wallets _should_ be tested and covered in CI.

---

# Development Plan

In the form of a TODO list, explain how the ticket is going to be tackled and how
you intend to proceed. 

- [x] Implement ISOurs prefiltering for Hd Sequential wallets https://github.com/input-output-hk/cardano-wallet/pull/238
- [x] Extend prefiltering to EOS wallets for ApplyBlock https://github.com/input-output-hk/cardano-wallet/pull/321
- [x] On wallet startup, recover EOS wallet AccountPKs and use track the EOS accounts during applyBlock https://github.com/input-output-hk/cardano-wallet/pull/321
- [x] Extend NewTx (new pending/foreign) to EO wallets https://github.com/input-output-hk/cardano-wallet/pull/340
- [ ] test EOS tracking

# QA 

We need to verify that EOS wallets are being tracked:

* create a new EOS wallet in Cardano-wallet
* send the new wallet some money (from another known wallet)
* verify that the utxo appears in the cardan-wallet as the wallet applies new blocks

# Retrospective

KtorZ:
@uroboros Hey, we are missing an estimate for that task. May you add one, according to what came up during your analysis of the issue?

KtorZ:
Also, if you're working on this, may you please move it to the `In Progress` column ?

PR #264, opened 2019-01-28
[BIP44-239] (WIP) Extend prefiltering to EOS wallets for ApplyBlock

uroboros:

<p align="right">#</p>

# Overview

- [x] I have extended `prefilterBlocks` in the BListener to also prefilter for EOS wallets (this means that ApplyBlock will prefilter for EOS wallets as well)
- [x] I have extended newPending/newForeign prefiltering of tx outputs to deal with Eos wallet addresses
- [x] I have implemented restoreEosWallet and refactored the HdRnd restoreWallet (to enable sharing of code)

# TODO

- [ ] reconstruct EOS wallets/address pools  from Acidstate on wallet startup
- [ ] at entry point for Restoration, dispatch between Eos/HdRnd restorations
- [ ] remove `isOursSkip` from IsOurs class (try out Matt's idea for this)
- [ ] split Eos/Rnd wallet code more clearly (e.g. in Wallets.hs / Restore.hs)
- [ ] improve naming of Wallets.hs create* functions
- [ ] extend restoreKnownWallet and other restore* to EOS, where applicable

# Comments

- [ ] this is still WIP because I've not wired up the reading of EOS wallets from AcidState

### DIFF
--- a/src/Cardano/Wallet/Kernel/BListener.hs
+++ b/src/Cardano/Wallet/Kernel/BListener.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LambdaCase   #-}
+{-# LANGUAGE ViewPatterns #-}
+
 -- | React to BListener events
 module Cardano.Wallet.Kernel.BListener (
     -- * Respond to block chain events
@@ -25,15 +27,19 @@ import qualified Formatting.Buildable
 
 import           Pos.Chain.Block (HeaderHash)
 import           Pos.Chain.Genesis (Config (..))
-import           Pos.Chain.Txp (TxId)
+import           Pos.Chain.Txp (TxId, TxIn)
+import           Pos.Core (Address)
 import           Pos.Core.Chrono (OldestFirst (..))
+import           Pos.Crypto (EncryptedSecretKey)
 import           Pos.DB.Block (getBlund)
 import           Pos.Util.Log (Severity (..))
 
-import           Cardano.Wallet.Kernel.DB.AcidState (ApplyBlock (..),
+import           Cardano.Wallet.Kernel.AddressPool (AddressPool)
+import           Cardano.Wallet.Kernel.DB.AcidState (ApplyBlock (..), DB,
                      ObservableRollbackUseInTestsOnly (..), SwitchToFork (..),
                      SwitchToForkInternalError (..))
 import           Cardano.Wallet.Kernel.DB.BlockContext
+import           Cardano.Wallet.Kernel.DB.HdRootId (HdRootId)
 import           Cardano.Wallet.Kernel.DB.HdWallet
 import           Cardano.Wallet.Kernel.DB.InDb (fromDb)
 import           Cardano.Wallet.Kernel.DB.Resolved (ResolvedBlock, rbContext,
@@ -61,24 +67,61 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
 
 type PrefilterResult = ((BlockContext, PrefilteredBlock), [TxMeta])
 
--- | Prefilter the block for each account. Returns 'Nothing' if the prefiltering
--- is irrelevant for this node, i.e. there are no user wallets stored, so we
--- can avoid doing work (i.e. writing into the acid-state DB log) by skipping
--- such block application.
+-- | Prefilter a list of resolved blocks for all accounts in all wallets.
+-- Since different types wallets have different prefiltering mechanisms, we need
+-- to do prefiltering seperately for each type of wallet and then combine the results.
 prefilterBlocks :: PassiveWallet
                 -> [ResolvedBlock]
                 -> IO (Maybe [PrefilterResult])
+-- Returns 'Nothing' if the prefiltering is irrelevant for this node, i.e.
+-- there are no user wallets stored, so we can avoid doing work
+-- (e.g. writing into the acid-state DB log) by skipping such block application).
+prefilterBlocks _ [] = return Nothing
 prefilterBlocks pw bs = do
-    res <- getWalletCredentials pw
-    isForeign <- fmap Pending.txIns . foreignPendingByAccount <$> getWalletSnapshot pw
-    case res of
-         [] -> return Nothing
-         xs -> fmap Just $ flip evalStateT xs $ forM bs $ \rb -> do
-            pb <- state (prefilterBlock isForeign rb)
-            metas <- state (resolvedToTxMetas rb) >>= \case
-                Left e  -> throwM e
-                Right x -> return x
-            return ((rb ^. rbContext, pb), metas)
+    db <- getWalletSnapshot pw
+    let foreigns = fmap Pending.txIns . foreignPendingByAccount $ db
+    hdRnds <- getHdRndWallets db
+    hdSeqs <- getHdSeqWallets db
+
+    listToMaybe' <$>
+        ((<>) <$> prefilter foreigns hdRnds <*> prefilter foreigns hdSeqs)
+    where
+        -- Prefilter blocks for any class of wallets for which prefiltering is
+        -- defined (that is, a type constrained by IsOurs)
+        prefilter
+          :: IsOurs s
+          => Map HdAccountId (Set TxIn)
+          -> s
+          -> IO [PrefilterResult]
+        prefilter _ (isOursSkip -> True) = return []
+        prefilter fs ws = do
+            flip evalStateT ws $ forM bs $ \rb -> do
+                pb <- state (prefilterBlock fs rb)
+                metas <- state (resolvedToTxMetas rb) >>= \case
+                   Left e  -> throwM e
+                   Right x -> return x
+                return ((rb ^. rbContext, pb), metas)
+
+        -- A variation on Map.listToMaybe
+        listToMaybe' :: [PrefilterResult] -> Maybe [PrefilterResult]
+        listToMaybe' l = if null l then Nothing else (Just l)
+
+        -- Get the data required for prefiltering of HdRnd wallets
+        getHdRndWallets
+            :: DB
+            -> IO (Map HdRootId EncryptedSecretKey)
+        getHdRndWallets db = getWalletCredentials db
+                    (pw ^. walletKeystore)
+                    (pw ^. walletProtocolMagic)
+                    (pw ^. walletLogMessage)
+
+        -- TODO @uroboros
+        -- Get the data required for prefiltering of HdSeq wallets
+        getHdSeqWallets
+            :: DB
+            -> IO (Map HdAccountId (AddressPool Address))
+        getHdSeqWallets _db
+            = return Map.empty
 
 data BackfillFailed
     = SuccessorChanged BlockContext (Maybe BlockContext)
--- a/src/Cardano/Wallet/Kernel/BListener.hs
+++ b/src/Cardano/Wallet/Kernel/BListener.hs
@@ -66,14 +66,16 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
 type PrefilterResult = ((BlockContext, PrefilteredBlock), [TxMeta])
 
 -- | Prefilter a list of resolved blocks for all accounts in all wallets.
--- Since different types wallets have different prefiltering mechanisms, we need
--- to do prefiltering seperately for each type of wallet and then combine the results.
-prefilterBlocks :: PassiveWallet
-                -> [ResolvedBlock]
-                -> IO (Maybe [PrefilterResult])
+-- Since different types wallets have different prefiltering mechanisms,
+-- we need to do prefiltering seperately for each type of wallet and then
+-- combine the results.
+--
 -- Returns 'Nothing' if the prefiltering is irrelevant for this node, i.e.
 -- there are no user wallets stored, so we can avoid doing work
 -- (e.g. writing into the acid-state DB log) by skipping such block application).
+prefilterBlocks :: PassiveWallet
+                -> [ResolvedBlock]
+                -> IO (Maybe [PrefilterResult])
 prefilterBlocks _ [] = return Nothing
 prefilterBlocks pw bs = do
     db <- getWalletSnapshot pw
--- a/src/Cardano/Wallet/Kernel/BListener.hs
+++ b/src/Cardano/Wallet/Kernel/BListener.hs
@@ -28,13 +28,11 @@ import qualified Formatting.Buildable
 import           Pos.Chain.Block (HeaderHash)
 import           Pos.Chain.Genesis (Config (..))
 import           Pos.Chain.Txp (TxId, TxIn)
-import           Pos.Core (Address)
 import           Pos.Core.Chrono (OldestFirst (..))
 import           Pos.Crypto (EncryptedSecretKey)
 import           Pos.DB.Block (getBlund)
 import           Pos.Util.Log (Severity (..))
 
-import           Cardano.Wallet.Kernel.AddressPool (AddressPool)
 import           Cardano.Wallet.Kernel.DB.AcidState (ApplyBlock (..), DB,
                      ObservableRollbackUseInTestsOnly (..), SwitchToFork (..),
                      SwitchToForkInternalError (..))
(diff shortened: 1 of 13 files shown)


src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -522,7 +524,11 @@ hdAccountRestorationState a = case a ^. hdAccountState of
 -------------------------------------------------------------------------------}
 
 class IsOurs s where
+    -- For the given state, check whether an address is "ours"
     isOurs :: Core.Address -> s -> (Maybe HdAddress, s)
+    -- Marks the state as empty and enables us to skip single or batch calls
+    -- to isOurs in the case of trivial state.
+    isOursSkip :: s -> Bool
```
KtorZ, line 542: I do not understand where this decision comes from :thinking: ? And I'd like us *not to* do or think about any optimizations until we have benchmarks and actual KPIs that allow us to:

- Justify the effort and complexity of developing an optimization
- Measure whether we are actually optimizing something 

First, let's make it work. Then, let's make it fast.

src/Cardano/Wallet/Kernel/Read.hs
```diff
@@ -35,18 +36,21 @@ getWalletSnapshot pw = query' (pw ^. wallets) Snapshot
...
-        aux walletId <$> Keystore.lookup nm walletId (pw ^. walletKeystore)
-    unless (null missing) $ (pw ^. walletLogMessage) Error (errMissing missing)
-    return creds
+        aux walletId <$> Keystore.lookup nm walletId ks
+    unless (null missing) $ logger Error (errMissing missing)
+    return (Map.fromList creds)
   where
     nm :: NetworkMagic
-    nm = makeNetworkMagic (pw ^. walletProtocolMagic)
+    nm = makeNetworkMagic pm
```
KtorZ, line 71: Ok. This is, I suppose, to avoid concurrency issues when accessing the snapshot multiple times throughout the life of a request call :thinking:  ?

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -531,8 +537,11 @@ class IsOurs s where
 -- | NOTE: We could modify the given state here to actually store decrypted
 -- addresses in a `Map` to trade a decryption against a map lookup for already
 -- decrypted addresses.
-instance IsOurs [(HdRootId, Core.EncryptedSecretKey)] where
-    isOurs addr s = (,s) $ foldl' (<|>) Nothing $ flip map s $ \(rootId, esk) -> do
+instance IsOurs (Map HdRootId Core.EncryptedSecretKey) where
```
KtorZ, line 551: Not sure to understand the rational here either :thinking: ? What's pushing for this change?

KtorZ (changes requested):

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: to clarify, this was an existing optimisation that I _preserved_ in the more general case (skipping an isOurs computation for each Address even when there are no credentials)

src/Cardano/Wallet/Kernel/Read.hs
uroboros: at the main call sites of getWalletCredentials, we already have the snapshot

uroboros: (we were doing a double getWalletSnapshot)

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: the IsOurs instance for EOS was using Map while IsOurs for ESK was using a list of tuples. Aligning these two made things more consistent (and semantically, the (Map RootId ESK) is more correct - it's a bag of wallets more than an (ordered) list

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
KtorZ: I'd rather not have it here; it adds unnecessary complexity and only save us from one 

```hs
(,s) $ foldl' (<|>) Nothing $ Map.mapWithKey s
```
when `s` is `mempty`. This is okay and don't need any optimization :+1:

uroboros: no I can't do it inside isOurs (which is being called for each address) - I want a function to tell me that there are no wallets - this is non-trivial because of the abstract `s` in IsOurs - hence this added function to say "whatever my type, there are no wallets to prefilter". This is required for the significant optimisation if simply skipping processing any blocks of there are no wallets.

uroboros: having said that, there may be a better way to achieve that with just type annotations. However we do it, we need to know empty from not empty for the given "prefiltering state"

KtorZ: I am not saying we should do anything inside the `isOurs` function. I am saying we should not do anything at all. This optimization buys us nothing; The case where there's no wallet is a true edge-case we do not intend to optimize. If one uses the wallet, that's for keeping track of wallets! Otherwise, one's free to start a standard node :+1:

uroboros: IsOurs acts on a class of wallet (EO/FO) - since it's quite likely that one has only one class of wallets, say HdSeq, the other class(es) need to be trivially ignored

uroboros: (I'm saying it remains a necessary optimisation)

KtorZ: I don't understand your point here. When we prefilter, we'll call the `isOurs` function for either some PublicKey + AddressPool, or for some HdRootId + SecretKey. Thus, we either have a E.O. wallet, or a F.O. one. There's no case where we'll call `isOurs` with... nothing to call it with in a first place. Otherwise, something should be fixed at the caller site, not here in this function.

We could even constraint the state to be a non empty list of credentials because conceptually, that's what it is.

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -61,24 +65,53 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
...
--- such block application.
+-- | Prefilter a list of resolved blocks for all accounts in all wallets.
+-- Since different types wallets have different prefiltering mechanisms, we need
+-- to do prefiltering seperately for each type of wallet and then combine the results.
 prefilterBlocks :: PassiveWallet
                 -> [ResolvedBlock]
                 -> IO (Maybe [PrefilterResult])
+-- Returns 'Nothing' if the prefiltering is irrelevant for this node, i.e.
+-- there are no user wallets stored, so we can avoid doing work
+-- (e.g. writing into the acid-state DB log) by skipping such block application).
```
parsonsmatt, line 75: This comment should be part of the Haddock comment over the function declaration -- we don't really have haddocks now, but it'd be good to not obstruct that as much as possible.

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -61,24 +65,53 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
...
+            flip evalStateT ws $ forM bs $ \rb -> do
+                pb <- state (prefilterBlock fs rb)
+                metas <- state (resolvedToTxMetas rb) >>= \case
+                   Left e  -> throwM e
+                   Right x -> return x
+                return ((rb ^. rbContext, pb), metas)
+
+        -- A variation on Map.listToMaybe
+        listToMaybe' :: [PrefilterResult] -> Maybe [PrefilterResult]
+        listToMaybe' l = if null l then Nothing else (Just l)
```
parsonsmatt, line 107: This is also `Data.List.NonEmpty.nonEmpty`, but that preserves the invariant that the list is non-empty.

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -522,7 +524,11 @@ hdAccountRestorationState a = case a ^. hdAccountState of
 -------------------------------------------------------------------------------}
 
 class IsOurs s where
+    -- For the given state, check whether an address is "ours"
```
parsonsmatt: We can tag these comments as Haddocsk with the `-- |` syntax, then they could show up whenever we render docs or with the `:doc` GHCi command.

src/Cardano/Wallet/Kernel/Pending.hs
```diff
@@ -1,3 +1,6 @@
+{-# LANGUAGE LambdaCase #-}
```
parsonsmatt, line 1: needs a `stylish-haskell` run

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
parsonsmatt: I don't believe this is a necessary *class* method -- we can replace it with direct calls in the implementations.

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -531,8 +537,11 @@ class IsOurs s where
 -- | NOTE: We could modify the given state here to actually store decrypted
 -- addresses in a `Map` to trade a decryption against a map lookup for already
 -- decrypted addresses.
-instance IsOurs [(HdRootId, Core.EncryptedSecretKey)] where
-    isOurs addr s = (,s) $ foldl' (<|>) Nothing $ flip map s $ \(rootId, esk) -> do
+instance IsOurs (Map HdRootId Core.EncryptedSecretKey) where
+    isOursSkip = Map.null
+
+    isOurs _ s@(isOursSkip -> True) = (Nothing, s)
```
parsonsmatt, line 554: eg, we use `isOursSkip` here -- but we could just use `Map.null` directly and not add a class member.

In the case of an empty `Map`, `flip Map.mapWithKey` is a noop, as is `foldl' (<|>) Nothing`. I think that this optimization isn't warranted.

src/Cardano/Wallet/Kernel/Pending.hs
```diff
@@ -106,15 +118,48 @@ newTx ActiveWallet{..} accountId tx partialMeta upd = do
...
+        -- If the accountId is part of an HdSeq wallet, then this selects
+        -- all HdAccountId's that have the same root of the given accountId.
+        thisPrefSeq
+            :: Map HdAccountId (AddressPool Address)
+            -> Map HdAccountId (AddressPool Address)
+        thisPrefSeq
+            = Map.filterWithKey (\a _ -> accountId ^. hdAccountIdParent == a ^. hdAccountIdParent)
+
+        -- filter the Tx outputs for "ours" in the prefilter context _s_
         allOurs
```
parsonsmatt, line 156: so this type sig was generalized to work with `IsOurs s`, so we can use it over two different kinds of `Map`. Would it possibly make sense to change the type class definition to the following?

```haskell
class IsOurs k v | k -> v, v -> k where
  isOurs :: Core.Addres -> Map k v -> (Maybe HdAddress, Map k v)
```

That allows us to define

```haskell
isOursSkip :: IsOurs k v => Map k v -> Bool
isOursSkip = Map.null
```

and recover the optimization without polluting the class with extraneous `Foldable`-like methods. The only instances are the ones defined in this PR, and both share a similar state.

parsonsmatt: Alternatively, `class IsOurs db a | db -> a, a -> db where isOurs :: Core.Address -> db a -> ...`, which lets us vary the database type, and we can then require a `Foldable db` instance to give us `null :: Foldable f => f a -> Bool`.

src/Cardano/Wallet/Kernel/Read.hs
```diff
@@ -56,3 +65,12 @@ getWalletCredentials pw = do
...
     errMissing = sformat ("Root key missing for " % listJson)
+
+-- TODO @uroboros
+-- Get HD Sequential wallet accounts along with the associated AddressPool
+-- required for prefiltering each account.
+getHdSeqWallets
+    :: DB
+    -> IO (Map HdAccountId (AddressPool Address))
+getHdSeqWallets _db
+    = return Map.empty
```
parsonsmatt: Is this going to be done as part of this PR?

parsonsmatt (changes requested):

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: I was driven to the need for the "is empty state" predicate because at the call sites I mostly want to call it for multiple state types, and then just merge the results (your suggestion below solves this nicely)

src/Cardano/Wallet/Kernel/BListener.hs
uroboros: 👍

src/Cardano/Wallet/Kernel/Pending.hs
uroboros: that's clever, I'll try that! The `IsOurs k v` is probably too closely tied to Map for our purposes, but `IsOurs db a` seems just what I need to get at that Foldable constraint.

src/Cardano/Wallet/Kernel/Read.hs
uroboros: yes, that's waiting for another PR to be merged first

src/Cardano/Wallet/Kernel/BListener.hs
denisshevchenko: Maybe this variant is better?

```
listToMaybe' :: [PrefilterResult] -> Maybe [PrefilterResult]
listToMaybe' [] = Nothing
listToMaybe' l  = Just l
```

src/Cardano/Wallet/Kernel/Pending.hs
```diff
@@ -106,15 +118,48 @@ newTx ActiveWallet{..} accountId tx partialMeta upd = do
     where
         (txOut :: [TxOut]) = NE.toList $ (_txOutputs . taTx $ tx)
 
-        -- | NOTE: we recognise addresses in the transaction outputs that belong to _all_ wallets,
-        --  not only for the wallet to which this transaction is being submitted
+        -- Prefiltering context for HdRnd and Eos wallet types
+        fullPref
```
denisshevchenko, line 123: IMHO it's not a good name. Why not `fullPrefilter`?

src/Cardano/Wallet/Kernel/Pending.hs
```diff
@@ -86,17 +91,24 @@ newTx :: forall e. ActiveWallet
...
-            -- process transaction on success
-            -- myCredentials should be a list with a single element.
-            let myCredentials = filter (\(hdRoot, _) -> accountId ^. hdAccountIdParent == hdRoot) allCredentials
-                ourOutputCoins = snd <$> allOurs myCredentials
+            let ourOutputCoins
+                    = snd <$> allOurs (thisPrefRnd hdRnds) <> allOurs (thisPrefSeq hdSeqs)
+                -- | NOTE: above, we prefilter coins that belong to _all_ accounts
+                -- which share the root of the given accountId.
+                -- Since we don't know the type of this wallet, we prefilter
+                -- it as either type (one of which will be trivial)
```
KtorZ, line 112: This is not _true_. We do know the `HdAccountId`, so we are one call away from knowing the corresponding `HdAccount` and its kind. That'll greatly simplify the code below.

src/Cardano/Wallet/Kernel/Restore.hs
```diff
@@ -109,29 +106,90 @@ restoreWallet
...
+    :: HD.IsOurs s
+    => Kernel.PassiveWallet
+    -> s
+    -> (Map HD.HdAccountId (Utxo,[HD.HdAddress])
+        -> IO (Either CreateHdRootError HD.HdRoot))
+    -> (Map HD.HdAccountId (Utxo, Utxo, [HD.HdAddress])
+        -> BlockContext
+        -> IO (Either CreateHdRootError HD.HdRoot))
+    -> IO (Either CreateHdRootError (HD.HdRoot, Coin))
+restoreHdWallet pw prefContext createWallet_ restoreWallet_
```
KtorZ, line 180: I didn't quite review any of the above _yet_ ^^^ , because I thought this PR and task was about prefiltering. I'd recommend we first figure out how to do prefiltering correctly and finalize the other bits before we move on to restoration...

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -332,6 +350,144 @@ createWalletHdRnd pw hasSpendingPassword defaultCardanoAddress name assuranceLev
...
+    -> a
+    -> Map HD.HdAccountId a
+    -> (PublicKey, HD.HdAccountIx)
+    -> Map HD.HdAccountBase a
+mkEosBase rootId gap def utxos (pk,ix)
+    = Map.singleton base val
+    where
+        accId = HD.HdAccountId rootId ix
+        base = HD.HdAccountBaseEO accId pk gap
+        val = maybe def id (Map.lookup accId utxos)
```
KtorZ, line 473: Same here. Let's move with incremental steps. And test the things we add as we add them.

KtorZ (changes requested):

uroboros:
Will introduce this work in a series of PRs, closing this now

Issue #190, opened 2019-01-02
Improve instructions (and testing) about multi-output transactions


KtorZ:
# Context

The new coin selection algorithm has some limitations that aren't clearly stated from the documentation (nor in the specification actually). And a few points about multi-outputs transactions are still unresolved. 
The coin selection can't split a UTxO across multiple outputs. In order to make a multi-outputs transaction, there should be several available UTxO. 

# Decision

When trying to make a transaction with more outputs than available UTxO, the API fails with `CoinSelHardErrUtxoExhausted` which isn't really helpful from a user point of view. Ideally, we would improve the error message and suggest a possible fix or, way of approaching the error (by suggesting to have a look at the utxo fragmentation through the utxo statistics endpoint).

Also, we want to add integration tests illustrating the scenario above (failing and successful one) and, with this, extend the API documentation to explain how to deal with multi-outputs transactions and what are the limitations.

# Acceptance Criterias

1. The API _must_ contain instructions about fragmenting a wallet UTxO 
1. The API _must_ contain instructions about creating multi-outputs transactions, suggesting an ideal and maximum level of fragmentation for exchanges.
1. Some Integration tests _must_ illustrate the following scenarios:
    - A multi-output transaction fails because there's not enough UTxO to cover for all outputs
    - A multi-output transaction succeeds 
1. The API _should_ return a more meaningful error when the case for (failing) multi-output is encountered, suggesting the user to have a look at `/api/v1/wallets/{walletId}/statistics`

---

# Development Plan

- [x] Realize what is required above
- [x] add boundary tests to multioutput transaction suite

# PR

| Number   | Base      |
| ---      | ---       |
| #272       | `develop` |
| #288       | `develop` |
| #289       | `develop` |
| #290        | `develop` |
| input-output-hk/cardano-sl#4058 | `develop` |
| #298         | `develop` |

# QA 

When the number of transaction outputs is greater than the number of utxo new error is triggered two tests in Transaction section of integration tests are added that cover the situation.

When the number of transaction outputs is greater than the number of utxo new error is triggered . Previously it the following error was triggered:
```
expectWalletError (NotEnoughMoney (ErrAvailableBalanceIsInsufficient 0)
```
which is a bit misleading.
Now, the following error is expected:
```
UtxoNotEnoughFragmented (Client.ErrUtxoNotEnoughFragmented 1 Client.msgUtxoNotEnoughFragmented)
```
[...]

PR #272, opened 2019-01-29
Add error support when not enough utxo is met upon multioutput …

paweljakubas:
#190 

<p align="right">#</p>

# Overview

- [x] Integration test illustrating the following scenarios were added:
- [x] A multi-output transaction fails because there's not enough UTxO to cover for all outputs
- [x] A multi-output transaction succeeds
- [x] The API returns a more meaningful error when the case for (failing) multi-output is encountered, suggesting the user to have a look at /api/v1/wallets/{walletId}/statistics/utxo

# Comments

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -125,6 +125,8 @@ module Cardano.Wallet.API.V1.Types (
   -- * Wallet Errors
   , WalletError(..)
   , ErrNotEnoughMoney(..)
+  , ErrUtxoNotEnoughFragmented(..)
+  , msgUtxoNotEnoughFragmented
   , toServantError
   , toHttpErrorStatus
   , module Cardano.Wallet.Types.UtxoStatistics
@@ -2410,6 +2412,23 @@ instance FromJSON ErrNotEnoughMoney where
             ErrAvailableBalanceIsInsufficient <$> (o .: "availableBalance")
 
 
+data ErrUtxoNotEnoughFragmented = ErrUtxoNotEnoughFragmented {
+      theMissingUtxos :: !Int
+    , theHelp         :: !Text
+    } deriving (Eq, Generic, Show)
+
+
+msgUtxoNotEnoughFragmented :: Text
+msgUtxoNotEnoughFragmented = "Utxo is not enough fragmented to handle the number of outputs of this transaction. Query /api/v1/wallets/{walletId}/statistics/utxos endpoint for more information"
+
+deriveJSON Aeson.defaultOptions ''ErrUtxoNotEnoughFragmented
+
+instance Buildable ErrUtxoNotEnoughFragmented where
+    build (ErrUtxoNotEnoughFragmented missingUtxos _ ) =
+        bprint ("Missing "%build%" utxo(s) to accommodate all outputs of the transaction") missingUtxos
+
+
+
 -- | Type representing any error which might be thrown by wallet.
 --
 -- Errors are represented in JSON in the JSend format (<https://labs.omniti.com/labs/jsend>):
@@ -2474,6 +2493,9 @@ data WalletError =
     | RequestThrottled !Word64
     -- ^ The request has been throttled. The 'Word64' is a count of microseconds
     -- until the user should retry.
+    | UtxoNotEnoughFragmented !ErrUtxoNotEnoughFragmented
+    -- ^ available Utxo is not enough fragmented, ie., there is more outputs of transaction than
+    -- utxos
     deriving (Generic, Show, Eq)
 
 deriveGeneric ''WalletError
@@ -2516,6 +2538,9 @@ instance Arbitrary WalletError where
         , NodeIsStillSyncing <$> arbitrary
         , CannotCreateAddress <$> arbitraryText
         , RequestThrottled <$> arbitrary
+        , UtxoNotEnoughFragmented <$> Gen.oneof
+          [ ErrUtxoNotEnoughFragmented <$> Gen.choose (1, 10) <*> arbitrary
+          ]
         ]
       where
         arbitraryText :: Gen Text
@@ -2572,7 +2597,8 @@ instance Buildable WalletError where
             bprint "Cannot create derivation path for new address, for external wallet."
         RequestThrottled _ ->
             bprint "You've made too many requests too soon, and this one was throttled."
-
+        UtxoNotEnoughFragmented x ->
+            bprint build x
 
 -- | Convert wallet errors to Servant errors
 instance ToServantError WalletError where
@@ -2615,6 +2641,9 @@ instance ToServantError WalletError where
             err500
         RequestThrottled{} ->
             err400 { errHTTPCode = 429 }
+        UtxoNotEnoughFragmented{} ->
+            err403
+
 
 -- | Declare the key used to wrap the diagnostic payload, if any
 instance HasDiagnostic WalletError where
@@ -2657,3 +2686,5 @@ instance HasDiagnostic WalletError where
             "msg"
         RequestThrottled{} ->
             "microsecondsUntilRetry"
+        UtxoNotEnoughFragmented{} ->
+            "details"
--- a/src/Cardano/Wallet/Kernel/Transactions.hs
+++ b/src/Cardano/Wallet/Kernel/Transactions.hs
@@ -10,6 +10,7 @@ module Cardano.Wallet.Kernel.Transactions (
     , PaymentError(..)
     , EstimateFeesError(..)
     , RedeemAdaError(..)
+    , NumberOfMissingUtxos(..)
     , cardanoFee
     , mkStdTx
     , prepareUnsignedTxWithSources
@@ -87,6 +88,17 @@ import           UTxO.Util (shuffleNE)
   Generating payments and estimating fees
 -------------------------------------------------------------------------------}
 
+data NumberOfMissingUtxos = NumberOfMissingUtxos Int
+
+instance Buildable NumberOfMissingUtxos where
+    build (NumberOfMissingUtxos number) =
+        bprint ("NumberOfMissingUtxos " % build) number
+
+instance Arbitrary NumberOfMissingUtxos where
+    arbitrary = oneof [ NumberOfMissingUtxos <$> arbitrary
+                      ]
+
+
 data NewTransactionError =
     NewTransactionUnknownAccount UnknownHdAccount
   | NewTransactionUnknownAddress UnknownHdAddress
@@ -94,6 +106,7 @@ data NewTransactionError =
   | NewTransactionErrorCreateAddressFailed Kernel.CreateAddressError
   | NewTransactionErrorSignTxFailed SignTransactionError
   | NewTransactionInvalidTxIn
+  | NewTransactionNotEnoughUtxoFragmentation NumberOfMissingUtxos
 
 instance Buildable NewTransactionError where
     build (NewTransactionUnknownAccount err) =
@@ -108,6 +121,9 @@ instance Buildable NewTransactionError where
         bprint ("NewTransactionErrorSignTxFailed " % build) err
     build NewTransactionInvalidTxIn =
         bprint "NewTransactionInvalidTxIn"
+    build (NewTransactionNotEnoughUtxoFragmentation err) =
+        bprint ("NewTransactionNotEnoughUtxoFragmentation" % build) err
+
 
 instance Arbitrary NewTransactionError where
     arbitrary = oneof [
@@ -119,6 +135,7 @@ instance Arbitrary NewTransactionError where
       , NewTransactionErrorCreateAddressFailed <$> arbitrary
       , NewTransactionErrorSignTxFailed <$> arbitrary
       , pure NewTransactionInvalidTxIn
+      , NewTransactionNotEnoughUtxoFragmentation <$> arbitrary
       ]
 
 data PaymentError = PaymentNewTransactionError NewTransactionError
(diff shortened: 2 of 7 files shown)


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -2409,6 +2410,28 @@ instance FromJSON ErrNotEnoughMoney where
...
+    build = \case
+        ErrUtxoNotEnoughFragmentedBy _ ->
+             bprint "Not enough utxos to accomodate all outputs of the transaction"
+
+instance ToJSON ErrUtxoNotEnoughFragmented where
+    toJSON = \case
+        (ErrUtxoNotEnoughFragmentedBy missingUtxos) -> object
+            [ "msg"          .= String "Utxo is not enough fragmented to handle the number of outputs of this transaction. see /api/v1/wallets/{walletId}/statistics/utxos for more information"
+            , "missingUtxos" .= missingUtxos
+            ]
```
KtorZ: API errors are kinda... complicated at the moment but we do want to derive most things generically. I'd suggest to have:

```hs
data ErrUtxoNotEnoughFragmented = ErrUtxoNotEnoughFragmented
    { missingUtxos :: Int
    , msg :: Text
    }
```

And, provide a corresponding message as text in the `ReifyError`. This way, we can simply derive the ToJSON and FromJSON instance generically and don't need this manual wiring.

Then, make `details` the `diagnosticKey` such that it will generate a corresponding JSEnd error.

src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -216,6 +233,10 @@ newUnsignedTransaction ActiveWallet{..} options accountId payees = runExceptT $
     availableUtxo <- withExceptT NewTransactionUnknownAccount $ exceptT $
                        currentAvailableUtxo snapshot accountId
 
+
+    _ <- withExceptT NewTransactionNotEnoughUtxoFragmentation $ exceptT $
+           determineCorrectUtxoFragmentation payees availableUtxo
```
KtorZ: I'll suggest to change the signature of `determineCorrectUxoFragmentation` to `checkUtxoFragmentation :: NonEmpty (Address, Coin) -> Utxo -> Either NumberOfMissingUtxos ()` 

There's no need to return something here and the check is really just about stopping the computation or, moving forward.

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -4,12 +4,116 @@ module Test.Integration.Scenario.Transactions
...
+            & initialCoins .~ [500000, 500000]
+
+
+        randomBackPhrase1 <- liftIO $ generate arbitrary :: Scenario Context  IO Client.BackupPhrase
+        wallet1 <- successfulRequest $ Client.postWallet $- NewWallet
+                      randomBackPhrase1
+                      noSpendingPassword
+                      NormalAssurance
+                      "NewWallet1"
+                      CreateWallet
```
KtorZ: Please use the fixture for that. Fixture is used to create an initial wallet without the hassle of generating all the bits yourself. Wallets come default accounts `defaultAccountId` that you can use in the scenario. So basically, the lines above can just be replaced with:

```hs
fixture01 <- setup $ defaultSetup & name .~ "newWallet1"
```

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -4,12 +4,116 @@ module Test.Integration.Scenario.Transactions
...
+
+        randomBackPhrase2 <- liftIO $ generate arbitrary :: Scenario Context  IO Client.BackupPhrase
+        wallet2 <- successfulRequest $ Client.postWallet $- NewWallet
+                      randomBackPhrase2
+                      noSpendingPassword
+                      NormalAssurance
+                      "NewWallet2"
+                      CreateWallet
+        account2 <- successfulRequest $ Client.getAccounts
+                    $- wallet2 ^. walletId
```
KtorZ: Same as above :+1:

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -4,12 +4,116 @@ module Test.Integration.Scenario.Transactions
...
+        let recepientWalAddresses = map (\acc -> acc ^. walAddresses)
+                                    $ concat
+                                    $ map (\acc -> acc ^. addresses)
+                                    $ account1 ++ account2
+
+        let customizedDistribution = fromList
+                $ zipWith
+                Client.PaymentDistribution
+                recepientWalAddresses
+                (map (\coin -> Client.WalletCoin $ mkCoin coin) [10,11])
```
KtorZ: We really want to keep the complexity and the internal logic outside of the scenario. The `DSL` already offers:

```hs
defaultDistribution
    :: HasType (NonEmpty Address) s
    => Word64
    -> s
    -> NonEmpty PaymentDistribution
```

which isn't enough in your case, so I'd suggest to add a new:

```hs
customDistribution
    :: NonEmpty (Address, Word64)
    -> NonEmpty PaymentDistribution
```

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -4,12 +4,116 @@ module Test.Integration.Scenario.Transactions
...
+
+        response <- request $ Client.postTransaction $- Payment
+            (defaultSource fixture)
+            customizedDistribution
+            defaultGroupingPolicy
+            noSpendingPassword
+
+        verify response
+            [ expectTxInHistoryOf (fixture ^. wallet)
+            , expectTxStatusEventually [Creating, Applying, InNewestBlocks, Persisted]
```
KtorZ: I'd rather wait for `InNewestBlocks` here, to make sure the transaction is actually indeed accepted by the nodes.

KtorZ (changes requested):

src/Cardano/Wallet/API/V1/Types.hs
paweljakubas: done

src/Cardano/Wallet/Kernel/Transactions.hs
paweljakubas: done

test/integration/Test/Integration/Scenario/Transactions.hs
paweljakubas: applied. now it is much easier to read and all imp delegated to DSL

test/integration/Test/Integration/Scenario/Transactions.hs
paweljakubas: :+1:

test/integration/Test/Integration/Scenario/Transactions.hs
paweljakubas: done

test/integration/Test/Integration/Scenario/Transactions.hs
paweljakubas: indeed -> done

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -2410,6 +2412,26 @@ instance FromJSON ErrNotEnoughMoney where
...
+msgUtxoNotEnoughFragmented = "Utxo is not enough fragmented to handle the number of outputs of this transaction. Query /api/v1/wallets/{walletId}/statistics/utxos endpoint for more information"
+
+deriveJSON Aeson.defaultOptions ''ErrUtxoNotEnoughFragmented
+
+instance Buildable ErrUtxoNotEnoughFragmented where
+    build (ErrUtxoNotEnoughFragmented missingUtxos _ ) =
+        if missingUtxos == 1 then
+            bprint ("Missing "%build%" utxo to accommodate all outputs of the transaction") missingUtxos
+        else
+            bprint ("Missing "%build%" utxos to accommodate all outputs of the transaction") missingUtxos
```
KtorZ: Nice effort :smile: 
Pro tip: an `(s)` works marvelously for this kind of things:

```hs
             bprint ("Missing "%build%" utxo(s) to accommodate all outputs of the transaction") missingUtxos 
```

And, in general, as a backend, the accuracy of your grammar on edge-cases is really your last concern ^^"

test/unit/Servant/JsendCompliance.hs
```diff
@@ -20,7 +20,7 @@ import           Servant.Swagger.Internal.Test (props)
 import           Servant.Swagger.Internal.TypeLevel (AddBodyType, AppendList,
                      Every, Nub, TMap)
 import           Test.QuickCheck (Arbitrary (..), Property, counterexample,
-                     (.&&.), (===))
+                     (===))
```
KtorZ, line 23: Minor remark: when you fix something like this (which has nothing to do with our PR per se), make it a separate commit. Such that, the history is clearer to understand and, commits can be cherry-picked / reverted easily when they need to. 

Thanks for fixing the warning though :+1:

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -49,7 +104,7 @@ spec = do
             noSpendingPassword
 
         verify response
-            [ expectWalletError (NotEnoughMoney (ErrAvailableBalanceIsInsufficient 0))
+            [ expectWalletError (UtxoNotEnoughFragmented (Client.ErrUtxoNotEnoughFragmented 1 Client.msgUtxoNotEnoughFragmented))
```
KtorZ, line 107: :+1: Note that, this is a small "breaking" change and we should notify the frontend folks about this!

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -2657,3 +2689,5 @@ instance HasDiagnostic WalletError where
             "msg"
         RequestThrottled{} ->
             "microsecondsUntilRetry"
+        UtxoNotEnoughFragmented{} ->
+            "msg"
```
KtorZ: Note, by using "msg" as a diagnostic key, we'll end up with an error like:

```json
{    
    "status": "error", 
    "diagnostic": {        
        "msg": {            
            "msg": "...",            
            "missingUtxos": 14        
        }    
    },    
    "message": "UtxoNotEnoughFragmented"
}
```

So that's quite a lot of 'msg / messages' ... May I recommand to change the diagnostic key to `details` and maybe also, to change the `theMsg` by `theHelp` in the `ErrUtxoNotEnoughFragmented` type ? This will generate something more like:

```json
{    
    "status": "error", 
    "diagnostic": {        
        "details": {            
            "help": "...",            
            "missingUtxos": 14        
        }    
    },    
    "message": "UtxoNotEnoughFragmented"
}
```

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -265,6 +269,18 @@ defaultDistribution
 defaultDistribution c s = pure $
     PaymentDistribution (WalAddress $ head $ s ^. typed) (WalletCoin $ mkCoin c)
 
+customDistribution
+    :: [Account]
+    -> [Word64]
+    -> NonEmpty PaymentDistribution
+customDistribution accounts coins =
+    let recepientWalAddresses = map (\acc -> acc ^. walAddresses)
+                                $ concatMap (\acc -> acc ^. addresses)
```
KtorZ: NOTE: `\x -> x ^. myLens === view myLens`, favor functions over operator when passing functions around :+1:

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -265,6 +269,18 @@ defaultDistribution
...
+                                $ accounts
+    in fromList $
+       zipWith PaymentDistribution recepientWalAddresses (map (\coin -> WalletCoin $ mkCoin coin) coins)
```
KtorZ: This may lead to unexpected results if:

- The `accounts` and `coins` list don't have the same number of elements
- The `accounts` have more than one address 

I'd recommend to either check that as invariant and fail hard if the invariant aren't met, or (better IMO), require the caller to provide the a `(Account, Word64)]` himself and only take the first address of each account -- failing if there's none maybe. Or asking directly for `[(Address, Word64)]`

test/unit/Servant/JsendCompliance.hs
KtorZ: Also, this happens because `-WError` isn't on in `cardano-wallet.cabal` for the unit tests. So I'd rather turn it on as well while we're at it :)

KtorZ (changes requested):

src/Cardano/Wallet/API/V1/Types.hs
paweljakubas: yeah, a little bit overstretched here :-) corrected

test/unit/Servant/JsendCompliance.hs
paweljakubas: reverted the change, and included in another commit (together with switching on of -Werror flag for unit tests)

test/integration/Test/Integration/Scenario/Transactions.hs
paweljakubas: Will inform Nicola team about the change

src/Cardano/Wallet/API/V1/Types.hs
paweljakubas: done

test/integration/Test/Integration/Framework/DSL.hs
paweljakubas: yes, used `view` instead

test/integration/Test/Integration/Framework/DSL.hs
paweljakubas: done

KtorZ (approved):

KtorZ:
Nice work @paweljakubas! Thanks. Don't forget to put your ticket in QA after giving a few pointers to @piotr-iohk!

PR #273, opened 2019-01-29
getAddressPool function.

denisshevchenko:
<p align="right">#231</p>

# Overview

- [x] I have added `getAddressPool` function.

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
+++ b/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
@@ -97,6 +97,8 @@ module Cardano.Wallet.Kernel.DB.HdWallet (
   , assumeHdAccountExists
     -- * IsOurs
   , IsOurs(..)
+    -- Address pool
+  , mkAddressPool
   ) where
 
 import           Universum hiding ((:|))
@@ -120,7 +122,7 @@ import qualified Pos.Crypto as Core
 
 import           Cardano.Wallet.API.V1.Types (WalAddress (..))
 import           Cardano.Wallet.Kernel.AddressPool (AddressPool,
-                     lookupAddressPool)
+                     emptyAddressPool, lookupAddressPool)
 import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 import           Cardano.Wallet.Kernel.DB.BlockContext
 import           Cardano.Wallet.Kernel.DB.HdRootId (HdRootId)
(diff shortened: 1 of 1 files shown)


uroboros (approved):
LGTM

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -549,6 +553,23 @@ decryptHdLvl2DerivationPath hdPass addr = do
...
+{-------------------------------------------------------------------------------
+  create AddressPool for account
+-------------------------------------------------------------------------------}
+
+getAddressPool
+    :: (Core.PublicKey -> Core.Address)
+    -> HdAccount
+    -> Maybe (AddressPool Core.Address)
+getAddressPool _         (HdAccount (HdAccountBaseFO _) _ _ _) = Nothing
+getAddressPool mkAddress (HdAccount (HdAccountBaseEO _ accPK gap) _ _ _) =
```
KtorZ: We are loosing the benefits of having two branches for `HdAccount` if we start doing things like this. The caller should be pattern matching here and only creates an address pool when it makes sense.

KtorZ (changes requested):
Not sure what you're trying to achieve here :thinking: ?

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
denisshevchenko: Hm... We can create `emptyAddressPool` _only_ for `HdAccount` with EO-branch, right?

KtorZ:
So in the end, what was the need for this function again :thinking: ?

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: @KtorZ the function has a confusing name - it is meant for create a fresh pool for a new account (with the first gap of addresses)

denisshevchenko: I can rename it. For example, `createAddressPool` or `createEmptyAddressPool`.

uroboros: I think stay away from "empty" (it's not) - preferable something like initAddressPool, but that name is already taken. Maybe we should rename that rather? (mkAddressPool?)

KtorZ: Though the problem is more than the name here. That function _shouldn't_ take a full `HdAccount` because it has nothing to do with F.O. accounts :+1:

denisshevchenko: We assume that this function will be used _only_ with EO-accounts. But if not - it will return `Nothing`.

denisshevchenko: Another way - pass only things we really need.

denisshevchenko: I mean 

```
getAddressPool
    :: (Core.PublicKey -> Core.Address)
    -> Core.PublicKey
    -> AddressPoolGap
    -> AddressPool Core.Address
```

KtorZ: > Another way - pass only things we really need.

YES ! Don't assume things. Don't trust that "we" or even you will use the function in a "correct way", make it so! If there's only one way to use a function, and that's the good way, then you don't have to assume anything.

denisshevchenko:
> what was the need for this function again thinking ?

@uroboros ? :-)

uroboros:
@denisshevchenko when we create EOS wallets, we should use that function to compute the first pool of  HdAddreses, which we should persist on create. When I do prefiltering, I'll read from Acidstate and reconstruct pools using the "create existing pool" function

uroboros:
@denisshevchenko having said that... in create EOS wallets, when we create the pool of addrs for each new EOS account, we only need the `nextAddresses` function, so we _could_ lose "what used to be called emptyAddrPool". We need the other one though

KtorZ (approved):

PR #274, opened 2019-01-29
WALLETS_UPDATE_PASS tests

piotr-iohk:
<p align="right">#</p>

# Overview

- [x] I have added tests for `WALLETS_UPDATE_PASS` capability from https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk

# Comments

### DIFF
--- a/test/integration/Test/Integration/Scenario/Wallets.hs
+++ b/test/integration/Test/Integration/Scenario/Wallets.hs
@@ -517,129 +517,38 @@ spec = do
             , expectFieldEqual assuranceLevel StrictAssurance
             ]
 
-    describe "WALLETS_UPDATE_PASS_01,04,05,06,09 - Updating password to empty and non-empty" $ do
-
-        let matrix =
-                [ ( "non-empty old and new password"
-                  , "old raw password"
-                  , "new raw password" :: Text
-                  , [ expectFieldEqual hasSpendingPassword True ]
-                  )
-                 ,
-                  ( "old pass empty, new pass non-empty"
-                  , ""
-                  , "new raw password" :: Text
-                  , [ expectFieldEqual hasSpendingPassword True ]
-                  )
-                ,
-                  ( "old pass non-empty, new pass empty"
-                  , "old raw password"
-                  , "" :: Text
-                  , [ expectFieldEqual hasSpendingPassword False ]
-                  )
-                ,
-                  ( "empty old and new password"
-                  , ""
-                  , "" :: Text
-                  , [ expectFieldEqual hasSpendingPassword False ]
-                  )
-                ]
-
-        describe "Restored wallet" $ do
-
-            forM_ matrix $ \(title, oldPass, newPass, expectations) -> scenario title $ do
-                -- 1. Create and delete wallet
-                fixture <- setup $ defaultSetup
-                    & rawPassword .~ oldPass
-                successfulRequest $ Client.deleteWallet
-                    $- (fixture ^. wallet . walletId)
-
-                -- 2. Restore deleted wallet
-                let restoredPass = mkPassword (RawPassword newPass)
-                restoreResp <- request $ Client.postWallet $- NewWallet
-                    (fixture ^. backupPhrase)
-                    (Just $ restoredPass)
-                    StrictAssurance
-                    "My Wallet"
-                    RestoreWallet
-                verify restoreResp
-                    [ expectSuccess
-                    ]
-
-                -- 2. Update password
-                let latestUpdateTime = fixture ^. wallet ^. spendingPasswordLastUpdate
-                let fullExpectations = (expectFieldDiffer spendingPasswordLastUpdate latestUpdateTime) : expectations
-                updatePasswordResp <- request $ Client.updateWalletPassword
-                    $- fixture ^. wallet . walletId
-                    $- PasswordUpdate restoredPass (mkPassword (RawPassword newPass))
-                verify updatePasswordResp fullExpectations
-
-                -- 3. Check GET /api/v1/wallets/{walletId} presists the update
-                getWalletResp <- request $ Client.getWallet
-                    $- (fixture ^. wallet . walletId)
-                verify getWalletResp fullExpectations
-
-        describe "Created wallet" $ do
-
-            forM_ matrix $ \(title, oldPass, newPass, expectations) -> scenario title $ do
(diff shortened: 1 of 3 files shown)


test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -517,40 +517,431 @@ spec = do
...
+                let restoredPass = mkPassword (RawPassword newPass)
+                restoreResp <- request $ Client.postWallet $- NewWallet
+                    (fixture ^. backupPhrase)
+                    (Just $ restoredPass)
+                    StrictAssurance
+                    "My Wallet"
+                    RestoreWallet
+                verify restoreResp
+                    [ expectSuccess
+                    ]
```
KtorZ: What's the need for restoring the wallet here (and deleting it :thinking: ?)

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -517,40 +517,431 @@ spec = do
...
+                newAddrResp <- request $ Client.postAddress $- NewAddress
+                    (Just $ newPass)
+                    defaultAccountId
+                    walId
+                verify newAddrResp
+                    [ expectAddressInIndexOf
+                    ]
+
+                -- 4. Attempt to create new address with old passowrd
+                let errMsg = "CreateAddressHdRndGenerationFailed HdAccountId { parent " <> (fromWalletId walId) <> ", ix     HdAccountIx " <> (show (Client.getAccIndex defaultAccountId)) <> "}"
```
KtorZ: Erf. That's a shame it has to be this way. I'd rather have the `expectWalletError` do some sort of "resemblance" equality. Having to reconstruct the error message looks really ad-hoc here.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -517,40 +517,431 @@ spec = do
...
+                newAddrAttemptResp <- request $ Client.postAddress $- NewAddress
+                    (Just $ fixture ^. spendingPassword)
+                    defaultAccountId
+                    walId
+                verify newAddrAttemptResp
+                    [ expectWalletError (CannotCreateAddress errMsg)
+                    ]
```
KtorZ, line 571: Isn't this whole scenario a bit redundant with the one above :thinking: ?

KtorZ (changes requested):

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: Indeed. First I had it in one scenario, like.
1. Update password
2. Create address with new password
3. Attempt to create address with old password and expect error

But then I decided to make them as separate scenarios, but forgot to remove no 3 from the first scenario. Will remove it.

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: It is:
1. Creating a wallet
2. Deleting a wallet
3. Restoring a wallet
4. Updating password on restored wallet

It seemed like a scenario closer to _real life_ (or am I getting it wrong somehow?)

piotr-iohk: @KtorZ - a question on a related note: Is it possible somehow to achieve in the integration test: _send update password request on a wallet which is actually in the process of restoring/syncing._ 

Currently it seems to be _Synced_ right away...

KtorZ: That's a delicate things to test. And actually, not something we can easily instrument in integration tests against a local cluster. We do keep the last 2160 blocks in memory, so restoring on a blockchain that is shorter than that will be _instantaneous_; there's no historical data to recover since we already have it all. 

This is also something I'd like to instrument and test but I don't see this happening in the local cluster, unless we do start the cluster on a "pre-constructed" DB dump. This would actually be nice.

KtorZ: > It seemed like a scenario closer to real life (or am I getting it wrong somehow?)

Hmmm... I think a real scenario would be more like "restoring with a given a password" or "restoring and  updating password". The creation + deletion has most likely happened on a different node. I don't really see any "real life" scenario where a user would perform those 4 ops in sequence to be honest ^^"

I do get your point though,.

piotr-iohk: >I don't really see any "real life" scenario where a user would perform those 4 ops in sequence to be honest ^^"

yes, that's true. 

I was rather after a scenario of something like: 
1. I have had a wallet, it's on a blockchain; 
2. I am restoring it, 
3. after I had it restored I'm updating password

In other words, wanted to restore the wallet that was already existing before (in that sense it seemed closer to "real life"). :)

piotr-iohk: @KtorZ, anyway if you think that _creating > deleting > restoring_ does not add additional value to the test I will remove _creating > deleting_ :+1: 

Also strange thing on CI. All tests are passing but the job itself ends with an error -> https://travis-ci.org/input-output-hk/cardano-wallet/jobs/486065904 :/

KtorZ:
> Also strange thing on CI. All tests are passing but the job itself ends with an error

Yeah, I noticed this on develop the other day. That's actually because we are outputting too many logs and travis kills the job in returns :|...
I think it's now time to remove this debugging 'cat' line on the CI config.

KtorZ:
@piotr-iohk Sorry for not being much responsive here. A bit busy with the 1,5 release. I'll have a look at that tomorrow (unless @parsonsmatt wants to have one as well :pray: )

KtorZ (approved):
Reviewed locally & made a couple of small style adjustments (mostly removing some unnecessary comments and newlines) while reading through it.

piotr-iohk:
@KtorZ, I have found the culprit. It was a wrong `spendingPassword` in the failing test (I was using destination wallet's spending password instead of source wallet's when making transaction).
Within this PR I also fixed the `setupWallet` to set the password for a fixture so the password started to matter. On develop fixture's `spendingPassword` was always `mempty` hence this test was passing.

I suppose it can be merged now. :+1:

PR #276, opened 2019-01-30
Remove duplicate default extensions.

jonathanknowles:
This change removes a few duplicate entries from `default-extensions` sections of the `cardano-wallet.cabal` file.

### DIFF
--- a/cardano-wallet.cabal
+++ b/cardano-wallet.cabal
@@ -27,12 +27,10 @@ library
       OverloadedStrings
       RecordWildCards
       ScopedTypeVariables
-      ScopedTypeVariables
       TemplateHaskell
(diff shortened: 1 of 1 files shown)


cardano-wallet.cabal
```diff
@@ -556,8 +552,8 @@ test-suite integration
       MultiParamTypeClasses
       NoImplicitPrelude
       NoMonomorphismRestriction
-      OverloadedStrings
       OverloadedLabels
+      OverloadedStrings
```
jonathanknowles, line 556: This line movement was a side-effect of running "sort -u".

KtorZ (approved):
Sweet :+1:

Issue #236, opened 2019-01-18
Write integration tests for (and finalize implementation of) creation of E.O.S wallets


KtorZ:
# Context

Ideally, we want to replay most of our existing integration test suite for F.O. wallets against E.O. wallets (where it makes sense). As a first step, we want to at least cover the creation of such wallets and make sure we can correctly send API requests.

# Decision

Cover creation of E.O.S wallets via integration tests.

# Acceptance Criterias

- [x] Creation of E.O.S Wallets _should_ be tested and covered in CI.

---

# Development Plan

- [x] Review creation of EOS wallet (API Types, Wallet Layer, Kernel and Acid State layers)
- [x] Implement dummy getter / read endpoint for EOS wallet
- [x] Write a scenario testing creation + lookup of E.O. wallet
- [x] Add an extra guard to EOS wallet creation to make sure that given wallets can actually only be created from a _consistent_ non empty list of accounts
- [x] More testing scenarios to illustrate the above.

# PR

| Number   | Base      |
| ---      | ---       |
| #277 | `develop` |
| #305 | `develop` |
| #307 | `develop` |
| #311 | `develop` |
| #310 | `develop` | 
| #325 | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| 1.      | This is rather "basic" (cf [Scenario/EosWallets](https://github.com/input-output-hk/cardano-wallet/blob/develop/test/integration/Test/Integration/Scenario/EosWallets.hs)) but really aims at testing the core capabilities of the E.O.S wallet (DB layer and such), making sure we aren't doing anything crazy there. Eventually, we want to add more test here about other parameters of the payload (e.g. assuranceLevel or name) in a very similar fashion as we did for fully owned wallet. Yet, I'd rather leave that for a bit later for now and move on with the testing the rest of the core capabilities.  |

piotr-iohk:
lgtm :+1: 
(with the assumption this suite will be extended in the nearest future)

PR #277, opened 2019-01-30
review creation of externally owned wallets

KtorZ:
<p align="right">#236</p>

# Overview

- Remove the intermediate type 'AccountUpdateBase' in favor of plain 'HdAccountBase'

- Provide a Lens' HdAccountBase HdAccountId to easily access account id
  from a base without pattern matching all the time

- Unify 'createHdWallet' and 'createEosHdWallet' to remove duplication
  and work with a slightly highler level abstraction at the DB level
  (provided the base is constructed correctly by the caller)

- Review 'createEosHdWallet' in the WalletLayer & Kernel to have types
  better reflecting an actual _real life scenario_

# Comments

This is some preparatory work in order to implement working integration tests for creation of EOS wallets.

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/AcidState.hs
+++ b/src/Cardano/Wallet/Kernel/DB/AcidState.hs
@@ -22,7 +22,6 @@ module Cardano.Wallet.Kernel.DB.AcidState (
     -- ** Updates on HD wallets
     -- *** CREATE
   , CreateHdWallet(..)
-  , CreateEosHdWallet(..)
   , RestoreHdWallet(..)
   , CreateHdAccount(..)
   , CreateHdAddress(..)
@@ -65,9 +64,7 @@ import           Pos.Chain.Block (HeaderHash)
 import           Pos.Chain.Txp (TxAux, TxId, Utxo)
 import           Pos.Chain.Update (SoftwareVersion)
 import           Pos.Core.Chrono (OldestFirst (..))
-import           Pos.Crypto (PublicKey)
 
-import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 import           Cardano.Wallet.Kernel.DB.BlockContext
 import           Cardano.Wallet.Kernel.DB.HdRootId (HdRootId)
 import           Cardano.Wallet.Kernel.DB.HdWallet
@@ -279,7 +276,7 @@ applyBlock k context restriction blocks = runUpdateDiscardSnapshot $ do
     mkUpdate :: (HdAccountId, Maybe PrefilteredBlock)
              -> AccountUpdate Void (Either Spec.ApplyBlockFailed (Set TxId))
     mkUpdate (accId, mPB) = AccountUpdate {
-          accountUpdateBase  = AccountUpdateBaseFO accId
+          accountUpdateBase  = HdAccountBaseFO accId
         , accountUpdateAddrs = pfbAddrs pb
         , accountUpdateNew   = AccountUpdateNewUpToDate Map.empty
         , accountUpdate      =
@@ -339,7 +336,7 @@ applyHistoricalBlock k context rootId blocks =
     mkUpdate :: (HdAccountId, Maybe PrefilteredBlock)
              -> AccountUpdate Spec.ApplyBlockFailed ()
     mkUpdate (accId, mPb) = AccountUpdate {
-          accountUpdateBase  = AccountUpdateBaseFO accId
+          accountUpdateBase  = HdAccountBaseFO accId
         , accountUpdateAddrs = pfbAddrs pb
         , accountUpdateNew   = AccountUpdateNewIncomplete mempty mempty context
         , accountUpdate      = void $ Z.wrap $ \acc -> do
@@ -451,7 +448,7 @@ switchToFork k oldest blocks toSkip =
     mkUpdate :: (HdAccountId, OldestFirst [] (BlockContext, PrefilteredBlock))
              -> AccountUpdate SwitchToForkInternalError (Pending, Set TxId)
     mkUpdate (accId, pbs) = AccountUpdate {
-          accountUpdateBase  = AccountUpdateBaseFO accId
+          accountUpdateBase  = HdAccountBaseFO accId
         , accountUpdateAddrs = concatMap (pfbAddrs . snd) pbs
         , accountUpdateNew   = AccountUpdateNewUpToDate Map.empty
         , accountUpdate      =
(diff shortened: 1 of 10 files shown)


src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -502,69 +499,29 @@ observableRollbackUseInTestsOnly = runUpdateDiscardSnapshot $
...
 -- definitely we do not want it to show up in our acid-state logs.
 --
 createHdWallet :: HdRoot
-               -> HdAccountId
-               -- ^ The default HdAccountId to go with this HdRoot. This
-               -- function will take responsibility of creating the associated
-               -- 'HdAccount'.
-               -> Maybe HdAddress
-               -- ^ An optional default HdAddress to go with this HdRoot.
-               -> Map HdAccountId (Utxo,[HdAddress])
```
KtorZ, line 518: There are two changes here:

- The map is now indexed by `HdAccountBase` (which requires 'Ord' instance on this type, which is fine)
- The default account id and addresses are optional for E.O. wallets (more than optional, they actually don't exist), so, in order to save us some work here, I've simply remove them from the signature and it's now up to the caller to correctly construct the underlying Map (in most cases, a singleton with a default account id). 

This gives us an interoperable DB interface between both F.O. and E.O. wallets (cf higher layers below)

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -615,20 +572,12 @@ restoreHdWallet newRoot defaultHdAccountId defaultHdAddress ctx utxoByAccount =
   Internal: support for updating accounts
 -------------------------------------------------------------------------------}
 
-data AccountUpdateBase
-    = AccountUpdateBaseFO !HdAccountId
-    | AccountUpdateBaseEO {
-          _accountUpdateBaseEOId             :: !HdAccountId
-        , _accountUpdateBaseEOPublicKey      :: !PublicKey
-        , _accountUpdateBaseEOAddressPoolGap :: !AddressPoolGap
-      }
```
KtorZ, line 624: This was actually not needed and doesn't buy us much since this is completely isomorphic to `HdAccountBase`.

src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -708,9 +653,6 @@ updateAccount AccountUpdate{..} = do
     mapM_ createAddress accountUpdateAddrs
     return res
   where
-    getAccountId updateBase = case updateBase of
-        AccountUpdateBaseFO accId     -> accId
-        AccountUpdateBaseEO accId _ _ -> accId
```
KtorZ, line 713: As noted in a previous PR, a lens does a nicer job here instead of having to pattern match all the time :+1:

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -297,7 +298,8 @@ data HdAccountBase =
           _hdAccountBaseEOId             :: !HdAccountId
         , _hdAccountBaseEOAccountKey     :: !Core.PublicKey
         , _hdAccountBaseEOAddressPoolGap :: !AddressPoolGap
-      }
+        }
+    deriving (Eq, Ord)
```
KtorZ, line 302: Needed to have `HdAccountBase` as a `Map` key.

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -462,7 +479,7 @@ updatePassword pw hdRootId oldPassword newPassword = do
                                        else HD.HasSpendingPassword lastUpdateNow
 
                                res <- update' (pw ^. wallets)
-                                              (UpdateHdRootPassword hdRootId hasSpendingPassword)
```
KtorZ, line 465: name conflicts with the lens `hdRootId` now imported :man_shrugging:

denisshevchenko (approved):

src/Cardano/Wallet/Kernel/DB/AcidState.hs
parsonsmatt: While we're changing this around, mind making the indentation 4spc?

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -220,9 +222,14 @@ createHdWallet pw mnemonic spendingPassword assuranceLevel walletName = do
...
+--
+-- Here, we review the definition of a wallet down to a list of account public keys with
+-- no relationship whatsoever from the wallet's point of view. New addresses can be derived
+-- for each account at will and discovered using the address pool discovery algorithm
+-- described in BIP-44. Public keys are managed and provided from an external sources.
 createEosHdWallet :: PassiveWallet
-                  -> [(PublicKey, Word32)]
-                  -- ^ External wallet's accounts public keys.
+                  -> [(PublicKey, HdAccountIx)]
+                  -- ^ External wallets' accounts
```
parsonsmatt, line 232: since we're touching it, we might as well fix up the  indentation here too

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -233,26 +240,36 @@ createEosHdWallet :: PassiveWallet
...
+    toAccountBase
+        :: HdRootId
+        -> (PublicKey, HdAccountIx)
+        -> Map HdAccountBase (Utxo, [HdAddress])
+    toAccountBase rootId =
+        let
+            accId ix = HdAccountId rootId ix
+            mkBase (pk, ix) = HdAccountBaseEO (accId ix) pk gap
+        in
+            flip Map.singleton (mempty, mempty) . mkBase
```
parsonsmatt, line 271: I think we are kind of reaching on the point-free here.

```haskell
toAccountBase rootId (pk, ix) =
    Map.singleton 
        (HdAccountBaseEO (HdAccountId rootId ix) pk gap)
        (mempty, mempty)
```

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -152,31 +151,26 @@ createEosWallet :: MonadIO m
...
                 -> m (Either CreateWalletError V1.EosWallet)
-createEosWallet wallet newEosWalletRequest = runExceptT $ do
-    let accountsPublicKeysWithIxs =
-            map (\(V1.AccountPublicKeyWithIx pk ix) -> (pk, V1.getAccIndex ix)) $
-                V1.neweoswalAccounts newEosWalletRequest
-        addressPoolGap = maybe (def :: AddressPoolGap) id $
-            V1.neweoswalAddressPoolGap newEosWalletRequest
-        name = V1.neweoswalName newEosWalletRequest
-        assuranceLevel = V1.neweoswalAssuranceLevel newEosWalletRequest
+createEosWallet wallet req = runExceptT $ do
```
parsonsmatt, line 154: this looks significantly nicer :smile: :+1:

parsonsmatt (approved):

Issue #36, opened 2018-11-07
Finalize CRUD operations for E.O.S Wallets


KtorZ:
# Context

<What is the issue that we are seeing that is motivating this decision or
change. Give any elements that help understanding where this issue comes from.
Leave no room for suggestions or implicit deduction.>

Until here, we didn't need the remaining classic CRUD operations on E.O.S. wallets.

# Decision

<Give details about the architectural decision and what it is doing. Be
extensive: use schemas and references when possible. Do not hesitate to use
schemas and references when possible.>

 For completeness and because we want to make our users happy, let's implement (and test) them.

# Acceptance Criteria

<Use standard vocabulary to describe requirement levels RFC-2119: Must-Should-May.
e.g.: The API _must_ support creation of wallets through a dedicated endpoint.>

1. - [x] The API _must_ support 'Read' operation on E.O.S. Wallets.
2. - [x] The API _must_ support 'Update' operation on E.O.S. Wallets.
3. - [x] The API _must_ support 'Delete' operation on E.O.S. Wallets.
4. - [x] The API _must_ support 'List' operation on E.O.S. Wallets.
5. - [x] The 'Update' operation from #36.2 _may_ act upon the followings: name, assurance level, address gap.

---

# PR

| Number   | Base      |
| ---      | ---       |
| #278       | `develop` |
| #304       | `develop` |

# Consequences

<To be completed in retrospective. What becomes easier or more difficult to do
because of this change.>

piotr-iohk:
@denisshevchenko, @KtorZ  
There is no QA section on the ticket, however I understand that it is an old format. :)
In any case, in terms of QA, I think we should add some integration tests for CRUD operations. Probably we want to approach this in different task? (I haven't found any, so if there is one can you point me to it pls?).

There are some tests already added in #236. But they only touch on create and read. There are no tests covering list, update, delete.

KtorZ:
Indeed there's no QA / tests for this _yet_. I'll leave that decision to @denisshevchenko here; would you rather write them as part of this ticket? Or should we open another ticket a bit like #236 where you or I can write those integration tests?

piotr-iohk:
@denisshevchenko I'm closing this one. As discussed I'll be adding more integration tests for EOS wallets CRUD operations under #331.

PR #278, opened 2019-01-30
RDL-operations for EOS-wallets.

denisshevchenko:
<p align="right">#36</p>

# Overview

- [x] I have added `Read`-operation for EOS-wallets.
- [x] I have added `Delete`-operation for EOS-wallets.
- [x] I have added `List`-operation for EOS-wallets.

# Comments

From `acid-state` point of view, we store EOS-wallets as the same `HdRoot`s as FO-wallets, so implementation of `Read`, `Delete` and `List` operations is almost copy-n-paste. That's why I put it in one PR.

### DIFF
--- a/src/Cardano/Wallet/API/V1/ReifyWalletError.hs
+++ b/src/Cardano/Wallet/API/V1/ReifyWalletError.hs
@@ -43,6 +43,7 @@ translateWalletLayerErrors ex = do
 
         <|> try' @CreateWalletError createWalletError
         <|> try' @GetWalletError getWalletError
+        <|> try' @GetEosWalletError getEosWalletError
         <|> try' @UpdateWalletError updateWalletError
         <|> try' @UpdateWalletPasswordError updateWalletPasswordError
         <|> try' @DeleteWalletError deleteWalletError
@@ -155,6 +156,24 @@ getWalletError e = case e of
             V1.WalletNotFound
     (GetWalletWalletIdDecodingFailed _txt) ->
             V1.WalletNotFound
+
+getAddressPoolGapError :: GetAddressPoolGapError -> V1.WalletError
+getAddressPoolGapError e = case e of
+    ex@(GetEosWalletErrorNoAccounts _txt) ->
+        V1.EosWalletDoesNotHaveAccounts (sformat build ex)
+    ex@(GetEosWalletErrorWrongAccounts _txt) ->
+        V1.EosWalletHasWrongAccounts (sformat build ex)
+    ex@(GetEosWalletErrorGapsDiffer _txt) ->
+        V1.EosWalletGapsDiffer (sformat build ex)
+
+getEosWalletError :: GetEosWalletError -> V1.WalletError
+getEosWalletError e = case e of
+    (GetEosWalletError (HD.UnknownHdRoot _rootId)) ->
+        V1.WalletNotFound
+    (GetEosWalletWalletIdDecodingFailed _txt) ->
+        V1.WalletNotFound
+    (GetEosWalletErrorAddressPoolGap e') ->
+        getAddressPoolGapError e'
 
 updateWalletError :: UpdateWalletError -> V1.WalletError
 updateWalletError e = case e of
--- a/src/Cardano/Wallet/WalletLayer.hs
+++ b/src/Cardano/Wallet/WalletLayer.hs
@@ -6,6 +6,8 @@ module Cardano.Wallet.WalletLayer
     -- ** Errors
     , CreateWalletError(..)
     , GetWalletError(..)
+    , GetAddressPoolGapError(..)
+    , GetEosWalletError(..)
     , UpdateWalletError(..)
     , UpdateWalletPasswordError(..)
     , DeleteWalletError(..)
@@ -112,6 +114,40 @@ instance Buildable GetWalletError where
         bprint ("GetWalletErrorNotFound " % build) walletId
     build (GetWalletWalletIdDecodingFailed txt) =
         bprint ("GetWalletWalletIdDecodingFailed " % build) txt
+
+data GetAddressPoolGapError =
+      GetEosWalletErrorNoAccounts Text
+    | GetEosWalletErrorWrongAccounts Text
+    | GetEosWalletErrorGapsDiffer Text
+    deriving Eq
+
+instance Buildable GetAddressPoolGapError where
+    build (GetEosWalletErrorNoAccounts txt) =
+        bprint ("GetEosWalletErrorNoAccounts " % build) txt
+    build (GetEosWalletErrorWrongAccounts txt) =
+        bprint ("FO-accounts found in EOS-wallet " % build) txt
+    build (GetEosWalletErrorGapsDiffer txt) =
+        bprint ("Address pool gaps differ, for EOS-wallet " % build) txt
+
+data GetEosWalletError =
+      GetEosWalletError Kernel.UnknownHdRoot
+    | GetEosWalletWalletIdDecodingFailed Text
+    | GetEosWalletErrorAddressPoolGap GetAddressPoolGapError
+    deriving Eq
+
+-- | Unsound show instance needed for the 'Exception' instance.
+instance Show GetEosWalletError where
+    show = formatToString build
+
+instance Exception GetEosWalletError
+
+instance Buildable GetEosWalletError where
+    build (GetEosWalletError kernelError) =
+        bprint ("GetEosWalletError " % build) kernelError
+    build (GetEosWalletWalletIdDecodingFailed txt) =
+        bprint ("GetEosWalletWalletIdDecodingFailed " % build) txt
+    build (GetEosWalletErrorAddressPoolGap gapError) =
+        bprint ("GetEosWalletErrorAddressPoolGap " % build) gapError
 
 data UpdateWalletError =
       UpdateWalletError Kernel.UnknownHdRoot
@@ -415,9 +451,9 @@ data PassiveWalletLayer m = PassiveWalletLayer
     -- fully-owned wallets
       createWallet         :: CreateWallet -> m (Either CreateWalletError Wallet)
     , getWallets           :: m (IxSet Wallet)
-    , getEosWallets        :: m (IxSet EosWallet)
+    , getEosWallets        :: m (Either GetEosWalletError (IxSet EosWallet))
     , getWallet            :: WalletId -> m (Either GetWalletError Wallet)
-    , getEosWallet         :: WalletId -> m (Either GetWalletError EosWallet)
+    , getEosWallet         :: WalletId -> m (Either GetEosWalletError EosWallet)
     , updateWallet         :: WalletId
                            -> WalletUpdate
                            -> m (Either UpdateWalletError Wallet)
--- a/src/Cardano/Wallet/WalletLayer.hs
+++ b/src/Cardano/Wallet/WalletLayer.hs
@@ -9,7 +9,6 @@ module Cardano.Wallet.WalletLayer
     , UpdateWalletError(..)
     , UpdateWalletPasswordError(..)
     , DeleteWalletError(..)
-    , DeleteEosWalletError(..)
     , GetUtxosError(..)
     , NewPaymentError(..)
     , EstimateFeesError(..)
@@ -167,19 +166,6 @@ instance Buildable DeleteWalletError where
         bprint ("DeleteWalletWalletIdDecodingFailed " % build) txt
     build (DeleteWalletError kernelError) =
         bprint ("DeleteWalletError " % build) kernelError
-
-data DeleteEosWalletError =
-    DeleteEosWalletError Kernel.UnknownHdRoot
-
--- | Unsound show instance needed for the 'Exception' instance.
-instance Show DeleteEosWalletError where
-    show = formatToString build
-
-instance Exception DeleteEosWalletError
-
-instance Buildable DeleteEosWalletError where
-    build (DeleteEosWalletError kernelError) =
-        bprint ("DeleteEosWalletError " % build) kernelError
 
 data GetUtxosError =
       GetUtxosWalletIdDecodingFailed Text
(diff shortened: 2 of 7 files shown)


src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
```diff
@@ -180,6 +181,23 @@ toWallet db hdRoot = V1.Wallet {
     walletId         = sformat build rootId
     v1AssuranceLevel = toAssuranceLevel $ hdRoot ^. HD.hdRootAssurance
 
+-- | Converts an 'HdRoot' into a V1 'EosWallet'.
+toEosWallet :: Kernel.DB -> HD.HdRoot -> V1.EosWallet
+toEosWallet db hdRoot = V1.EosWallet {
+      eoswalId             = V1.WalletId walletId
+    , eoswalName           = hdRoot ^. HD.hdRootName
+                                     . to HD.getWalletName
+    , eoswalAddressPoolGap = error "TODO, we have to take it from one of accounts, will be done later, see issue #36"
```
denisshevchenko: We store address pool gap in `HdAccount`s, not in `HdRoot`, so we must take any of corresponding accounts and get the gap from it. It will be done later.

src/Cardano/Wallet/WalletLayer.hs
```diff
@@ -168,19 +167,6 @@ instance Buildable DeleteWalletError where
...
--- | Unsound show instance needed for the 'Exception' instance.
-instance Show DeleteEosWalletError where
-    show = formatToString build
-
-instance Exception DeleteEosWalletError
-
-instance Buildable DeleteEosWalletError where
-    build (DeleteEosWalletError kernelError) =
-        bprint ("DeleteEosWalletError " % build) kernelError
-
```
denisshevchenko, line 183: I thought that we can use the same `DeleteWalletError`, because we actually delete the same `HdRoot` as for FO-wallets.

akegalj (approved):
I just skimmed through it but it looks good to me

src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
```diff
@@ -159,26 +160,51 @@ toAccount snapshot account = V1.Account {
...
+       , Text
+       , Text
+       )
+getBasicWalletData hdRoot = (rootId, createdAt, level, walletName, walletId)
+  where
+    rootId     = hdRoot ^. HD.hdRootId
+    createdAt  = hdRoot ^. HD.hdRootCreatedAt . fromDb
+    level      = hdRoot ^. HD.hdRootAssurance
+    walletName = hdRoot ^. HD.hdRootName . to HD.getWalletName
+    walletId   = sformat build rootId
```
KtorZ: I'd rather not have that function. Having some similar where clauses in both `toEosWallet` and `toWallet` is fine...

src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
KtorZ: What about doing it as part of this PR :) ?

KtorZ (changes requested):

src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
denisshevchenko: `hlint` is angry in this case (Duplication). :-)

KtorZ: Well, that wouldn't be the first case `hlint` suggests something wrong :upside_down_face:

KtorZ: Unless you truly think and argue that this makes the code looks better ...

KtorZ: So we've reduce duplication to avoid having to modify code in two different places. But now, we'd have to modify it in 3 different places. Not sure what we've improved here to be honest. 

In order to really remove duplication, we could have provided a different `toWallet` function that may just construct the first part common part of an `EosWallet` or `Wallet` and leave the rest to the caller... But to be honest, this complexifies the whole for close to no value. 

So yes, I am arguably saying that `hlint` can go to hell here :)

src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
denisshevchenko: @KtorZ Done (please see last commit).

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -242,6 +248,36 @@ getWallet wallet wId db = runExceptT $ do
...
+addressPoolGapByRootId rootId db =
+    if IxSet.null accountSet
+        then def :: AddressPoolGap
+        else gap
+  where
+    accountSet = Kernel.accountsByRootId db rootId
+    -- Since this function is using for EOS-wallets only,
+    -- we definitely know that this is an account with EO-branch.
+    HD.HdAccountBaseEO _ _ gap = firstAccount ^. HD.hdAccountBase
+    (firstAccount:_) = IxSet.toList accountSet
```
KtorZ: Hmm. This feels wrong and may eventually bite us. 
We should:

- Forbid empty E.O. wallets creation (not in this PR, but I'll add a bullet point to #236)
- Make sure that all accounts have a same gap within a wallet
- Return an error here if there's no account

src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
```diff
@@ -180,6 +186,23 @@ toWallet db hdRoot = V1.Wallet {
     walletId         = sformat build rootId
     v1AssuranceLevel = toAssuranceLevel $ hdRoot ^. HD.hdRootAssurance
 
+-- | Converts an 'HdRoot' into a V1 'EosWallet'.
+toEosWallet :: AddressPoolGap -> Kernel.DB -> HD.HdRoot -> V1.EosWallet
+toEosWallet gap db hdRoot = V1.EosWallet {
+      eoswalId             = V1.WalletId walletId
+    , eoswalName           = hdRoot ^. HD.hdRootName
+                                     . to HD.getWalletName
```
KtorZ: Please, no more variable-length indent :pray:

src/Cardano/Wallet/WalletLayer/Kernel/Conv.hs
```diff
@@ -62,6 +64,10 @@ import qualified Cardano.Wallet.Kernel.Read as Kernel
 import           UTxO.Util (exceptT)
 -- import           Cardano.Wallet.WalletLayer (InvalidRedemptionCode (..))
 
+-- Functions 'toWallet' and 'toEosWallet' contain duplications in 'where'
+-- sections, but this is not a problem.
```
KtorZ: This doesn't really explain _why_ it's not a problem :/ ..

KtorZ (changes requested):

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
denisshevchenko: > Make sure that all accounts have a same gap within a wallet

Of course they have the same gap, because they was _created_ with the same gap (during EOS-wallet creation).

> Return an error here if there's no account

Do you mean "if there's no accounts at all"? In this case we just return default value of gap. But I strongly believe that we shouldn't be able to create new EOS-wallet without accounts!

KtorZ: > Of course they have the same gap,

They **ought to** have the same gap. But there's nothing that can be checked at compile-time that guarantees this invariant. Hence the need for checking it and, throwing an exception if it doesn't hold. 

>  In this case we just return default value of gap.

No, that's the point. If there's no accounts, that's an error! This shouldn't happen and we should enforce that via the API. If it ever happens, that's a bug / fault in our implementation

denisshevchenko:
@KtorZ I've added checks for `HdAccount`s in EOS-wallets, with corresponding errors.

src/Cardano/Wallet/API/V1/ReifyWalletError.hs
```diff
@@ -156,6 +157,19 @@ getWalletError e = case e of
...
+    (GetEosWalletError (HD.UnknownHdRoot _rootId)) ->
+        V1.WalletNotFound
+    (GetEosWalletWalletIdDecodingFailed _txt) ->
+        V1.WalletNotFound
+    ex@(GetEosWalletErrorNoAccounts _txt) ->
+        V1.UnknownError (sformat build ex)
+    ex@(GetEosWalletErrorWrongAccounts _txt) ->
+        V1.UnknownError (sformat build ex)
+    ex@(GetEosWalletErrorGapsDiffer _txt) ->
+        V1.UnknownError (sformat build ex)
```
KtorZ: This is a new feature, so we have some freedom here in what errors we define. There's actually no particular need to reify anything here and we could have a 1:1 mapping between those errors and new API errors. 

As a reminder, those API errors are so far, our only debugging tools when somethings goes wrong on the wallet of our users and for them, that's the only feedback they get from the API. So, returning `UnknownError` shouldn't only happen in exceptional situation (read this as: never, if possible).

KtorZ (changes requested):

src/Cardano/Wallet/API/V1/ReifyWalletError.hs
denisshevchenko: Thanks, fixed.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -242,6 +252,38 @@ getWallet wallet wId db = runExceptT $ do
...
+    hdRoot <- withExceptT GetEosWalletError $ exceptT $
+        Kernel.lookupHdRootId db rootId
+    fmap (toEosWallet db hdRoot) $ exceptT $
+        addressPoolGapByRootId rootId db
+
+addressPoolGapByRootId
+    :: HD.HdRootId
+    -> Kernel.DB
+    -> Either GetEosWalletError AddressPoolGap
+addressPoolGapByRootId rootId db = do
```
KtorZ, line 274: So ideally, this would return a specific error as well. Truth is that the branches below of `GetEosWalletError` are only met within this function. And, other branches can never be met by this function.
So, saying that this function can return a `WalletIdDecodingFailed` is a bit of a lie :sweat_smile: ...

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -259,6 +301,22 @@ getWallets wallet db =
...
+    -> Kernel.DB
+    -> m (Either GetEosWalletError (IxSet V1.EosWallet))
+getEosWallets _wallet db = do
+    result <- forM allRoots $ \root -> runExceptT $
+        fmap (toEosWallet db root) $ exceptT $
+            addressPoolGapByRootId (root ^. HD.hdRootId) db
+    let (errors, eosWallets) = partitionEithers result
+    return $ if not . null $ errors
+        then let (err:_) = errors in Left err
+        else Right $ IxSet.fromList eosWallets
```
KtorZ: A case reads better for this kind of manipulation :+1:

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -259,6 +301,22 @@ getWallets wallet db =
...
+getEosWallets
+    :: MonadIO m
+    => Kernel.PassiveWallet
+    -> Kernel.DB
+    -> m (Either GetEosWalletError (IxSet V1.EosWallet))
+getEosWallets _wallet db = do
+    result <- forM allRoots $ \root -> runExceptT $
+        fmap (toEosWallet db root) $ exceptT $
+            addressPoolGapByRootId (root ^. HD.hdRootId) db
+    let (errors, eosWallets) = partitionEithers result
```
KtorZ: We aren't quite leveraging the applicative aspect of `Either` and `ExceptT` here. Especially if we just intend to stop at the first error, we could instead `traverse` the list of hd roots and end up with a `Either Error [EosWallet]`

KtorZ (changes requested):

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
denisshevchenko: Done.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
denisshevchenko: You're right! Fixed.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
denisshevchenko: Done.

denisshevchenko:
@KtorZ All checks have passed (`hlint` is happy again).

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -259,6 +302,21 @@ getWallets wallet db =
   where
     allRoots = db ^. dbHdWallets . HD.hdWalletsRoots
 
+getEosWallets
+    :: MonadIO m
+    => Kernel.PassiveWallet
+    -> Kernel.DB
+    -> m (Either GetEosWalletError (IxSet V1.EosWallet))
+getEosWallets _wallet db = do
+    let result = traverse (\root -> either Left (Right . toEosWallet db root) $
```
KtorZ, line 311: Note that:

```hs
\root -> either Left (Right . toEosWallet db root)
````

can be written more idiomatically as:

```
fmap (toEosWallet db)
```

KtorZ (approved):

PR #283, opened 2019-01-30
Start node api server as a plugin

Anviking:
Previously we launched a client expecting a node monitoring api server
to be running, without actually starting the server. The integration
tests still worked, as we connect to the api of a relay node.

This commit adds a wallet `Plugin` for starting the node server, as well
as CLI options for its tls settings for starting the node server, as
well as CLI options for its tls settings.

<p align="right">#87</p>

# Overview

- [x] `Plugin` for launching node monitoring api server
- [x] Added CLI param ` --node-tls-server-cert            ./state-integration/tls/edge/server.crt`
- [x] Added CLI param `--node-tls-server-key             ./state-integration/tls/edge/server.key`

# Comments

- [ ] Will need to update wiki
- [ ] Have not reused wallet and node tls params. Maybe I should
- [ ] Naming of CLI params should be looked at (I might have messed it up completely)

### DIFF
--- a/src/Cardano/Wallet/Action.hs
+++ b/src/Cardano/Wallet/Action.hs
@@ -38,6 +38,7 @@ import           Cardano.Wallet.Server.Middlewares
 import qualified Cardano.Wallet.Server.Plugins as Plugins
 import           Cardano.Wallet.WalletLayer (PassiveWalletLayer)
 import qualified Cardano.Wallet.WalletLayer.Kernel as WalletLayer.Kernel
+import           Pos.Web (TlsParams (..))
 
 
 -- | The "workhorse" responsible for starting a Cardano edge node plus a number of extra plugins.
@@ -59,12 +60,15 @@ actionWithWallet
     ntpStatus <- withNtpClient (ntpClientSettings ntpConfig)
     userSecret <- readTVarIO (ncUserSecret $ nrContext nodeRes)
 
+    logInfo $ "Params: " <> show params
+
     let (nodeIp, nodePort) = walletNodeAddress
+    let walletNodeTlsCaCertPath = tpCaPath $ maybe (error "TODO") id walletTLSParams
     nodeClient <- Plugins.setupNodeClient
          (BS8.unpack nodeIp, fromIntegral nodePort)
          walletNodeTlsClientCert
          walletNodeTlsCaCertPath
-         walletNodeTlsPrivKey
+         walletNodeTlsClientKey
 
     let nodeState = newNodeStateAdaptor
             genesisConfig
--- a/src/Cardano/Wallet/Action.hs
+++ b/src/Cardano/Wallet/Action.hs
@@ -30,15 +30,13 @@ import qualified Cardano.Wallet.Kernel.Mode as Kernel.Mode
 import           Cardano.Wallet.Kernel.NodeStateAdaptor (newNodeStateAdaptor)
 import qualified Cardano.Wallet.Kernel.NodeStateAdaptor as NodeStateAdaptor
 import           Cardano.Wallet.Server.CLI (WalletBackendParams (..),
-                     getFullMigrationFlag, getWalletDbOptions, walletDbPath,
-                     walletNodeAddress, walletRebuildDb)
+                     walletDbPath, walletNodeAddress, walletRebuildDb)
 import           Cardano.Wallet.Server.Middlewares
                      (faultInjectionHandleIgnoreAPI, throttleMiddleware,
                      unsupportedMimeTypeMiddleware, withDefaultHeader)
 import qualified Cardano.Wallet.Server.Plugins as Plugins
 import           Cardano.Wallet.WalletLayer (PassiveWalletLayer)
 import qualified Cardano.Wallet.WalletLayer.Kernel as WalletLayer.Kernel
-import           Pos.Web (TlsParams (..))
 
 
 -- | The "workhorse" responsible for starting a Cardano edge node plus a number of extra plugins.
@@ -60,15 +58,10 @@ actionWithWallet
     ntpStatus <- withNtpClient (ntpClientSettings ntpConfig)
     userSecret <- readTVarIO (ncUserSecret $ nrContext nodeRes)
 
-    logInfo $ "Params: " <> show params
-
     let (nodeIp, nodePort) = walletNodeAddress
-    let walletNodeTlsCaCertPath = tpCaPath $ maybe (error "TODO") id walletTLSParams
     nodeClient <- Plugins.setupNodeClient
-         (BS8.unpack nodeIp, fromIntegral nodePort)
-         walletNodeTlsClientCert
-         walletNodeTlsCaCertPath
-         walletNodeTlsClientKey
+        (BS8.unpack nodeIp, fromIntegral nodePort)
+        nodeTLSParams
 
     let nodeState = newNodeStateAdaptor
             genesisConfig
@@ -78,9 +71,8 @@ actionWithWallet
             nodeClient
 
     liftIO $ Keystore.bracketLegacyKeystore userSecret $ \keystore -> do
-        let dbOptions = getWalletDbOptions params
-        let dbPath = walletDbPath dbOptions
-        let rebuildDB = walletRebuildDb dbOptions
+        let dbPath = walletDbPath walletDbOptions
+        let rebuildDB = walletRebuildDb walletDbOptions
         let dbMode = Kernel.UseFilePath (Kernel.DatabaseOptions {
               Kernel.dbPathAcidState = dbPath <> "-acid"
             , Kernel.dbPathMetadata  = dbPath <> "-sqlite.sqlite3"
@@ -88,7 +80,7 @@ actionWithWallet
             })
         let pm = configProtocolMagic genesisConfig
         WalletLayer.Kernel.bracketPassiveWallet pm dbMode logMessage' keystore nodeState (npFInjects nodeParams) $ \walletLayer passiveWallet -> do
-            migrateLegacyDataLayer passiveWallet dbPath (getFullMigrationFlag params)
+            migrateLegacyDataLayer passiveWallet dbPath forceFullMigration
 
             let plugs = plugins (walletLayer, passiveWallet) nodeClient dbMode
 
@@ -122,8 +114,6 @@ actionWithWallet
         -- server which doesn't require client x509 certificates to
         -- connect, but still serves the doc through TLS
         , maybe [] (pure . ("doc server",)) (Plugins.docServer params)
-        -- The monitoring API for the Core node.
-        , Plugins.monitoringServer params
         ]
 
     -- Extract the logger name from node parameters
--- a/src/Cardano/Wallet/Action.hs
+++ b/src/Cardano/Wallet/Action.hs
@@ -112,6 +112,7 @@ actionWithWallet
 
             -- Periodically compact & snapshot the acid-state database.
             , ("acid state cleanup", Plugins.acidStateSnapshots (view Kernel.Internal.wallets (snd w)) params dbMode)
+            , ("node monitoring server", Plugins.nodeAPIServer params genesisConfig ntpConfig nodeRes)
             ]
         -- The corresponding wallet documention, served as a different
         -- server which doesn't require client x509 certificates to
--- a/src/Cardano/Wallet/Server/CLI.hs
+++ b/src/Cardano/Wallet/Server/CLI.hs
@@ -74,17 +74,10 @@ data WalletBackendParams = WalletBackendParams
     , forceFullMigration      :: !Bool
     , walletNodeAddress       :: !NetworkAddress
     -- ^ The IP address and port for the node backend.
-    , walletNodeTlsClientCert :: FilePath
-    -- ^ A filepath to the Node's public certficate
-    , walletNodeTlsServerCert :: FilePath
-    -- ^ A filepath to the Node's private certficate
-    , walletNodeTlsPrivKey    :: FilePath
+    , walletNodeTlsClientCert :: !FilePath
+    -- ^ A filepath to the Node's client certficate
+    , walletNodeTlsClientKey  :: !FilePath
     -- ^ A filepath to the TLS private key for the Node API.
-    , walletNodeTlsPublicKey  :: FilePath
-    -- ^ A filepath to the TLS public key for the Node API.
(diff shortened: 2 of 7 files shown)


src/Cardano/Wallet/Server/CLI.hs
```diff
@@ -76,8 +76,12 @@ data WalletBackendParams = WalletBackendParams
     -- ^ The IP address and port for the node backend.
     , walletNodeTlsClientCert :: FilePath
     -- ^ A filepath to the Node's public certficate
+    , walletNodeTlsServerCert :: FilePath
+    -- ^ A filepath to the Node's private certficate
     , walletNodeTlsPrivKey    :: FilePath
     -- ^ A filepath to the TLS private key for the Node API.
+    , walletNodeTlsPublicKey  :: FilePath
+    -- ^ A filepath to the TLS public key for the Node API.
```
KtorZ: What do you mean by 'PublicKey' :thinking: ?

TLS Private Key Infrastructures (PKI) for Cardano comes as follows:

- One Certificate Authority (CA) public certificate (`ca.crt`) and a corresponding private key (`ca.key`) 
- One server public certificate (`servert.crt`) and corresponding private key (`server.crt`)
- One client public certificate (`client.crt`) and corresponding private key (`client.key`)

> NOTE: Our x509 tool also generates some `*.pem` files that are simply an aggregatoin of the key + crt, handy to use with `cURL` of whatever tool that supports this format.

The `ca.key` is used to sign the `server.crt` and `client.crt`. So, using the `ca.cert`, both the server and the client are able to verify whether a given certificate has been signed using the CA key without having access to the CA private key! (And that's the whole goal of this asymmetric cryptographic scheme after all. It allows the server and client to "handshake" and agreed on a symmetric encryption algorithm to continue speaking, after they both have verified their respective identity). 

In the context of the wallet CLI, we actually have 2 servers, one of them being also the client of the second one.
So, the node monitoring server is one, and therefore, needs a CA public certs to verify clients certificates that'll presented to it. It also needs a server public cert and private key in order to:

1/ Advertise its own identity
2/ Sign its message to prove it's actually who is says it is.

Then, we have the wallet backend server (the API), which also needs a CA cert, server private key and server public key for the same reason. Yet, it also needs a public client certificate and corresponding private key in order to authenticate to the node monitoring API. 

In theory, we could have 2 PKI here (with two differents CA), one for the node server certs (and its corresponding client) and one for the wallet backend. In practice, since the wallet here is starting the node server as one of its plugin, we could simply re-use the certificates already provided to the wallet backend. Both servers would use the same server certs and keys coming from a single CI. Then, one can use the same client cert and key to talk to the wallet backend and the node monitoring API. 
So, we should need:

- The corresponding CA public cert
- A server public cert signed by the CA
- A server private key corresponding to the server cert
- A client public cert signed by the CA
- A client private key corresponding to the client cert

And.. that's it. Is that ~complicated~ clear enough :smile: ?

src/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -207,6 +215,44 @@ instance Buildable Servant.NoContent where
...
+                    genConfig
+                    (hoistDiffusion
+                        (\x -> runReaderT x c)
+                        (\x -> lift x)
+                        diffusion))
+  where
+    tls = getNodeServerTlsParams walletParams
+
+    walletToNodeArgs WalletBackendParams{..} =
+        NodeApiArgs walletNodeAddress (Just tls) False ("localhost", 4323)
```
KtorZ: Isn't this already somewhere in `lib` or `node` on cardano-sl :thinking: ?

parsonsmatt (commented):
I thought the entire point of splitting off the `node` server into it's own process was to *not* run the node server from the wallet executable. This PR appears to reintroduce the node server into the wallet process. Is that what we want?

KtorZ:
@parsonsmatt Ultimately, that's our goal yes (to split off the node server). But for the next three releases (1.5, 1.6 and 1.7), we won't likely do that because this doesn't add any particular business value but introduces a few risks and extra complexity for Daedalus and exchanges.

We need this decoupling to start experimenting with the Rust node and the new Haskell node, but that's something we're going to do "on the side", in a "dev mode" fashion while we still release a coupled wallet (there's no way to release the wallet in a decoupled fashion ATM anyway :|)

KtorZ:
@Anviking So, while reviewing and testing this locally, I got mad at the CLI and thought this would just be the time to also bite the bullet and clean-up this whole thing :sweat_smile: We'll have to update docs on `cardano-sl` and probably a few nix scripts when we decide to merge this into `cardano-sl/develop` (so, for 1.6 or later). For now, we end up with something like this:

```
stack exec -- cardano-wallet-server \
  --node-id                    edge                                   \
  --system-start               0                                      \
                                                                      \
  --topology                   state-integration/topology/edge.json   \
  --configuration-file         test/integration/configuration.yaml    \
                                                                      \
  --db-path                    state-integration/db                   \
  --wallet-db-path             state-integration/wallet-db            \
  --wallet-rebuild-db                                                 \
                                                                      \
  --wallet-api-address         127.0.0.1:8090                         \
  --wallet-doc-address         127.0.0.1:8190                         \
  --node-api-address           127.0.0.1:8080                         \
  --node-doc-address           127.0.0.1:8180                         \
                                                                      \
  --tls-ca-cert                state-integration/tls/edge/ca.crt      \
  --tls-wallet-server-cert     state-integration/tls/edge/server.crt  \
  --tls-wallet-server-key      state-integration/tls/edge/server.key  \
  --tls-node-client-cert       state-integration/tls/edge/client.crt  \
  --tls-node-client-key        state-integration/tls/edge/client.key  \
```

src/Cardano/Wallet/Action.hs
```diff
@@ -61,10 +60,8 @@ actionWithWallet
 
     let (nodeIp, nodePort) = walletNodeAddress
     nodeClient <- Plugins.setupNodeClient
-         (BS8.unpack nodeIp, fromIntegral nodePort)
-         walletNodeTlsClientCert
-         walletNodeTlsCaCertPath
-         walletNodeTlsPrivKey
```
KtorZ, line 67: Small refactor here, I've favored the record alias as it makes a bit less ambiguous than 3 `FilePath` arguments.

src/Cardano/Wallet/Action.hs
```diff
@@ -74,17 +71,16 @@ actionWithWallet
             nodeClient
 
     liftIO $ Keystore.bracketLegacyKeystore userSecret $ \keystore -> do
-        let dbOptions = getWalletDbOptions params
-        let dbPath = walletDbPath dbOptions
-        let rebuildDB = walletRebuildDb dbOptions
```
KtorZ, line 79: While cleaning up, I've removed the `getWalletDbOptions` accessor which was just.... mirroring `walletDbOptions`. Cf below.

src/Cardano/Wallet/Action.hs
```diff
@@ -74,17 +71,16 @@ actionWithWallet
...
+        let dbPath = walletDbPath walletDbOptions
+        let rebuildDB = walletRebuildDb walletDbOptions
         let dbMode = Kernel.UseFilePath (Kernel.DatabaseOptions {
               Kernel.dbPathAcidState = dbPath <> "-acid"
             , Kernel.dbPathMetadata  = dbPath <> "-sqlite.sqlite3"
             , Kernel.dbRebuild       = rebuildDB
             })
         let pm = configProtocolMagic genesisConfig
         WalletLayer.Kernel.bracketPassiveWallet pm dbMode logMessage' keystore nodeState (npFInjects nodeParams) $ \walletLayer passiveWallet -> do
-            migrateLegacyDataLayer passiveWallet dbPath (getFullMigrationFlag params)
```
KtorZ, line 87: Same as `getWalletDbOptions`.

src/Cardano/Wallet/Action.hs
```diff
@@ -112,13 +108,12 @@ actionWithWallet
...
             -- Periodically compact & snapshot the acid-state database.
             , ("acid state cleanup", Plugins.acidStateSnapshots (view Kernel.Internal.wallets (snd w)) params dbMode)
+            , ("node monitoring server", Plugins.nodeAPIServer params genesisConfig ntpConfig nodeRes)
             ]
         -- The corresponding wallet documention, served as a different
         -- server which doesn't require client x509 certificates to
         -- connect, but still serves the doc through TLS
         , maybe [] (pure . ("doc server",)) (Plugins.docServer params)
-        -- The monitoring API for the Core node.
-        , Plugins.monitoringServer params
```
KtorZ, line 121: Removed old monitoring server.

src/Cardano/Wallet/Server/CLI.hs
```diff
@@ -51,58 +45,24 @@ data WalletDBOptions = WalletDBOptions {
...
-data RunMode = ProductionMode
-             -- ^ Run in production mode
-             | DebugMode
-             -- ^ Run in debug mode
-             deriving Show
-
--- | Converts a @GenesisKeysInclusion@ into a @Bool@.
-isDebugMode :: RunMode -> Bool
-isDebugMode ProductionMode = False
-isDebugMode DebugMode      = True
```
KtorZ, line 104: I've removed the debug mode. It's only usage was to disable TLS and.. well, we'll do without.

KtorZ (approved):
Moving on with this to finally get the wallet server back up.

src/Cardano/Wallet/Action.hs
akegalj: yes, this can be easily mixed up

src/Cardano/Wallet/Action.hs
```diff
@@ -112,13 +108,12 @@ actionWithWallet
 
             -- Periodically compact & snapshot the acid-state database.
             , ("acid state cleanup", Plugins.acidStateSnapshots (view Kernel.Internal.wallets (snd w)) params dbMode)
+            , ("node monitoring server", Plugins.nodeAPIServer params genesisConfig ntpConfig nodeRes)
```
akegalj, line 111: my first intuition was that api-server is not monitoring server but wallet-api or node-api , but I might be biased as I am a developer of wallet-api

src/Cardano/Wallet/Server/CLI.hs
```diff
@@ -26,15 +26,9 @@ import           Pos.Web (TlsParams (..))
 -- plus the wallet backend specific options.
 data WalletStartupOptions = WalletStartupOptions {
       wsoNodeArgs            :: !CommonNodeArgs
-    , wsoWalletBackendParams :: !ChooseWalletBackend
+    , wsoWalletBackendParams :: !WalletBackendParams
```
akegalj, line 29: :+1:

src/Cardano/Wallet/Server/CLI.hs
akegalj: so - tls will be always on?

This makes sense - as tls configuration for debug mode was introduced as an artifact of us adding tls after initial release - so it was cumbersome for our users to switch to tls. It was also annoying for some teams within which were not interested in tls

src/Cardano/Wallet/Server/CLI.hs
```diff
@@ -119,116 +79,71 @@ getWalletNodeOptions = execParser programInfo
...
-    monitoringApiPortParser :: Parser Word16
-    monitoringApiPortParser = CLI.webPortOption 8080
-        $ "Port for the monitoring API. This option doesn't do anything "
-        <> "anymore, as the monitoring API has been folded into the API exposed"
-        <> "by the node process."
 
     walletAddressParser :: Parser NetworkAddress
     walletAddressParser = networkAddrOption
-        $  long "wallet-address"
+        $  long "wallet-api-address"
```
akegalj, line 126: these will be breaking change for exchanges - so they should be pinged about it (or release notes about this should be written)

src/Cardano/Wallet/Server/CLI.hs
akegalj: Edit: I see I was wrong above ^, there is still an option to configure tls but not with debug mode https://github.com/input-output-hk/cardano-wallet/pull/283/files#diff-d76582b43d6f60e46f6101e79c115719R162 

That makes sense as well

src/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -206,37 +197,47 @@ acidStateSnapshots dbRef params dbMode = const worker
...
+        ntpConfig
+        nodeResources
+        updateConfiguration
+        compileInfo
+        genConfig
+        (hoistDiffusion (flip runReaderT env) lift diffusion)
+  where
+    apiArgs = NodeApiArgs
+        (walletNodeAddress params)
+        (Just $ walletTLSParams params) -- NOTE Using the same certs for both wallet server and node server
```
akegalj, line 220: I think this is ok

src/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -206,37 +197,47 @@ acidStateSnapshots dbRef params dbMode = const worker
...
+        (walletNodeDocAddress params)
 
--- Copied from the old wallet/integration/SEtupTestEnv.hs
 setupNodeClient
     :: MonadIO m
     => (String, Int)
-    -> FilePath
-    -> FilePath
-    -> FilePath
+    -> TlsParams
```
akegalj, line 227: :+1:

test/integration/Main.hs
```diff
@@ -30,7 +30,7 @@ main = do
         , ("relay", NodeRelay)
         , ("edge", NodeEdge)
         ]
-    let wAddr   = toBaseUrl $ env ! "WALLET_ADDRESS"
+    let wAddr   = toBaseUrl $ env ! "WALLET_API_ADDRESS"
```
akegalj, line 33: :+1:

akegalj (commented):
Sorry for the late review, LGTM

src/Cardano/Wallet/Action.hs
Anviking: I called it "node monitoring server" because it was the term I'd seen everywhere, though I was a bit confused by it

@KtorZ ? I guess we (still) want to call it "monitoring server" because it is a very auxiliary feature of the node?

KtorZ: Indeed @Anviking . We call it monitoring api so far because that's so-to-speak its purpose. Yet, calling that "node api" is just slightly more general.

PR #298, opened 2019-02-04
[190] Adding better checks when multioutputTransactionScenario

paweljakubas:
<p align="right">#190</p>

# Overview

- [x] I have added two element list type to better check in multioutputTransactionScenario function

# Comments

### DIFF
--- a/test/integration/Test/Integration/Scenario/Transactions.hs
+++ b/test/integration/Test/Integration/Scenario/Transactions.hs
@@ -17,40 +17,40 @@ spec = do
     multioutputTransactionScenario
         "proper fragmentation of utxo - both utxos can seperately cover fee and outputs"
         [200000, 200000]
-        [10,11]
+        (10, 11)
         [ expectTxStatusEventually [InNewestBlocks] ]
 
     -- estimated fee amount for this transaction is 187946
     multioutputTransactionScenario
         "proper fragmentation of utxo - 2 utxos available, each cannot seperately cover fee and outputs, but jointly can"
         [100000, 100000]
-        [10,11]
+        (10, 11)
         [ expectTxStatusEventually [InNewestBlocks] ]
 
     -- estimated fee amount for this transaction is 196076
     multioutputTransactionScenario
         "proper fragmentation of utxo - 3 utxos available, each cannot seperately cover fee and outputs, but jointly can"
         [70000, 70000, 70000]
-        [10,11]
+        (10, 11)
         [ expectTxStatusEventually [InNewestBlocks] ]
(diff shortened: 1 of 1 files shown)


test/integration/Test/Integration/Framework/Auxiliary.hs
```diff
@@ -0,0 +1,24 @@
...
+import qualified Data.List as L
+
+data FixedSizedList (n :: Nat) a = UnsafeMkList { getList :: [a] }
+    deriving Show
+
+type TwoElementList a = FixedSizedList 2 a
+
+mkFixedSizedList :: forall n a. KnownNat n => [a] -> Maybe (FixedSizedList n a)
+mkFixedSizedList v | L.length v == l = Just (UnsafeMkList v)
+                   | otherwise       = Nothing
```
KtorZ: Minor things but: no more variable-length indentation please :pray:

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -344,35 +345,42 @@ spec = do
     multioutputTransactionScenario
         :: String
         -> [Word64]
-        -> [Word64]
+        -> Maybe (TwoElementList Word64)
```
KtorZ: This shouldn't be a `Maybe`; this is made fairly obvious by the way the function handles that with a `case` right away, failing hard in the case it receives a `Nothing`. Generally-speaking, make the input domain no more than what you need. 
This is up to the caller to produce a value of the right type.

KtorZ (changes requested):
I think we are a bit over-doing it here. Since this is just a local function and a fairly limited scope, it sounds a bit too much albeit being indeed more type-safe. 
Though, using just a 2-tuple or a record or a dedicated type would work the same.  

In any case, I think we could simply get away with a clear invariant, before executing the scenario, verifying a bunch of precondition on the inputs. If not met, it fails hard with a clear and helpful error message. This code runs only in our testsuite and we should strive to keep it rather simple; still, making sure we can debug it easily is of the utmost importance.

paweljakubas:
thanks for the remarks! I reconsidered the issue and finally opted for 2-tuple (although still think FixedSizedList has the potential to be used in the future but needs more helper functions to be added)

test/integration/Test/Integration/Scenario/Transactions.hs
paweljakubas: yes, the caller should take responsibility here

KtorZ (approved):
:tada:

Issue #299, opened 2019-02-04
Unable to link to tags in docs


Anviking:
| Release | Operating System       | Cause |
| ---     | ---                    | ---   |
| next    | Windows & OSX & Linux) | Code  v Configuration v Environment v Human v Unknown |

# Context
 
 

# Steps to Reproduce

1. Visit https://input-output-hk.github.io/cardano-wallet/#tag/Info

## Expected behavior

The docs load scrolled down to the "Info" section

<img width="1361" alt="skarmavbild 2019-02-04 kl 15 53 14" src="https://user-images.githubusercontent.com/304423/52215742-06c01300-2895-11e9-879a-2eb0534b0010.png">

## Actual behavior
 
<A clear and concise description of what you observe instead. If applicable add screenshots to help explain your problem.>

The page loads at the top of the page.

<img width="1317" alt="skarmavbild 2019-02-04 kl 15 53 10" src="https://user-images.githubusercontent.com/304423/52215737-0293f580-2895-11e9-9afb-69439f833c8f.png">

---

# Resolution Plan

https://github.com/input-output-hk/cardano-wallet/issues/299#issuecomment-460285541

# PR

| Number   | Base      |
| ---      | ---       |
| #300        | `develop` |
| input-output-hk/cardano-sl#4064       | `develop` |

# QA 

### Wallet API Docs
 Visiting https://input-output-hk.github.io/cardano-wallet/#tag/Info, now loads to the `Info` section/tag.

### Node API Docs

```bash
stack exec -- cardano-wallet-server \
                                                              --node-id                    edge                                   \
                                                              --system-start               0                                      \
                                                                                                                                  \
                                                              --topology                   state-integration/topology/edge.json   \
                                                              --configuration-file         test/integration/configuration.yaml    \
                                                                                                                                  \
                                                              --db-path                    state-integration/db                   \
                                                              --wallet-db-path             state-integration/wallet-db            \
                                                              --wallet-rebuild-db                                                 \
                                                                                                                                  \
                                                              --wallet-api-address         127.0.0.1:8090                         \
                                                              --wallet-doc-address         127.0.0.1:8190                         \
                                                              --node-api-address           127.0.0.1:8080                         \
                                                              --node-doc-address           127.0.0.1:8180                         \
                                                                                                                                  \
                                                              --tls-ca-cert                state-integration/tls/edge/ca.crt      \
[...]

PR #300, opened 2019-02-04
preserve query and fragment part of the URI while redirecting for relative paths (swagger doc)

KtorZ:
<p align="right">#299</p>

# Overview

- [ ] I have modified the `pathname` rather than the whole `href` to preserve URI fragment (and query params) 
- [ ] This still works us around the access of assets using relative path and, preserve the quite useful deep-linking feature of the API doc 

# Comments

This is roughly what a `location` object looks like in the DOM. So, modifying the `href` was a bit wrong in the sense that it completely breaks the whole `fragment` used by the underlying React app to route the request and perform deeplinking accordingly.

```js
{ hash: ""
, host: "github.com"
, hostname: "github.com"
, href: "https://github.com/input-output-hk/cardano-wallet/compare/KtorZ/299/deeplink-api-doc?expand=1"
, origin: "https://github.com"
, pathname: "/input-output-hk/cardano-wallet/compare/KtorZ/299/deeplink-api-doc"
, port: ""
, protocol: "https:"
```

### DIFF
--- a/src/Cardano/Wallet/API/V1/Swagger.hs
+++ b/src/Cardano/Wallet/API/V1/Swagger.hs
@@ -1061,8 +1061,8 @@ swaggerSchemaUIServer =
     <script>
         // Force Strict-URL Routing for assets relative paths
         (function onload() {
-            if (!window.location.href.endsWith("/")) {
-                window.location.href += "/";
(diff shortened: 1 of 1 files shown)


Anviking (approved):
Nice! Works for me.

PR #303, opened 2019-02-05
[WIP] Checks occasional CI failure

paweljakubas:
<p align="right">#</p>

# Overview

**DO NOT MERGE**
**DO NOT REVIEW**

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/DB/Spec/Read.hs
+++ b/src/Cardano/Wallet/Kernel/DB/Spec/Read.hs
@@ -34,19 +34,21 @@ import qualified Cardano.Wallet.Kernel.Util.Core as Core
   defined in the Wallet Spec
 -------------------------------------------------------------------------------}
 
+import qualified Debug.Trace as Debug
+
 -- | Available UTxO
 --
 -- The available UtxO is the current UTxO minus outputs spent by pending txs
 cpAvailableUtxo :: IsCheckpoint c => c -> Core.Utxo
 cpAvailableUtxo c =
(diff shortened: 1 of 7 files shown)


src/Cardano/Wallet/Kernel/DB/Spec/Read.hs
```diff
@@ -34,14 +34,16 @@ import qualified Cardano.Wallet.Kernel.Util.Core as Core
...
 -- | Available UTxO
 --
 -- The available UtxO is the current UTxO minus outputs spent by pending txs
 cpAvailableUtxo :: IsCheckpoint c => c -> Core.Utxo
 cpAvailableUtxo c =
-    Core.utxoRemoveInputs (c ^. cpUtxo) pendingIns
+    Debug.trace (("####### cpAvailableUtxo: cpUtxo:" <> show (c ^. cpUtxo))::String) $ Core.utxoRemoveInputs (c ^. cpUtxo) pendingIns
   where
-    pendingIns = Pending.txIns (c ^. cpPending)
+    pendingIns = Debug.trace (("####### cpAvailableUtxo: pendingIns:" <> show (Pending.txIns (c ^. cpPending)))::String) $  Pending.txIns (c ^. cpPending)
```
KtorZ: @paweljakubas small tip: when doing trace like this with `Universum`, you can combine strings with `++` instead of the Semigroup combinator `<>`. Then GHC automatically infers it's a `String` (or more exactly a `[Char]`) and there's not need for type annotation. It's not much, but it personally makes my life easier sometimes :smile:

paweljakubas:
This was only to make sure ocassional CI failure is not recurring

PR #304, opened 2019-02-05
Update operation, for EOS-wallets.

denisshevchenko:
<p align="right">#36</p>

# Overview

- [x] I have added `Update`-operation for EOS-wallets. It allows to update name, assurance level and address pool gap.

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/Accounts.hs
+++ b/src/Cardano/Wallet/Kernel/Accounts.hs
@@ -2,6 +2,7 @@ module Cardano.Wallet.Kernel.Accounts (
       createAccount
     , deleteAccount
     , updateAccount
+    , updateAccountGap
     -- * Errors
     , CreateAccountError(..)
     ) where
@@ -19,14 +20,16 @@ import           Data.Acid (update)
 import           Pos.Core.NetworkMagic (makeNetworkMagic)
 import           Pos.Crypto (EncryptedSecretKey, PassPhrase)
 
+import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 import           Cardano.Wallet.Kernel.DB.AcidState (CreateHdAccount (..), DB,
-                     DeleteHdAccount (..), UpdateHdAccountName (..))
+                     DeleteHdAccount (..), UpdateHdAccountGap (..),
+                     UpdateHdAccountName (..))
 import           Cardano.Wallet.Kernel.DB.HdRootId (HdRootId)
 import           Cardano.Wallet.Kernel.DB.HdWallet (AccountName (..),
                      HdAccount (..), HdAccountBase (..), HdAccountId (..),
                      HdAccountIx (..), HdAccountState (..),
                      HdAccountUpToDate (..), UnknownHdAccount (..),
-                     hdAccountName)
+                     UpdateGapError (..), hdAccountName)
 import           Cardano.Wallet.Kernel.DB.HdWallet.Create
                      (CreateHdAccountError (..), initHdAccount)
 import           Cardano.Wallet.Kernel.DB.HdWallet.Derivation
@@ -168,3 +171,13 @@ updateAccount hdAccountId newAccountName pw = do
     return $ case res of
          Left dbError        -> Left dbError
          Right (db, account) -> Right (db, account)
+
+-- | Updates address pool gap in an HD 'Account' (in EOS-wallet).
+updateAccountGap
+    :: HdAccountId
+    -> AddressPoolGap
+    -- ^ The new adress pool gap for this account.
+    -> PassiveWallet
+    -> IO (Either UpdateGapError (DB, HdAccount))
+updateAccountGap hdAccountId newGap pw = liftIO $
+    update (pw ^. wallets) (UpdateHdAccountGap hdAccountId newGap)
--- a/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
+++ b/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
@@ -77,6 +77,7 @@ module Cardano.Wallet.Kernel.DB.HdWallet (
   , UnknownHdRoot(..)
   , UnknownHdAccount(..)
   , UnknownHdAddress(..)
+  , UpdateGapError(..)
   , embedUnknownHdRoot
   , embedUnknownHdAccount
     -- * Zoom to parts of the HD wallet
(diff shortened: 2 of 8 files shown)


src/Cardano/Wallet/Kernel/Accounts.hs
```diff
@@ -168,3 +171,16 @@ updateAccount hdAccountId newAccountName pw = do
...
+-- | Updates address pool gap in an HD 'Account' (in EOS-wallet).
+updateAccountGap
+    :: HdAccountId
+    -> AddressPoolGap
+    -- ^ The new adress pool gap for this account.
+    -> PassiveWallet
+    -> IO (Either UnknownHdAccount (DB, HdAccount))
+updateAccountGap hdAccountId newGap pw = do
+    res <- liftIO $ update (pw ^. wallets) (UpdateHdAccountGap hdAccountId newGap)
+    return $ case res of
```
uroboros: `return res` ? ;)

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -481,6 +483,19 @@ hdAccountBaseId = lens getHdAccountId setHdAccountId
...
+hdAccountBaseGap :: Lens' HdAccount AddressPoolGap
+hdAccountBaseGap = hdAccountBase . lens getHdAccountGap setHdAccountGap
+  where
+    getHdAccountGap :: HdAccountBase -> AddressPoolGap
+    -- This lens is only for 'HdAccount' with EO-branch, so if it will apply
+    -- to 'HdAccount' with FO-branch - just return default value of gap.
+    getHdAccountGap (HdAccountBaseFO _)       = def :: AddressPoolGap
+    getHdAccountGap (HdAccountBaseEO _ _ gap) = gap
+
+    setHdAccountGap :: HdAccountBase -> AddressPoolGap -> HdAccountBase
```
uroboros: more lens here?

`setHdAccountGap base@(HdAccountBaseEO _ _ _) newGap = base ^. hdAccountBaseEOAddressPoolGap .~ newGap`

`setHdAccountGap base _ = base`

uroboros (commented):
mostly LGTM

src/Cardano/Wallet/Kernel/Accounts.hs
denisshevchenko: Ops. :-) You're right! Copy-n-paste is EVIL!

src/Cardano/Wallet/Kernel/DB/HdWallet/Update.hs
```diff
@@ -38,3 +40,10 @@ updateHdAccountName :: HdAccountId
 updateHdAccountName accId name = do
     zoomHdAccountId identity accId $ do
         modifyAndGetNew $ hdAccountName .~ name
+
+updateHdAccountGap :: HdAccountId
+                   -> AddressPoolGap
+                   -> Update' UnknownHdAccount HdWallets HdAccount
+updateHdAccountGap accId gap =
+    zoomHdAccountId identity accId $
+        modifyAndGetNew $ hdAccountBaseGap .~ gap
```
KtorZ: I'd actually be more in favor of a loud and explicit failure in case we attempt to update the gap of an HdAccount that doesn't refer to an externally owned account. We don't want to be silently swallowing this but report the failure as early as possible.

Similarly, we _should_ probably not want a lens like `hdAccountBaseGap` precisely because, it's in essence, a partial function we just happen to default in the case of the FO branch. It does make sense to define a lens like `hdRootId` because this exists on both branch, but for something like the gap, we definitely want to pattern match and fail in case we happen to be on the wrong branch.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -194,6 +197,34 @@ updateWallet wallet wId (V1.WalletUpdate v1Level v1Name) = runExceptT $ do
...
+    -> V1.UpdateEosWallet
+    -> m (Either UpdateEosWalletError V1.EosWallet)
+updateEosWallet wallet wId (V1.UpdateEosWallet v1Level v1Name newGap) = runExceptT $ do
+    rootId <- withExceptT UpdateEosWalletWalletIdDecodingFailed $ fromRootId wId
+    (db, newHdRoot) <- withExceptT UpdateEosWalletError $ ExceptT $ liftIO $
+        Kernel.updateHdWallet wallet rootId newLevel newName
+    -- 'HdRoot' doesn't contain address pool gap, only corresponding 'HdAccount's contain it.
+    -- So we have to update all these accounts.
+    accounts <- fmap (IxSet.toList) $ withExceptT UpdateEosWalletError $ exceptT $
+        return $ Kernel.accountsByRootId db rootId
```
KtorZ, line 214: So, I had this feeling today when writing the integration tests for the externally owned wallet and I believe we should hand-over to the DB layer the invariant checks on the eos accounts to make sure that, whatever we get from the db is actually valid data. And similarly, we want to check that whatever we put in the db is valid all-the-same.

KtorZ: I'll actually work on that and provide a PR ASAP.

KtorZ (changes requested):

src/Cardano/Wallet/Kernel/DB/HdWallet/Update.hs
```diff
@@ -38,3 +40,16 @@ updateHdAccountName :: HdAccountId
...
+                   -> Update' UnknownHdAccount HdWallets HdAccount
+updateHdAccountGap accId gap =
+    zoomHdAccountId identity accId $ do
+        acc <- get
+        case acc ^. hdAccountBase of
+            HdAccountBaseEO _ pKey _ ->
+                modifyAndGetNew $ hdAccountBase .~ HdAccountBaseEO accId pKey gap
+            HdAccountBaseFO _ ->
+                -- It's an error: we try to update a gap in account with FO-branch.
+                throwError $ UnknownHdAccount accId
```
KtorZ: This is rather misleading as an error because, the account actually exists. The problem is that it's not what we expected it to be. Imagine debugging this, we see an error `UnknownHdAccount` and start trying to figure out why this account doesn't exist despite having created it before. 
What we really want here is a clear signal that, we are trying to perform an invalid operation on a fully owned account.

src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -47,6 +47,7 @@ import           Cardano.Wallet.Kernel.DB.HdWallet (AssuranceLevel,
                      WalletName, hdRootId)
 import qualified Cardano.Wallet.Kernel.DB.HdWallet as HD
 import qualified Cardano.Wallet.Kernel.DB.HdWallet.Create as HD
+-- import qualified Cardano.Wallet.Kernel.DB.HdWallet.Read as HD
```
KtorZ: Why is this commented out :thinking: ? Why not remove it entirely ?

KtorZ (changes requested):

src/Cardano/Wallet/Kernel/DB/HdWallet/Update.hs
denisshevchenko: Yes, this is just a first code that compiles. :-)

src/Cardano/Wallet/Kernel/Wallets.hs
denisshevchenko: Ops, forgot.. Removed.

src/Cardano/Wallet/Kernel/DB/HdWallet/Update.hs
denisshevchenko: @KtorZ What is the best solution here: add new error in `UnknownHdAccount` type, or introduce new error type?

KtorZ: You tell me :)?

denisshevchenko: Done.

src/Cardano/Wallet/Kernel/DB/HdWallet/Update.hs
denisshevchenko: Done.

KtorZ (approved):

PR #307, opened 2019-02-05
Foundations for E.O. integration tests

KtorZ:
<p align="right">#236</p>

# Overview

- [x] I have added the necessary foundation for testing creation of externally owned wallet
    - This includes a new section in the fixture to generate a list of account public keys and their corresponding indexes from an initial backup phrase
    - Also includes the necessary lenses to extend the integration tests DSL

# Comments

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -122,6 +122,8 @@ module Cardano.Wallet.API.V1.Types (
   , CaptureAccountId
   -- * Core re-exports
   , Core.Address
+  , Core.PublicKey
+  , Core.PassPhrase
   -- * Wallet Errors
   , WalletError(..)
   , ErrNotEnoughMoney(..)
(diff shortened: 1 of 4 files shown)


src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1111,6 +1113,12 @@ instance Bounded AccountIndex where
     minBound = AccountIndex 2147483648
     maxBound = AccountIndex maxBound
 
+instance Enum AccountIndex where
+    fromEnum (AccountIndex a) = fromEnum a
+    toEnum a = case mkAccountIndex (toEnum a) of
```
uroboros, line 1118: I can't quite make sense of that 2nd toEnum?

uroboros (approved):
just one question - otherwise LGTM

src/Cardano/Wallet/API/V1/Types.hs
KtorZ: We have this `toEnum :: Int -> a` and `mkAccountIndex :: Word32 -> Either ... AccountIndex`, so the 2nd `toEnum` here is to go from `Int` to `Word32` :)

uroboros: ahhh thanks ;)

PR #309, opened 2019-02-05
Foundation Setup For Manutal Test Scripting

KtorZ:
<p align="right">#258</p>

# Overview

- [ ] I have added `docopt` as a vendor dependency in the `test/manual/scripts`
- [ ] I have added two examples for CLI bash scripts that can be written using `docopt`, making shell script a bit less of a mess.

# Comments


test/manual/scripts/docopts.sh
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env bash
```
KtorZ, line 1: This is imported from the docopt repository, I couldn't resort to have a submodule here, so I just copy/pasted the script and added a LICENSE header.

test/manual/scripts/generate-addresses.sh
```diff
@@ -0,0 +1,30 @@
...
+do
+  >&2 echo "Address no: $i"
+	ADDRESS=$(curl -kX POST https://localhost:${ARGS[--port]}/api/v1/addresses \
+  		-H "Accept: application/json; charset=utf-8" \
+  		-H "Content-Type: application/json; charset=utf-8" \
+  		--cert ${ARGS[--client]} \
+  		-d "{
+		  \"accountIndex\": ${ARGS[--account]},
+		  \"walletId\": \"${ARGS[--wallet]}\"
+      }" | jq .data.id)
```
piotr-iohk, line 28: `jq .data.id` - that's cool! :+1:

piotr-iohk (approved):
:+1: :+1: :+1:

test/manual/scripts/generate-addresses.sh
KtorZ: Not sure if `jq` is installed by default on most Mac OS and Linux distributions but that's definitely a must-have when working with JSON :sweat_smile:

PR #310, opened 2019-02-05
Fix `getEosWallet` + Add some debugging capabilities to the Kernel

KtorZ:
<p align="right">#236</p>

# Overview

- [ ] I have added some basic logging helpers to the Kernel internals
- [ ] I have added some debugging code to the 'getEosWallet' function
- [ ] I have fixed a partial function access which caused the `getEosWallet` function to throw a wild exception

# Comments

Note that, I think we should keep the debugging code for now and pretty much, even after we shipped this. I'll add more as I write more integration tests and I'll become quite handy when debugging situations from either QA testing or, our users.

### DIFF
--- a/src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
+++ b/src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
@@ -286,9 +286,10 @@ addressPoolGapByRootId rootId db = do
     if | null bases            -> Left $ GetEosWalletErrorNoAccounts anId
        | not . null $ errors   -> Left $ GetEosWalletErrorWrongAccounts anId
        | length (nub gaps) > 1 -> Left $ GetEosWalletErrorGapsDiffer anId
-       | otherwise             -> let [gap] = gaps in Right gap
+       | otherwise             -> let gap:_ = gaps in Right gap
   where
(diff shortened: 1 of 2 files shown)


src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -281,10 +286,11 @@ addressPoolGapByRootId rootId db = do
     if | null bases            -> Left $ GetEosWalletErrorNoAccounts anId
        | not . null $ errors   -> Left $ GetEosWalletErrorWrongAccounts anId
        | length (nub gaps) > 1 -> Left $ GetEosWalletErrorGapsDiffer anId
-       | otherwise             -> let [gap] = gaps in Right gap
+       | otherwise             -> let gap:_ = gaps in Right gap
```
KtorZ, line 289: Didn't quite spot that one in the review :upside_down_face: ... But, gaps are only `nubs` in the branch before, so here, they are still a full list and `[gap]` doesn't pattern match well with more than one account.

KtorZ:
@denisshevchenko Did you got time to look into this ?

uroboros (approved):
LGTM

PR #311, opened 2019-02-05
initial tests for EOSWALLETS_CREATE capabilities

KtorZ:
<p align="right">#236</p>

# Overview

- [ ] I have added a few initial scenarios for externally-owned wallet creations
- [ ] I have set one of those tests as "pending" because failing. Will add the corresponding guard later and enable the pending scenario.

# Comments

Note that tests currently fail and require the fix in #310 in order to pass :+1:

### DIFF
--- a/cardano-wallet.cabal
+++ b/cardano-wallet.cabal
@@ -81,7 +81,6 @@ library
     , filepath
     , foldl
     , formatting
-    , formatting
     , generics-sop
     , http-api-data
     , http-client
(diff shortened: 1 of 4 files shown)


piotr-iohk:
lgtm :+1: 

We'll need to add capabilities for EOS Wallets into https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk/edit
(and expand this suite)

piotr-iohk (approved):
:+1:

PR #312, opened 2019-02-06
Matt/156 persistent

parsonsmatt:
<p align="right">156#</p>

# Overview

- [x] I have ported the Beam logic to `persistent` and `esqueleto` in a backwards-compatible manner that allows a direct comparison of the two libraries.

# Comments

The `beam` code used a SQL `UNION` operator to filter on addresses. This is not supported by `esqueleto`. Instead, I used a pair of `INNER JOIN`s to get the `inputs` and `outputs` working.


src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -35,740 +29,4 @@ module Cardano.Wallet.Kernel.DB.Sqlite (
...
-getInputsTable conn =
-    runBeamSqlite conn $ SQL.runSelectReturningList $ SQL.select $
-        SQL.all_ $ _mDbInputs metaDB
-
-getOutputsTable :: Sqlite.Connection
-                -> IO [TxOutput]
-getOutputsTable conn =
-    runBeamSqlite conn $ SQL.runSelectReturningList $ SQL.select $
-        SQL.all_ $ _mDbOutputs metaDB
+import Cardano.Wallet.Kernel.DB.Sqlite.Persistent
```
parsonsmatt: To preserve backwards compatibility with the rest of the codebase, I kept the interface the same and moved all the code into a different module. I then reimplemented the new code with the same interface, and did the swap once all the types were checking out OK.

src/Cardano/Wallet/Kernel/DB/Sqlite/Beam.hs
```diff
@@ -0,0 +1,769 @@
+{-# LANGUAGE DeriveGeneric      #-}
```
parsonsmatt, line 1: This module contains the original code, unchanged.

src/Cardano/Wallet/Kernel/DB/Sqlite/Persistent/Orphans.hs
```diff
@@ -0,0 +1,78 @@
...
+    toPersistValue =
+        toPersistValue . sformat Core.addressF
+    fromPersistValue f = do
+        addr <- Core.decodeTextAddress <$> fromPersistValue f
+        case addr of
+           Left _  -> Left $ "Failed to parse Address out of PersistValue: " <> show f
+           Right a -> pure a
+
+instance Read Core.Address where
+    readsPrec _ = error "Needed for HttpApiData"
```
parsonsmatt, line 85: This is... dubious, but it's "fine" as long as we never actually call `Read` on this. The `Show` instance is not derived, so it's borked.

src/Cardano/Wallet/Kernel/DB/Sqlite/Persistent.hs
```diff
@@ -0,0 +1,612 @@
...
+import           Cardano.Wallet.Kernel.DB.Sqlite.Persistent.Orphans ()
+import           Cardano.Wallet.Kernel.DB.TxMeta.Types (AccountFops (..),
+                     FilterOperation (..), Limit (..), Offset (..),
+                     SortCriteria (..), SortDirection (..), Sorting (..))
+import qualified Cardano.Wallet.Kernel.DB.TxMeta.Types as Kernel
+
+import qualified Pos.Chain.Txp as Txp
+import qualified Pos.Core as Core
+
+share [ mkPersist sqlSettings { mpsPrefixFields = False } , mkMigrate "migrateAll" ] [persistLowerCase|
```
parsonsmatt, line 72: This quasiquoter is what `persistent` uses to define the entity datatypes, infer migrations, and create all the query helpers and fields.

src/Cardano/Wallet/Kernel/DB/Sqlite/Persistent.hs
```diff
@@ -0,0 +1,612 @@
...
+TxMeta sql=tx_metas
+    txMetaTableId         Txp.TxId       sql=meta_id
+    txMetaTableAmount     Core.Coin      sql=meta_amount
+    txMetaTableCreatedAt  Core.Timestamp sql=meta_created_at
+    txMetaTableIsLocal    Bool           sql=meta_is_local
+    txMetaTableIsOutgoing Bool           sql=meta_is_outgoing
+    txMetaTableWalletId   Core.Address   sql=meta_wallet_id
+    txMetaTableAccountIx  Word32         sql=meta_account_ix
+
+    Primary txMetaTableId txMetaTableWalletId txMetaTableAccountIx
```
parsonsmatt, line 83: These primary keys are pulled from the original schema. In reality, this should be more of a unique key, and the actual *primary* key is the `txMetaTableId` field. That's the field that the other models use to refer to the `meta`.

src/Cardano/Wallet/Kernel/DB/Sqlite/Persistent.hs
```diff
@@ -0,0 +1,612 @@
...
+
+TxOutput sql=tx_metas_outputs
+    outputTableTxId    Txp.TxId     sql=meta_id
+    outputTableIndex   Word32       sql=output_index
+    outputTableAddress Core.Address sql=output_address
+    outputTableCoin    Core.Coin    sql=output_coin
+
+    Primary outputTableTxId outputTableIndex
+
+    deriving Show Generic
```
parsonsmatt, line 95: We can get automatic foreign keys if we used the `txMetaTableId` as the primary key/ID: https://stackoverflow.com/questions/30062707/foreign-key-constraints-in-yesod-persistent

Defining the `outputTableTxId` as having the type `outputTableTxId TxMetaId sql=meta_id` would automatically create the foreign key that we want.

src/Cardano/Wallet/Kernel/DB/Sqlite/Persistent.hs
```diff
@@ -0,0 +1,612 @@
...
+-- | Migrates the 'MetaDB', failing with an IO exception in case this is not
+-- possible.
+unsafeMigrateMetaDB :: SqlBackend -> IO ()
+unsafeMigrateMetaDB conn = do
+    unsafeRunPersistConn conn (void (runMigrationSilent migrateAll))
+
+-- | Simply a conveniency wrapper to avoid 'Kernel.TxMeta' to explicitly
+-- import Sqlite modules.
+newConnection :: FilePath -> IO SqlBackend
+newConnection fp = open' (mkSqliteConnectionInfo (Text.pack fp)) (\_ _ _ _ -> pure ())
```
parsonsmatt, line 275: There's a slight change to the interface of the module -- this function returns a `SqlBackend` instead of a raw SQLite `Connection`. The `SqlBackend` provides many features, like query memoization and statement finalization. Since this shouild *really* be an opaque type, and doesn't break anything else, I don't mind this interface change.

test/unit/Test/Spec/TxMetaStorage.hs
```diff
@@ -175,7 +175,9 @@ spec = do
                 let ins = testMeta ^. txMetaInputs
                 let doubleSpendTx = testMeta {_txMetaInputs = ins <> ins}
                 putTxMetaT hdl doubleSpendTx `shouldThrow`
-                    (\(StorageFailure _) -> True)
+                    (\case
+                        StorageFailure _ -> True
+                        _ -> False)
```
parsonsmatt, line 180: This would fail with a `PatternMatchError` if it wasn't that constructor.

KtorZ:
Thanks for the PR @parsonsmatt :pray: 
I won't likely have time to look at it this week to be honest. I just skimmed through it and I like the approach already.

KtorZ:
Closing the PR for now. Keeping the branch for insights. We are most likely going under a rewrite of the wallet soon. This may come in handy.

PR #314, opened 2019-02-06
Retry when checking node api connection

Anviking:
<p align="right">#87</p>

# Context

When starting a wallet we would see
```
[2019-02-06 11:08:25.49 UTC] There was an error connecting to the node: ErrFromServant (ConnectionError "HttpExceptionRequest Request {\n  host                 = \"127.0.0.1\"\n  port                 = 8080\n  secure               = True\n  requestHeaders       = [(\"Accept\",\"application/json;charset=utf-8,application/json\")]\n  path                 = \"/api/v1/node-settings\"\n  queryString          = \"\"\n  method               = \"GET\"\n  proxy                = Nothing\n  rawBody              = False\n  redirectCount        = 10\n  responseTimeout      = ResponseTimeoutDefault\n  requestVersion       = HTTP/1.1\n}\n (InternalException (HostCannotConnect \"127.0.0.1\" [Network.Socket.connect: <socket: 49>: does not exist (Connection refused)]))")
```

In the logs from https://github.com/input-output-hk/cardano-wallet/blob/3d769cff56bc754d2355cdb8369858f67e2ff892/src/Cardano/Wallet/Server/Plugins.hs#L103-L111

as the server hadn't started yet.

# Overview

- [x] I have exported `retrying` in the `NodeStateAdaptor` module
- [x] Used this `retrying` when checking node client connection on startup
- [x] We no longer get an error logged about the node connection having failed

# Comments

### DIFF
--- a/src/Cardano/Wallet/Server/Plugins.hs
+++ b/src/Cardano/Wallet/Server/Plugins.hs
@@ -51,6 +51,7 @@ import qualified Cardano.Wallet.API.V1.Types as V1
 import           Cardano.Wallet.Kernel (DatabaseMode (..), PassiveWallet)
 import qualified Cardano.Wallet.Kernel.Diffusion as Kernel
 import qualified Cardano.Wallet.Kernel.Mode as Kernel
+import qualified Cardano.Wallet.Kernel.NodeStateAdaptor as NodeStateAdaptor
 import qualified Cardano.Wallet.Server as Server
 import           Cardano.Wallet.Server.CLI (WalletBackendParams (..),
                      WalletBackendParams (..), walletAcidInterval,
@@ -101,7 +102,7 @@ apiServer
     env <- ask
     let diffusion' = Kernel.fromDiffusion (lower env) diffusion
     logInfo "Testing node client connection"
-    eresp <- liftIO . runExceptT $ NodeClient.getNodeSettings nodeClient
+    eresp <- liftIO . NodeStateAdaptor.retrying . runExceptT $ NodeClient.getNodeSettings nodeClient
     case eresp of
         Left err -> do
             logError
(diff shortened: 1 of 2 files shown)


src/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -101,7 +102,7 @@ apiServer
     env <- ask
     let diffusion' = Kernel.fromDiffusion (lower env) diffusion
     logInfo "Testing node client connection"
-    eresp <- liftIO . runExceptT $ NodeClient.getNodeSettings nodeClient
+    eresp <- liftIO . NodeStateAdaptor.retrying . runExceptT $ NodeClient.getNodeSettings nodeClient
```
Anviking, line 105: This *does* feel a bit strange.

Making the check internal to `NodeStateAdaptor` wouldn't make sense as the `NodeStateAdaptor` is — almost — just an interface that doesn't care about the node client.

Maybe
1. The retrying things (`toMonadIO`) I put in `NodeStateAdaptor` should be moved to like the `NodeClient`?
2.  ~`NodeStateAdaptor ` should get a proper `lift`  function — why not?~ Maybe `NodeHttpClient` should have `hoist`

KtorZ: I am going to need slightly more context here to grasp what you mean and what problem you're trying to solve in this PR :no_mouth:

Anviking: Updated pr description.

Anviking: What about putting the retrying inside the node-client-monad itself? (see #317 , though maybe it is not what we want?)

KtorZ (approved):

PR #317, opened 2019-02-06
Alternative Retry when checking node api connection

Anviking:
<p align="right">#87</p>

# Context

Alternative, more chaotic version of #314 

When starting a wallet we would see
```
[2019-02-06 11:08:25.49 UTC] There was an error connecting to the node: ErrFromServant (ConnectionError "HttpExceptionRequest Request {\n  host                 = \"127.0.0.1\"\n  port                 = 8080\n  secure               = True\n  requestHeaders       = [(\"Accept\",\"application/json;charset=utf-8,application/json\")]\n  path                 = \"/api/v1/node-settings\"\n  queryString          = \"\"\n  method               = \"GET\"\n  proxy                = Nothing\n  rawBody              = False\n  redirectCount        = 10\n  responseTimeout      = ResponseTimeoutDefault\n  requestVersion       = HTTP/1.1\n}\n (InternalException (HostCannotConnect \"127.0.0.1\" [Network.Socket.connect: <socket: 49>: does not exist (Connection refused)]))")
```

In the logs from https://github.com/input-output-hk/cardano-wallet/blob/3d769cff56bc754d2355cdb8369858f67e2ff892/src/Cardano/Wallet/Server/Plugins.hs#L103-L111

as the server hadn't started yet.

# Overview

- [x] I have added `hoist` to `NodeClient`
- [x] I have put retrying in the NodeClient itself that is returned from `setupNodeClient `

# Comments

A side effect is that when the wallet is forwarding a request to the node, like 

https://github.com/input-output-hk/cardano-wallet/blob/dd1a28204136e3b9b73f00fe8b4ea48ce7d71a13/src/Cardano/Wallet/API/V1/Handlers/Info.hs#L21

we get the retrying here too.

### DIFF
--- a/src/Cardano/Wallet/Server/Plugins.hs
+++ b/src/Cardano/Wallet/Server/Plugins.hs
@@ -20,6 +20,9 @@ module Cardano.Wallet.Server.Plugins
 
 import           Universum
 
+import           Control.Retry (RetryPolicyM, RetryStatus, fullJitterBackoff,
+                     limitRetries)
+import qualified Control.Retry
 import           Data.Acid (AcidState)
 import           Data.Aeson (encode)
 import qualified Data.ByteString.Char8 as BS8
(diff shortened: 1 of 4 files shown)


src/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -237,4 +240,18 @@ setupNodeClient (serverHost, serverPort) params = liftIO $ do
         walletClient :: NodeHttpClient
         walletClient = NodeClient.mkHttpClient baseUrl manager
 
-    return walletClient
+    return $ NodeClient.hoist retryingExcept walletClient
```
Anviking, line 243: Here is the hoist

KtorZ:
Went for #314 :+1:

Issue #220, opened 2019-01-14
Integration tests fail with expectTxStatusEventually: waited too long for statuses.


piotr-iohk:
| Release | Operating System       | Cause |
| ---     | ---                    | ---   |
| next    | Windows & OSX & Linux | Unknown |

# Context
 
Recently I started to observe that my integration scenarios randomly tend to fail on the scenarios which are dealing with wallets that have some coins. The following error is observed:

```
       uncaught exception: IOException of type UserError
       user error (expectTxStatusEventually: waited too long for statuses.)
```

I observe that the issue is more likely to happen when there is more than one wallet *with some coins* used in the scenario, e.g. fails always on such setup code:

```
        forM_ matrix $ \(endpoint, expectations) -> scenario endpoint $ do
            forM_ ([3,6,6,9,0]) $ \coins -> do
                setup $ defaultSetup
                    & initialCoins .~ [coins]
```

The failing expectation `expectTxStatusEventually` that is used in the `setupWallet` waits for the transaction to have `expectTxStatusEventually [InNewestBlocks, Persisted] txn` after the wallet is fed with some coins. The method has a timeout of 60s, but even changing it to 600s does not help for the setup code above.

Also occasionally I saw other tests failing due to this issue (ofc only tests working with wallets with coins), also then changing the timeout did not help. The situation with other tests was strange as at some point they started to work properly again.

It seems that occasionally the tests may fail also on the CI: https://travis-ci.org/input-output-hk/cardano-wallet/jobs/478587172#L894

 

# Steps to Reproduce

1. Following code put in the `scenario` will most likely result in the error:

```
        forM_ matrix $ \(endpoint, expectations) -> scenario endpoint $ do
            forM_ ([3,6,6,9,0]) $ \coins -> do
                setup $ defaultSetup
                    & initialCoins .~ [coins]
```

## Expected behavior

Tests should be passing.

## Actual behavior
 
<A clear and concise description of what you observe instead. If applicable add screenshots to help explain your problem.>

Tests fail with:

```
       uncaught exception: IOException of type UserError
       user error (expectTxStatusEventually: waited too long for statuses.)
```

---

# Resolution Plan

So, it appears that some transactions are flagged as "WontApply" (which is a non-transient status) by the backend and therefore, the DSL keeps on polling transactions hoping for the tx to reach one of the expected status; but it won't, regardless of how long we wait. 

[...]

PR #318, opened 2019-02-06
[220] Add NewTransactionZeroAmountCoin

paweljakubas:
[220] Finishing error

<p align="right">#220</p>

# Overview

- [x] I have added ZeroAmountOutput error together with tests

# Comments

It was seen that one cannot construct transactions with zero amount output as  in the current version it can incur long response times. The decision was to forbid it on the level of wallet by introduction of respective error. When it will be supported underneath it can be revoked though (eg. during Shelley era)

### DIFF
--- a/src/Cardano/Wallet/API/V1/Swagger.hs
+++ b/src/Cardano/Wallet/API/V1/Swagger.hs
@@ -336,6 +336,7 @@ $errors
     -- 'UnsupportedMimeTypeError'
     , mkRow fmtErr $ UnsupportedMimeTypePresent "Expected Content-Type's main MIME-type to be 'application/json'."
     , mkRow fmtErr $ UtxoNotEnoughFragmented (ErrUtxoNotEnoughFragmented 1 msgUtxoNotEnoughFragmented)
+    , mkRow fmtErr $ ZeroAmountCoin (ErrZeroAmountCoin 1 msgZeroAmountCoin)
     -- TODO 'MnemonicError' ?
     ]
   mkRow fmt err = T.intercalate "|" (fmt err)
@@ -682,6 +683,7 @@ curl -X POST https://localhost:8090/api/v1/transactions \
   "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
 }'
 ```
+One has to make sure that each amount field in destinations is positive.
 
 
 About UTXO Fragmentation
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE ConstraintKinds            #-}
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE DuplicateRecordFields      #-}
 {-# LANGUAGE ExplicitNamespaces         #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE KindSignatures             #-}
@@ -129,6 +130,8 @@ module Cardano.Wallet.API.V1.Types (
   , ErrNotEnoughMoney(..)
   , ErrUtxoNotEnoughFragmented(..)
   , msgUtxoNotEnoughFragmented
+  , ErrZeroAmountCoin(..)
+  , msgZeroAmountCoin
   , toServantError
   , toHttpErrorStatus
   , module Cardano.Wallet.Types.UtxoStatistics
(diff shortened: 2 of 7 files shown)


src/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -682,6 +683,8 @@ curl -X POST https://localhost:8090/api/v1/transactions \
   "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
 }'
 ```
+One has to make sure that each amount field in destinations must be positive.
+If anyone is set to zero (also in single recipient transaction) a `ZeroAmountCoin` error is triggered.
```
KtorZ: I'd only go for the first sentence here.

KtorZ (approved):

src/Cardano/Wallet/API/V1/Swagger.hs
paweljakubas: deleted second sentenced, rebased and squashed

KtorZ:
Actually. Why don't we even fix that at the API-level as-well ? 

https://input-output-hk.github.io/cardano-wallet/#tag/Transactions%2Fpaths%2F~1api~1v1~1transactions%2Fpost

I mean, here, if we make sure that the amount can't be negative or null then we even prevent such transactions from making it into the kernel.

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -2663,6 +2689,8 @@ instance ToServantError WalletError where
             err400 { errHTTPCode = 429 }
         UtxoNotEnoughFragmented{} ->
             err403
+        ZeroAmountCoin{} ->
+            err403
```
KtorZ: Also, just noticed that. This should be a `400`. `403` is usually reserved for lack of permissions on an authenticated client. 
Likely, our API will never return a 403 because we don't have any particular notion of access control layer (ACL) on our API.

KtorZ (changes requested):

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -2663,6 +2689,8 @@ instance ToServantError WalletError where
             err400 { errHTTPCode = 429 }
         UtxoNotEnoughFragmented{} ->
             err403
```
paweljakubas, line 2692: shouldn't here also be 400 instead of 403 ?

src/Cardano/Wallet/WalletLayer/Kernel/Active.hs
```diff
@@ -122,6 +123,7 @@ createUnsignedTx activeWallet grouping regulation payment = liftIO $ do
     res <- runExceptT $ do
         (opts, accId, payees) <- withExceptT NewTransactionWalletIdDecodingFailed $
             setupPayment policy grouping regulation payment
+        withExceptT NewTransactionZeroAmountCoinsDetected $ exceptT $ checkCoins payees
```
paweljakubas: I added here also API support to check against zero amount coins. Here, however when we use `pstUnsignedTransaction` we will end up with  `UnknownError "NewUnsignedTransactionError"` as these errors are not exposed up in **ReifyWalletError**

Moreover, we can also add catching these same error in `estimateFees`. Right now it bases on : 
```
data EstimateFeesError = EstFeesTxCreationFailed NewTransactionError
```

so it will descend to Kernel internals in order to grab this error

src/Cardano/Wallet/API/V1/Types.hs
KtorZ: For a lack of better error code, no, I believe this one is okay here.

test/integration/Test/Integration/Scenario/Transactions.hs
```diff
@@ -53,6 +54,52 @@ spec = do
...
+            & initialCoins .~ [1000000]
+
+        response <- request $ Client.getTransactionFee $- Payment
+            (defaultSource fixture)
+            (defaultDistribution 0 fixture)
+            defaultGroupingPolicy
+            noSpendingPassword
+
+        verify response
+            [ expectWalletError ( ZeroAmountCoin (Client.ErrZeroAmountCoin 1 Client.msgZeroAmountCoin))
```
paweljakubas, line 100: Here, is the the proof that `getTransactionFee` embraces wallet errors. But to do it it calls Kernel stuff. Which is in contrast to what happens when `postTransaction` or `postUnsignedTransaction` is called (though in the latter case it is not welcomed on the wallet errors board)

src/Cardano/Wallet/WalletLayer/Kernel/Active.hs
KtorZ: Hmm. That's not what I meant; here we are basically doing two times the same check but at different places. 
I meant that, instead of having a wallet coin amount here (https://github.com/input-output-hk/cardano-wallet/blob/b7dcca7a061e1fffe1ef86bd7a017e67daa14ceb/src/Cardano/Wallet/API/V1/Types.hs#L277-L301) ranging from 0 (included) to max coin value, it could range from 1 to max coin value.

Let's not bother with that now actually. We have more important stuff to deal with.

src/Cardano/Wallet/WalletLayer/Kernel/Active.hs
```diff
@@ -236,3 +238,16 @@ setupPayment policy grouping regulation payment = do
...
+    :: NonEmpty (Address, Coin)
+    -> Either Int ()
+checkCoins outputs =
+    let numberOfZeroAmountOutputs = length
+                                    $ NE.filter (== 0)
+                                    $ NE.map (getCoin . snd) outputs
+    in if numberOfZeroAmountOutputs > 0 then
+           Left numberOfZeroAmountOutputs
+       else
+           Right ()
```
KtorZ: This is duplicated with the above. We don't need to double-check that, or if we do, let's re-use functions and not duplicate them.

KtorZ (changes requested):

src/Cardano/Wallet/WalletLayer/Kernel/Active.hs
paweljakubas: So I opted for avoiding duplication. Just relying only on Kernel level checks and removing WalletLayer ones

KtorZ (approved):

Issue #221, opened 2019-01-14
Extend Integration Tests DSL to support ClientHttpError and returned HTTP code


piotr-iohk:
# Context

This ticket addresses ask for extending DSL: 
1. extend API DSL to support assertion for `ClientHttpError`
2. extend API DSL to support assertion for HTTP code returned by the API

Integration tests should not only cover happy path scenarios but also test that API behaves appropriately and returns correct error messages in case it is fed with faulty requests. (e.g. not supported API parameters and param values, param values out of bounds etc.) Often for such operations the error message is returned as `ClientHttpError` type (which actually exposes the `ServantError` type).

Currently there is no way to set expectations for the error message returned as such.
The ask is to extend API DSL to have the method similar to the one below:

```
expectClientHttpError
    :: (MonadIO m, MonadFail m, Show a)
    => String
    -> Either ClientError a
    -> m ()
expectClientHttpError excerpt = \case
    Right a -> wantedErrorButSuccess a
    Left (ClientHttpError (ServantError (DecodeFailure t r))) ->
        -- here I would like to get r.responseBody and make sure it contains excerpt
    Left e ->
        fail $ "expectClientHttpError: got something else than a ClientHttpError (ServantError (DecodeFailure t r)) failure: " <> show e
```

Additionally currently there is no way to set expectation for the API returned http code. API uses several codes for successful operations (`200`, `201`, `204`) and unsuccessful operations (`400`, `401`, `404` etc.) It would be nice to have a way for asserting for those codes in integration tests

# Decision

Extend the API DSL to allow setting expectations for the `ClientHttpError` error messages.
Extend the API DSL to allow setting expectations for the HTTP code returned by the API

# Acceptance Criterias

1. The Integration tests DSL _must_ allow setting expectations for error message returned as `ClientHttpError`
2. The Integration tests DSL _must_ allow setting expectations for HTTP code returned by the API

---

# Development Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #?       | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| ?        | -        |

---

# Retrospective

KtorZ:
Thanks for opening this ticket @piotr-iohk :pray:

PR #319, opened 2019-02-07
additional integration tests for ADDRESSES_IMPORT

piotr-iohk:
<p align="right">#</p>

# Overview

- [x] I have added capabilities for batch address import into https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk [ADDRESS_IMPORT] 
- [x] I intend to add missing integration tests covering - ADDRESS_IMPORT_04, 05, 06

# Comments

### DIFF
--- a/test/integration/Test/Integration/Scenario/Addresses.hs
+++ b/test/integration/Test/Integration/Scenario/Addresses.hs
@@ -6,6 +6,7 @@ import           Universum
 
 import           Cardano.Wallet.API.V1.Types (Account (accAddresses),
                      WalletAddress (..))
+import           Cardano.Wallet.Client.Http (ClientError, BatchImportResult)
 import qualified Cardano.Wallet.Client.Http as Client
 import           Test.Integration.Framework.DSL
 
@@ -92,7 +93,6 @@ spec = do
         response <- request $ Client.importAddresses
             $- fixture ^. wallet . walletId
             $- map (view address) addrs
-
         verify response
             [ expectFieldEqual totalSuccess 0
             , expectFieldEqual failures (map (view address) addrs)
@@ -101,19 +101,37 @@ spec = do
     scenario "ADDRESSES_IMPORT_04 - Can import addresses from different account into default account of the wallet" $ do
         fixture <- setup $ defaultSetup
 
-
         accountResp <- successfulRequest $ Client.postAccount
             $- (fixture ^. wallet . walletId)
             $- NewAccount
                 noSpendingPassword
                 "New Account"
-        print accountResp
+
         addrs <- sequence $ [mkAddress (fixture ^. backupPhrase) (Client.accIndex accountResp) 1]
 
         response <- request $ Client.importAddresses
             $- fixture ^. wallet . walletId
-            $- defaultAccountId
             $- addrs
         verify response
             [ expectFieldEqual totalSuccess 1
+            ]
+
+    scenario "ADDRESSES_IMPORT_05 - Returns error when wallet id is invalid" $ do
+        response <- unsafeRequest ("POST", "api/v1/wallets/aaa/addresses") $ Just $ [json|[
+            "ff45d33f04aeccae04d840c3f2dc60815bfd131a7fbd87f6823f0d8e"
+        ]|]
+        verify (response :: Either ClientError (BatchImportResult Text))
+            [ expectJSONError "Not a valid Cardano Address"
+            ]
+
+    scenario "ADDRESSES_IMPORT_06 - Returns error address is invalid" $ do
+        fixture <- setup $ defaultSetup
+
+        let endpoint = "api/v1/wallets/" <> fromWalletId (fixture ^. wallet . walletId) <> ("/addresses" :: Text)
+        response <- unsafeRequest ("POST", endpoint) $ Just $ [json|[
+          "dasd",
+          "1"
+          ]|]
+        verify (response :: Either ClientError (BatchImportResult Text))
+            [ expectJSONError "Not a valid Cardano Address"
             ]
--- a/test/integration/Test/Integration/Scenario/Addresses.hs
+++ b/test/integration/Test/Integration/Scenario/Addresses.hs
@@ -121,7 +121,8 @@ spec = do
             "DdzFFzCqrhssoca9zmsbhqHxJRjrDyzR1wh4Rs9ffbFTiYkcnDsYU416MYe2A29BFigVPBQgnkQH64et6pAqSjAqPPFbHcG1zR7G6kGr"
         ]|]
         verify (response :: Either ClientError (BatchImportResult Text))
-            [ expectWalletError (WalletNotFound)
+            [ expectError
+              -- TODO: add more expectations after #221 is resolved
             ]
 
     scenario "ADDRESSES_IMPORT_05 - Returns error when wallet id is valid but missing" $ do
--- a/test/integration/Test/Integration/Scenario/Addresses.hs
+++ b/test/integration/Test/Integration/Scenario/Addresses.hs
@@ -121,7 +121,7 @@ spec = do
             "DdzFFzCqrhssoca9zmsbhqHxJRjrDyzR1wh4Rs9ffbFTiYkcnDsYU416MYe2A29BFigVPBQgnkQH64et6pAqSjAqPPFbHcG1zR7G6kGr"
         ]|]
         verify (response :: Either ClientError (BatchImportResult Text))
-            [ expectWalletError (UnknownError "ImportAddressError")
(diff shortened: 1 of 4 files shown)


test/integration/Test/Integration/Scenario/Addresses.hs
```diff
@@ -73,9 +72,9 @@ spec = do
             [ expectListSizeEqual 2 -- NOTE 2 because there's also a default address
             ]
 
-    scenario "can't import addresses that are already present (used or unused)" $ do
+    scenario "ADDRESSES_IMPORT_03 - Can't import addresses that are already present (used or unused)" $ do
         -- NOTE
-        -- The fixture looks a bit complexe here but in the end, we should end
+        -- The fixture looks a bit complex here but in the end, we should end
```
KtorZ, line 76: :man_facepalming:

test/integration/Test/Integration/Scenario/Addresses.hs
```diff
@@ -101,3 +100,21 @@ spec = do
...
+                noSpendingPassword
+                "New Account"
+
+        response <- request $ Client.importAddresses
+            $- fixture ^. wallet . walletId
+            $- Client.accIndex accountResp
+            $- addrs
+        verify response
+            [ expectFieldEqual totalSuccess 1
+            ]
```
KtorZ, line 117: Not sure that this scenario does what you expect it to do :thinking: ? 
Actually, it's maybe a good idea to change a bit the signature of the `mkAddress` function here to also accept an account index. The implementation currently uses the default account index for all addresses made this way. So if we want to test importing addresses from many different accounts, that's probably the easiest way.

piotr-iohk: The scenario won't be relevant now, I suppose, hence we wno't be providing accountId in the `importAddress` enpoint. Unless, like you say... `mkAddress` with different account. :+1:

KtorZ:
#320 merged :+1:

KtorZ:
@piotr-iohk #320 merged, feel free to update :+1:

test/integration/Test/Integration/Scenario/Addresses.hs
```diff
@@ -93,8 +93,68 @@ spec = do
...
+
+    scenario "ADDRESSES_IMPORT_05 - Returns error when wallet id is invalid" $ do
+        response <- unsafeRequest ("POST", "api/v1/wallets/aaa/addresses") $ Just $ [json|[
+            "DdzFFzCqrhssoca9zmsbhqHxJRjrDyzR1wh4Rs9ffbFTiYkcnDsYU416MYe2A29BFigVPBQgnkQH64et6pAqSjAqPPFbHcG1zR7G6kGr"
+        ]|]
+        verify (response :: Either ClientError (BatchImportResult Text))
+            [ expectWalletError (WalletNotFound)
+            ]
```
piotr-iohk, line 126: @KtorZ, ^^^ the one above throws `ClientHttpError`, and the one below `WalletNotFound`...

piotr-iohk: @akegalj, maybe you could help me here, since @KtorZ is on vac?
After this change #327 I was expecting to get `expectWalletError (WalletNotFound)` when providing not existing / not valid walletID.

Instead, I'm getting `ClientHttpError` (see below) when providing arbitrary wallet id (i.e. like `aaa` or `123`) cf test `ADDRESSES_IMPORT_05 - Returns error when wallet id is invalid`: 


```
  1) Addresses ADDRESSES_IMPORT_05 - Returns error when wallet id is invalid
       expected: ClientWalletError WalletNotFound
        but got: ClientHttpError (FailureResponse (Response {responseStatusCode = Status {statusCode = 404, statusMessage = "Reference to an unexisting wallet was given."}, responseHeaders = fromList [("Transfer-Encoding","chunked"),("Date","Fri, 08 Feb 2019 16:00:06 GMT"),("Server","Warp/3.2.25"),("Content-Type","application/json")], responseHttpVersion = HTTP/1.1, responseBody = "{\"status\":\"error\",\"diagnostic\":{},\"message\":\"WalletNotFound\"}"}))
```

Interesingly when I provide wallet id of the wallet that was deleted I get `ClientWalletError WalletNotFound` (so `expectWalletError (WalletNotFound)` works). cf test `ADDRESSES_IMPORT_05 - Returns error when wallet id is valid but missing`

From the comment (https://github.com/input-output-hk/cardano-wallet/pull/327#discussion_r255101525) I understood that I should expect `ClientWalletError WalletNotFound` here but maybe I am mistaken? (the `ClientHttpError` actually contains message `WalletNotFound` so maybe that's just a matter of internal errors mapping and this is expected?, just want to make sure)

akegalj: If we are using `unsafeRequest` it is going to wrap our error in `ClientHttpError` if underlying response returned 404 (https://github.com/input-output-hk/cardano-wallet/blob/4e0f40135ab6ac16c1378be4a5e7b283e7494326/test/integration/Test/Integration/Framework/Request.hs#L107). If response status wasn't successful (>=200, <300) and is not 404 - then we are going to return origin error (ie, WalletNotFound).

In our case we are going to mark any `WalletNotFound` error with 404 https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/API/V1/Types.hs#L2660 which is going to end up being wrapped up to `ClientHttpError` at https://github.com/input-output-hk/cardano-wallet/blob/4e0f40135ab6ac16c1378be4a5e7b283e7494326/test/integration/Test/Integration/Framework/Request.hs#L107 . 

Mapping at https://github.com/input-output-hk/cardano-wallet/pull/327#discussion_r255101525 only transforms some internal representation of errors (from Kernel - consider this implementation detail) to `WalletNotFound` error. This wallet not found will be then wrapped up to `ClientHttpError` when request has been received in `unsafeRequest).

So `unsafeRequest` , if wallet or address is not found will return `ClientHttpRequest` (ie because of 404).

The `request` caller used bellow should work because `ClientHttpRequest` won't be wrapped in this function call (its rather simple) so I advise to use `request` instead of 

```haskell
unsafeRequest ("POST", "api/v1/wallets/aaa/addresses")
```

if that's possible. I believe problem is in `expectWalletError` https://github.com/input-output-hk/cardano-wallet/blob/develop/test/integration/Test/Integration/Framework/DSL.hs#L647 doesn't really have a case for `ClientHttpError` . Instead of current:

```haskell
expectWalletError
    :: (MonadIO m, MonadFail m, Show a)
    => WalletError
    -> Either ClientError a
    -> m ()
expectWalletError e' = \case
    Right a -> wantedErrorButSuccess a
    Left (ClientWalletError e) -> constructorOf e `shouldBe` constructorOf e'
    Left e  -> e `shouldBe` (ClientWalletError e')
  where
    constructorOf = typeRepTyCon . typeOf
```

which is unable to match that case pattern - as last catchall statement won't handle it. 

Alternativelly one could define a function similar to `expectWalletError` and call it `expectHttpError` : 
```haskell
expectHttpError
    :: (MonadIO m, MonadFail m, Show a)
    => WalletError
    -> Either ClientError a
    -> m ()
expectHttpError e' = \case
    Right a -> wantedErrorButSuccess a
    Left (ClientHttpError e) -> constructorOf e `shouldBe` constructorOf e'
    Left e  -> e `shouldBe` (ClientHttpError e')
  where
    constructorOf = typeRepTyCon . typeOf
```

and instead of the above `expectWalletError (WalletNotFound)` use this function.

I think this last proposal makes most sense. Hope this helps

piotr-iohk: Thanks a lot @akegalj for your thorough explanation! Now I understand that this is expected :+1: 
For now I have added this instead:
```haskell
expectError
-- TODO: add more expectations after #221 is resolved
```
yes, #221 is a ticket for extending DSL to support `ClientHttpError` I filed some time ago. Feel free to jump on it if you have some spare time ;) (it was beyond my skill, I'm afraid) 

I mean, now I was able to actually add this method, but completely failed to _decode_ this actual exception that appears in the testcase :/

```haskell
expectHttpError
    :: (MonadIO m, MonadFail m, Show a)
    => Client.ServantError
    -> Either ClientError a
    -> m ()
expectHttpError e' = \case
    Right a -> wantedErrorButSuccess a
    Left (ClientHttpError e) -> constructorOf e `shouldBe` constructorOf e'
    Left e  -> e `shouldBe` (ClientHttpError e')
  where
    constructorOf = typeRepTyCon . typeOf
```

In any case I think this PR could be merged.

akegalj (approved):

PR #321, opened 2019-02-07
[BIP44-239] Prepare to extend prefiltering to EOS wallets for ApplyBlock

uroboros:

<p align="right">#</p>

# Overview

- [x] In the BListener, I have extended prefiltering in`prefilterBlocks` to also prefilter for EO wallets during ApplyBlock
- [x] I have refactored prefilterBlock to return a single PrefilterResult, including a merged result in the case of prefiltering multiple wallet types together (i.e. there is at least one Eo and Fo wallet)
- [x] I have made the prefilter context for HdRnd (which was a list of `[(rootId,esk)]`) more consistent with Eo prefilter, a Map. Now both are maps.
- [x] I've changed `getWalletCredentials` to accept a snapshot instead of PassiveWallet (since the snapshot is already available at the call sites and this prevents a double getSnapshot)
- [x] I have moved `addrPoolGapByRootId` from the wallet layer to the Kernel, since I needed it to write `getEosWallets`. I also extended `addrPoolGapByRootId` to include accounts data along with the gap
- [x] I have written `getEosWallets`, that reconstructs Eo wallets and AddressPools from the AcidState db

# Comments

This PR adds prefiltering for Eo wallets during ApplyBlock. This means that we can now create new Eo wallets and track them during ApplyBlock.

NOTE: at this stage we cannot yet create new txs for Eo wallets. This will be in a subsequent commit.

# QA

We should test that we can create an Eo wallet, and if there are funds transferred to that wallet (using cardano=cli, say) we should see the funds appear in Daedelus.

We should also test Eo tracking in the presence of an HdRnd wallet! (In this case we have to merge prefilter results for the different wallet types, hence it needs special testing.

### DIFF
--- a/src/Cardano/Wallet/Kernel/Addresses.hs
+++ b/src/Cardano/Wallet/Kernel/Addresses.hs
@@ -20,6 +20,7 @@ import qualified Formatting.Buildable
 import           System.Random.MWC (GenIO, createSystemRandom, uniformR)
 
 import           Data.Acid (update)
+import qualified Data.Map.Strict as Map
 
 import           Pos.Core (Address, IsBootstrapEraAddr (..), deriveLvl2KeyPair)
 import           Pos.Core.NetworkMagic (NetworkMagic, makeNetworkMagic)
@@ -218,7 +219,7 @@ importAddresses
     -> IO (Either ImportAddressError [Either Address ()])
 importAddresses rootId addrs pw = runExceptT $ do
     esk <- lookupSecretKey rootId
-    lift $ forM addrs (flip importOneAddress [(rootId, esk)])
+    lift $ forM addrs (flip importOneAddress (Map.singleton rootId esk))
   where
     lookupSecretKey
         :: HdRootId
@@ -232,7 +233,7 @@ importAddresses rootId addrs pw = runExceptT $ do
 
     importOneAddress
         :: Address
-        -> [(HdRootId, EncryptedSecretKey)]
+        -> Map HdRootId EncryptedSecretKey
         -> IO (Either Address ())
     importOneAddress addr = evalStateT $ do
         let updateLifted = fmap Just .  lift . update (pw ^. wallets)
--- a/src/Cardano/Wallet/Kernel/BListener.hs
+++ b/src/Cardano/Wallet/Kernel/BListener.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LambdaCase   #-}
+
 -- | React to BListener events
 module Cardano.Wallet.Kernel.BListener (
     -- * Respond to block chain events
@@ -25,15 +26,17 @@ import qualified Formatting.Buildable
 
 import           Pos.Chain.Block (HeaderHash)
 import           Pos.Chain.Genesis (Config (..))
-import           Pos.Chain.Txp (TxId)
+import           Pos.Chain.Txp (TxId, TxIn)
 import           Pos.Core.Chrono (OldestFirst (..))
+import           Pos.Crypto (EncryptedSecretKey)
 import           Pos.DB.Block (getBlund)
 import           Pos.Util.Log (Severity (..))
 
-import           Cardano.Wallet.Kernel.DB.AcidState (ApplyBlock (..),
+import           Cardano.Wallet.Kernel.DB.AcidState (ApplyBlock (..), DB,
                      ObservableRollbackUseInTestsOnly (..), SwitchToFork (..),
                      SwitchToForkInternalError (..))
 import           Cardano.Wallet.Kernel.DB.BlockContext
+import           Cardano.Wallet.Kernel.DB.HdRootId (HdRootId)
 import           Cardano.Wallet.Kernel.DB.HdWallet
 import           Cardano.Wallet.Kernel.DB.InDb (fromDb)
 import           Cardano.Wallet.Kernel.DB.Resolved (ResolvedBlock, rbContext,
@@ -61,24 +64,48 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
 
 type PrefilterResult = ((BlockContext, PrefilteredBlock), [TxMeta])
 
--- | Prefilter the block for each account. Returns 'Nothing' if the prefiltering
--- is irrelevant for this node, i.e. there are no user wallets stored, so we
--- can avoid doing work (i.e. writing into the acid-state DB log) by skipping
--- such block application.
+-- | Prefilter a list of resolved blocks for all accounts in all wallets.
+-- Since different types wallets have different prefiltering mechanisms, we need
+-- to do prefiltering seperately for each type of wallet and then combine the results.
+--
+-- Returns 'Nothing' if the prefiltering is irrelevant for this node, i.e.
+-- there are no user wallets stored, so we can avoid doing work
+-- (e.g. writing into the acid-state DB log) by skipping such block application).
 prefilterBlocks :: PassiveWallet
                 -> [ResolvedBlock]
-                -> IO (Maybe [PrefilterResult])
+                -> IO (Maybe (NonEmpty PrefilterResult))
+prefilterBlocks _ [] = return Nothing
 prefilterBlocks pw bs = do
-    res <- getWalletCredentials pw
-    isForeign <- fmap Pending.txIns . foreignPendingByAccount <$> getWalletSnapshot pw
-    case res of
-         [] -> return Nothing
-         xs -> fmap Just $ flip evalStateT xs $ forM bs $ \rb -> do
-            pb <- state (prefilterBlock isForeign rb)
-            metas <- state (resolvedToTxMetas rb) >>= \case
-                Left e  -> throwM e
-                Right x -> return x
-            return ((rb ^. rbContext, pb), metas)
+    db <- getWalletSnapshot pw
+    let foreigns = fmap Pending.txIns . foreignPendingByAccount $ db
+    hdRnds <- getHdRndWallets db
+
+    nonEmpty <$> prefilterBlocks_ bs foreigns hdRnds
+    where
+        -- Get the data required for prefiltering of HdRnd wallets
+        getHdRndWallets
+            :: DB
+            -> IO (Map HdRootId EncryptedSecretKey)
+        getHdRndWallets db = getWalletCredentials db
+                    (pw ^. walletKeystore)
+                    (pw ^. walletProtocolMagic)
+                    (pw ^. walletLogMessage)
+
+-- Prefilter blocks for any class of wallets for which prefiltering is
+-- defined (any type constrained by IsOurs)
+prefilterBlocks_
+    :: IsOurs s
+    => [ResolvedBlock]
+    -> Map HdAccountId (Set TxIn)
+    -> s
+    -> IO [PrefilterResult]
+prefilterBlocks_ bs fs ws = do
+    flip evalStateT ws $ forM bs $ \rb -> do
+        pb <- state (prefilterBlock fs rb)
+        metas <- state (resolvedToTxMetas rb) >>= \case
+           Left e  -> throwM e
+           Right x -> return x
+        return ((rb ^. rbContext, pb), metas)
 
 data BackfillFailed
     = SuccessorChanged BlockContext (Maybe BlockContext)
@@ -211,7 +238,7 @@ applyBlock pw@PassiveWallet{..} b = do
               -- The block is not relevant as there are no user wallets stored
               -- in the DB, so don't bother writing into the acid-state transaction log.
               Nothing -> return $ Right ()
-              Just [((ctxt, blocksByAccount), metas)] -> do
+              Just (((ctxt, blocksByAccount), metas) :| _) -> do
                   -- apply block to all Accounts in all Wallets
                   mConfirmed <- update' _wallets $ ApplyBlock k ctxt accts blocksByAccount
                   case mConfirmed of
@@ -220,9 +247,6 @@ applyBlock pw@PassiveWallet{..} b = do
                           modifyMVar_ _walletSubmission (return . Submission.remPending confirmed)
                           mapM_ (putTxMeta _walletMeta) metas
                           return $ Right ()
-              Just _ -> error $ "applyOneBlock: the impossible happened, "
-                             <> "prefilterBlocks returned "
-                             <> "a different number of elements than the input ones."
 
       -- Determine if a failure in 'ApplyBlock' was due to the account being ahead, behind,
       -- or incomparable with the provided block.
@@ -315,7 +339,7 @@ switchToFork pw@PassiveWallet{..} oldest bs = do
          -- in this node.
          Nothing -> return ()
          Just xs -> do
-             let (blockssByAccount, metas) = unzip xs
+             let (blockssByAccount, metas) = unzip (NE.toList xs)
 
              changes <- trySwitchingToFork k blockssByAccount
 
--- a/src/Cardano/Wallet/Kernel/BListener.hs
+++ b/src/Cardano/Wallet/Kernel/BListener.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LambdaCase   #-}
+{-# LANGUAGE ViewPatterns #-}
 
 -- | React to BListener events
 module Cardano.Wallet.Kernel.BListener (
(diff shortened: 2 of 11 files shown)


src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -540,8 +544,9 @@ class IsOurs s where
 -- | NOTE: We could modify the given state here to actually store decrypted
 -- addresses in a `Map` to trade a decryption against a map lookup for already
 -- decrypted addresses.
-instance IsOurs [(HdRootId, Core.EncryptedSecretKey)] where
-    isOurs addr s = (,s) $ foldl' (<|>) Nothing $ flip map s $ \(rootId, esk) -> do
+instance IsOurs HdRootId Core.EncryptedSecretKey where
+    isOurs _ s@(Map.null -> True) = (Nothing, s)
```
KtorZ: This is really not necessary :relaxed: ... That's exactly what happens when folding on an empty map, and this doesn't really save us from any computation :+1:

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -220,9 +248,6 @@ applyBlock pw@PassiveWallet{..} b = do
                           modifyMVar_ _walletSubmission (return . Submission.remPending confirmed)
                           mapM_ (putTxMeta _walletMeta) metas
                           return $ Right ()
-              Just _ -> error $ "applyOneBlock: the impossible happened, "
-                             <> "prefilterBlocks returned "
-                             <> "a different number of elements than the input ones."
```
KtorZ, line 225: Why are we getting rid of the invariant here?

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -530,8 +533,9 @@ hdAccountRestorationState a = case a ^. hdAccountState of
   Address relationship: isOurs?
 -------------------------------------------------------------------------------}
 
-class IsOurs s where
-    isOurs :: Core.Address -> s -> (Maybe HdAddress, s)
+class IsOurs k v | v -> k where
+    -- For the given state, check whether an address is "ours"
+    isOurs :: Core.Address -> Map k v -> (Maybe HdAddress, Map k v)
```
KtorZ: I don't get the need for this change to be honest. I believe it comes after Matt's comment on the previous PR (https://github.com/input-output-hk/cardano-wallet/pull/264#discussion_r252371660) but, this was really about a way of "optimizing" this in case there is any need for optimization.

Could we please first get to something workable before trying to shove in any sort of optimization :pray: ? There's no point in optimizing something we can't analyze and measure first.

KtorZ (changes requested):

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: yes true, I'll remove these from both IsOurs instances

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: this is an important change because we need to be able to recognise empty prefiltering context. You'll see that at the entry points of prefiltering, without such a mechanism, in the case of empty prefiltering context, we would have to trivial call to isOurs for every single output address in a block of transactions.

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -61,24 +65,48 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
...
+-- | Prefilter a list of resolved blocks for all accounts in all wallets.
+-- Since different types wallets have different prefiltering mechanisms, we need
+-- to do prefiltering seperately for each type of wallet and then combine the results.
 prefilterBlocks :: PassiveWallet
                 -> [ResolvedBlock]
-                -> IO (Maybe [PrefilterResult])
+                -> IO (Maybe (NonEmpty PrefilterResult))
+-- Returns 'Nothing' if the prefiltering is irrelevant for this node, i.e.
+-- there are no user wallets stored, so we can avoid doing work
+-- (e.g. writing into the acid-state DB log) by skipping such block application).
```
parsonsmatt: This should be in the Haddock comment

parsonsmatt (approved):

src/Cardano/Wallet/Kernel/BListener.hs
KtorZ: @uroboros This has been left unanswered but, what am I missing :thinking: ?

uroboros (commented):
@KtorZ regarding your question about removing the invariant - I see my comment was "Pending" ;( - here it is "because I'm using a NonEmpty list now, so there's no "impossible" anymore. The error msg is incorrect though, it was failing when there was more than one prefilter result (not "a different number of elements")"

KtorZ:
I believe the invariant error was _right_, and was there to enforce that after prefiltering a singleton block, we get back a singleton result. 
Note that using a `NonEmpty` list doesn't quite solve the issue because we are now swallowing the failing pattern match because of:

```
     nonEmpty <$> prefilterBlocks_ bs foreigns hdRnds 
```

If, given a non empty list of blocks, prefilterBlocks was to return an empty list (or, as the original invariant indicated, a different number of blocks), then this is now appears as if there were no credentials at all and we would discard this result, without even knowing that something has happened. 

I think the real question here is why do we use a version of `prefilterBlocks` that relies on a list of blocks, where we do know that we are only passing around a single block.

https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/Kernel/BListener.hs#L210

I agree there's no need for this invariant, the fix however isn't going in the right direction. Rather, make `prefilterBlock` singular and have it take a single block and then later, in

https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/Kernel/BListener.hs#L312

fallback to a `forM` or` mapM` of that function -- that's what the current implementation does anyway. 

Moreover, this new function of yours `prefilterBlocks_` does a fair job at focusing only on the prefiltering part, leaving to the caller the freedom of getting the corresponding credentials (and not doing this in `prefilterBlocks` itself). It's nice to keep function doing one-and-one thing only. 



PS: stylish-haskell isn't happy :/

KtorZ (approved):

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -64,44 +64,40 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
...
-                    (pw ^. walletKeystore)
-                    (pw ^. walletProtocolMagic)
-                    (pw ^. walletLogMessage)
+    hdFOs <- getFOWallets pw db
+    return (foreigns, hdFOs)
+
+-- TODO @uroboros this function is trivial as it stands,
+-- as we add support for EO wallets, we will here do a separate call to
+-- prefilterBlock_ for FO and EO wallets and return the combined result
+prefilterBlock
```
uroboros, line 92: @KtorZ when I introduce EO prefiltering, we must cater for both EO + FO wallets being present. That means this function should return `IO (NonEmpty PrefilterResult)` because there might be multiple results (one for each wallet class). Alternatively we could defined <> for PrefilterResult, but I am not sure enough yet of the impact of such a change.

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -61,24 +64,44 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
...
+prefilterContext
+    :: PassiveWallet
+    -> IO ( Map HdAccountId (Set TxIn)
+          , Map HdRootId EncryptedSecretKey)
+       -- ^ (Foreigns, FO Prefilter Context)
+prefilterContext pw = do
+    db <- getWalletSnapshot pw
+    let foreigns = fmap Pending.txIns . foreignPendingByAccount $ db
+    hdFOs <- getFOWallets pw db
+    return (foreigns, hdFOs)
```
KtorZ: Note that we are changing the original behavior here and getting rid of one of the "optimizations" introduced as part of a fix during the 1.4.x QA phase. We use to return `Nothing` when there was no credentials / wallet to deal with, in order to skip prefiltering completely after.

So, the `Maybe` is gone, but there's no additional check at the caller site here to verify whether the second `Map` actually contains some credentials. It shouldn't matter much in practice, but let's keep that in mind for testing.

uroboros: ok, I'll make sure to skip prefiltering if there are no credentials (this becomes even more important when I add the EO wallet type here)

uroboros: (on second thought, I'll have to return that Maybe PrefResult)

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -569,10 +573,25 @@ mkAddressPool
...
+    = initAddressPool gap (mkAddressBuilder mkAddress accPK) addrs
+
+mkAddressBuilder
+    :: (Core.PublicKey -> Core.Address)
+    -> Core.PublicKey
+    -> Word32
+    -> Core.Address
+mkAddressBuilder mkAddress accPK addrIx
+    = case deriveAddressPublicKey accPK ExternalChain addrIx of
         Nothing     -> error "mkAddressPool: maximum number of addresses reached."
```
denisshevchenko, line 595: We decided to keep `error` here, right? I mean, is it appropriate solution here?

uroboros: I'll make that part of ErrAddressPoolInvalid instead...

uroboros: (will remove error call)

uroboros: @denisshevchenko on further thought, this behaviour is correct: this function will be deferred to during prefiltering (when an address pool gets extended). In that case it will have to fail hard if address generation fails. There are some other hard fails that occur during SwitchToFork and Appyblock

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -569,10 +573,25 @@ mkAddressPool
     -> Core.PublicKey
     -> AddressPoolGap
     -> AddressPool Core.Address
-mkAddressPool mkAddress accPK gap = emptyAddressPool gap newAddress
-  where
-    newAddress :: Word32 -> Core.Address
-    newAddress addrIx = case deriveAddressPublicKey accPK ExternalChain addrIx of
+mkAddressPool mkAddress accPK gap
```
rvl, line 576: After adding `mkAddressPoolExisting`, should `mkAddressPool` be renamed to `mkAddressPoolFullyOwned`?

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -61,24 +68,62 @@ import           Cardano.Wallet.WalletLayer.Kernel.Wallets
...
-         xs -> fmap Just $ flip evalStateT xs $ forM bs $ \rb -> do
-            pb <- state (prefilterBlock isForeign rb)
-            metas <- state (resolvedToTxMetas rb) >>= \case
-                Left e  -> throwM e
-                Right x -> return x
-            return ((rb ^. rbContext, pb), metas)
+-- | Provides the context required to prefilter all FO wallets.
+prefilterContext
+    :: PassiveWallet
+    -> IO ( Map HdAccountId (Set TxIn)
```
rvl, line 74: Would it aid readability if the return value of `prefilterContext` were packed into a little local data type?

src/Cardano/Wallet/Kernel/Restore.hs
```diff
@@ -178,11 +178,11 @@ mkPrefilter
     -> (HD.HdRootId, EncryptedSecretKey)
     -> Blund
     -> IO (PrefilteredBlock, [TxMeta])
-mkPrefilter wallet creds blund = do
+mkPrefilter wallet (rootId,esk) blund = do
     foreigns <- fmap Pending.txIns . foreignPendingByAccount <$> getWalletSnapshot wallet
     blundToResolvedBlock (wallet ^. Kernel.walletNode) blund >>= \case
         Nothing -> return (mempty, [])
-        Just rb -> flip evalStateT [creds] $ do
+        Just rb -> flip evalStateT (M.singleton rootId esk) $ do
```
rvl, line 185: Could this be shortened to `M.fromList [creds]`? Same goes for `getWalletInitInfo`.

src/Cardano/Wallet/Kernel/Read.hs
```diff
@@ -56,3 +96,115 @@ getWalletCredentials pw = do
...
+-- | Gathers all Eo accounts in all Eo roots and builds an AddressPool
+-- for each Eo account by using the existing addresses for the account.
+--
+-- NOTE: Fo wallet accounts are ignored.
+getEosPools
+    :: MonadIO m
+    => DB
+    -> (PublicKey -> Address)
+    -> m (Map HdAccountId (AddressPool Address))
+getEosPools db mkAddress
```
rvl, line 113: Would this function need any logging to aid users if the exceptional cases occurred?

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -542,8 +546,8 @@ class IsOurs s where
 -- | NOTE: We could modify the given state here to actually store decrypted
 -- addresses in a `Map` to trade a decryption against a map lookup for already
 -- decrypted addresses.
-instance IsOurs [(HdRootId, Core.EncryptedSecretKey)] where
-    isOurs addr s = (,s) $ foldl' (<|>) Nothing $ flip map s $ \(rootId, esk) -> do
+instance IsOurs (Map HdRootId Core.EncryptedSecretKey) where
```
rvl, line 549: Does this change of type affect the DB schema at all? I'm thinking probably not.

rvl (approved):
Looks good as far as I can tell. I only have questions or stylistic suggestions that you are free to ignore.

src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -232,7 +233,7 @@ importAddresses rootId addrs pw = runExceptT $ do
 
     importOneAddress
         :: Address
-        -> [(HdRootId, EncryptedSecretKey)]
+        -> Map HdRootId EncryptedSecretKey
```
akegalj, line 236: :+1:

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
akegalj: I don't think we are storing this info in this format

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: both of these functions are only relevant to EO wallet AddressPools - but I agree these names don't sit well together. mkAddressPool is not being used yet (coming soon), I'll look for better names when I use it.

src/Cardano/Wallet/Kernel/BListener.hs
uroboros: I was tempted by that - but then this would be the only use case.

src/Cardano/Wallet/Kernel/Restore.hs
uroboros: here I think `M.singleton rootId esk` makes it more explicit that we are dealing with one wallet only

src/Cardano/Wallet/Kernel/Read.hs
uroboros: that does need more work yes - since I'll be touching this function in the next 2 PRs I'll carry that over to later

src/Cardano/Wallet/Kernel/DB/HdWallet.hs
uroboros: IsOurs is used for prefiltering and does land in the DB in any explicit way

src/Cardano/Wallet/Kernel/Read.hs
```diff
@@ -56,3 +96,115 @@ getWalletCredentials pw = do
...
+        = toAccountAddressPools root (eosAccountsByRootId (root ^. hdRootId) db)
+
+    toAccountAddressPools
+        :: HdRoot
+        -> Maybe (Either GetAddressPoolGapError ([(HdAccountId, PublicKey)], AddressPoolGap))
+        -> [(HdAccountId, AddressPool Address)]
+    toAccountAddressPools _root accs
+        = case accs of
+            Nothing                    -> [] -- not an Eos root
+            (Just (Left err))          -> error (sformat build err)
```
akegalj, line 129: we have decided to fatal fail here - are we sure there is no way to recover from `GetAddressPoolGapError` ? Maybe there is no need though

akegalj (approved):
LGTM

Some logic was harder to grasp for me as I need some concepts to soak in.

I wonder is `test/unit/Util/Prefiltering.hs` covering all changes introduced here or do we need to add more tests?

Issue #254, opened 2019-01-24
Integration Tests Coverage: [WALLETS_UPDATE_PASS] Update wallet password


piotr-iohk:
# Context

#172

# Decision

#172

This particular ticket addresses adding integration tests for `[WALLETS_UPDATE_PASS] Update wallet password` capability.

# Acceptance Criterias

1. Create integration test scenarios for wallet capability: `[WALLETS_UPDATE_PASS] Update wallet password` defined in https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk/edit
2. Each integration tests must have reference ID (in the scenario title) to the capabilities doc item ID so it is clear which capability is covered by the test.

---

# Development Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #274      | `develop` |
| #323      | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| WALLETS_UPDATE_PASS  | Capabilities related to updating wallet password are now (partially) covered as part of the integration tests and clearly identified with a requirement number.  Some capabilities have not been covered or partially covered. Details below  |

#### Following bugs have been created:

#282 
#284 
(the issues concern faulty/misleading error messages returned by the API. When fixed, few integration tests will need to be updated to assert against new error messages)

#### Following ADRs have been created:

#271 (remove spendingPassword from `/api/v1/transactions/fees` ) - during the tests it has been _discovered_ that `spendingPassword` is not used for this endpoint but still it is listed in the API doc.
#285 - blocker for testing capability `WALLETS_UPDATE_PASS_10 | One can update password for the wallet that is 'restoring' and 'synced'. Appropriate info should be in the response in the 'syncState'.
`

#### Already existing but affecting:

#221 - ADR, when resolved there can be new assertions added to some of the tests (marked in the code with the commend `-- TODO: add more expectations after #221 is resolved`)

#193 - bug, also concerning error messages, referenced also in #282 and #284.

---

# Retrospective

piotr-iohk:
@KtorZ, putting this item for your QA review. I have added review suggestions / answered comments and rebased on corresponding pr #274.

PR #323, opened 2019-02-07
Fix Integration Tests UPDATE_PASS

KtorZ:
<p align="right">#254</p>

# Overview

- [ ] I have changed `expectWalletError` to only compare top-level constructors of error, avoiding us the trouble of defining the entire error, especially since some are just raw text strings created internally.

# Comments

I've also disable the UTxO statistics test until we figure out why it's failing.


paweljakubas (approved):

paweljakubas:
should be ok modulo needed stack2nix patch to hush buildkite

PR #324, opened 2019-02-07
Derivation wallet flag

akegalj:
<p align="right">45#</p>

# Overview

- [x] I have added a type `DerivationSchemeVersion` to `Wallet` to distinguish between sequential and random derivation schemes
- [x] I have added a type `DerivationScheme` to `HdRoot` to store this data to db
- [ ] ~I have added a db migration to set all old wallets to random derivation scheme~
          - not needed for 1.6 release
- [ ] Add invariant check that all addresses created with new derivation scheme don't have any payload
- [ ] hook up and run correct restoration process (sequential vs random) done by @uroboros 

# Comments

I am not sure how to test this feature? I can test migration but as this new flag is not fully wired yet I think we will need to postpone the testing until everything is in place. This adds additional flag to the db but doesn't do any logic with the flag yet.

I believe @uroboros work will use this flag. Also work from few of my PRs will also benefit the flag (as testing transaction signer is closer to be working now)

If you have some idea how to test -  please sare.


src/Cardano/Wallet/Kernel/DB/HdWallet/Derivation.hs
```diff
@@ -29,4 +34,21 @@ deriveIndex pickRange mkA hardeningMode =
...
 
+-- Which derivation scheme is being used
+data DerivationScheme
+    -- Derivation scheme random is following bip32 scheme, ed25519v0 curve and addresses contain: account index and address index
+    = RandomDerivationScheme
+    -- Derivation scheme sequential is bip44 scheme, ed25519v1 curve and addresses contain no address payload. Root key derivation from mnemonic keys also differs from root key derivation in random scheme
+    | SequentialDerivationScheme
+    deriving (Show, Eq, Ord)
 
+derivationScheme :: Core.Address -> DerivationScheme
```
akegalj, line 45: this can be removed now as this information is contained in wallet type and in db - will remove

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -17,25 +17,28 @@ import           Test.Hspec.QuickCheck (prop)
...
-import qualified Data.ByteString as B
 import qualified Data.Map.Strict as M
 
 import           Data.Acid (update)
 import           Formatting (build, formatToString, sformat)
 
 import           Pos.Chain.Txp (TxOut (..), TxOutAux (..))
 import           Pos.Core (Address, Coin (..), IsBootstrapEraAddr (..),
-                     deriveLvl2KeyPair, getCurrentTimestamp, mkCoin)
+                     deriveLvl2KeyPair, getCurrentTimestamp,
```
akegalj: This file was pushed by mistake - will force push

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -447,30 +450,96 @@ instance ToHttpApiData WalletId where
...
+instance BuildableSafeGen DerivationSchemeVersion where
+    buildSafeGen _ (DerivationSchemeVersion RandomDerivationScheme) = "random"
+    buildSafeGen _ (DerivationSchemeVersion SequentialDerivationScheme) = "sequential"
+
 -- | A Wallet Operation
 data WalletOperation =
     CreateWallet
-  | RestoreWallet
-  deriving (Eq, Show, Enum, Bounded)
+  | RestoreWallet DerivationSchemeVersion
```
KtorZ: This introduces a rather unfortunate breaking change in the API + a quite ad-hoc JSON instance :/ 
What do you think about having the derivation scheme as an extra optional field on the restore's payload ?

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -848,6 +917,7 @@ data Wallet = Wallet {
     , walCreatedAt                  :: !WalletTimestamp
     , walAssuranceLevel             :: !AssuranceLevel
     , walSyncState                  :: !SyncState
+    , walDerivationSchemeVersion    :: !DerivationSchemeVersion
```
KtorZ: Why the `Version` suffix exactly here :thinking: ?

src/Cardano/Wallet/Kernel/DB/HdWallet/Derivation.hs
KtorZ: maybe we can have that as an invariant check somewhere :thinking:  ?

stack.yaml
```diff
@@ -4,7 +4,7 @@ packages:
 
 extra-deps:
 - git: https://github.com/input-output-hk/cardano-sl
-  commit: cdc77284bde949d4b1b9ac7be81b76b906a71fc7
+  commit: 4ecd69bebd98c27d1d20d68c64293eb303a6cf8c
```
KtorZ: Is this still needed?

test/unit/Test/Spec/NewPayment.hs
```diff
@@ -87,39 +90,57 @@ prepareFixtures :: NetworkMagic
...
 
     return $ \keystore aw -> do
         liftIO $ Keystore.insert newRootId esk keystore
         let pw = Kernel.walletPassive aw
 
         let accounts    = Kernel.prefilterUtxo newRootId esk utxo'
-            hdAccountId = Kernel.defaultHdAccountId newRootId
-            hdAddress   = Kernel.defaultHdAddress nm esk emptyPassphrase newRootId
+            hdAccountId = newAccountId -- Kernel.defaultHdAccountId newRootId
+            hdAddress   = Nothing -- Kernel.defaultHdAddress nm esk passPhrase newRootId
```
KtorZ: Not sure what's going on here :thinking: ? Are you trying to make the tests work with both random and sequential derivation schemes?

KtorZ (changes requested):

src/Cardano/Wallet/API/V1/Types.hs
akegalj: I agree - that was my first attempt but I didn't push that part

src/Cardano/Wallet/API/V1/Types.hs
akegalj: > DerivationSchemeVersion

its a wrapper arround newtype `DerivationScheme` in this case

> walDerivationSchemeVersion

it is not needed in this case - will remove it

akegalj:
@KtorZ 
> maybe we can have that as an invariant check somewhere  ?

yes - this could be used in tests at least. Additionally we could throw an exception if (or an error) if that invariant is not satisfied as a "failsafe" if something whent wrong.

stack.yaml
akegalj: no - I pushed few additional commits - will cleanup

test/unit/Test/Spec/NewPayment.hs
akegalj: sorry - that's part of annother PR - pushed by mistake. Will cleanup

src/Cardano/Wallet/Kernel/DB/HdWallet/Derivation.hs
akegalj: yes - this could be used in tests at least. Additionally we could throw an exception if (or an error) if that invariant is not satisfied as a "failsafe" if something whent wrong.

test/unit/Test/Spec/NewPayment.hs
akegalj: force pushed

stack.yaml
akegalj: force pushed

src/Cardano/Wallet/API/V1/Types.hs
akegalj: resolved https://github.com/input-output-hk/cardano-wallet/pull/324/commits/915a6e1d8f881c93ff2357a310ee24ae64056814

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -449,30 +452,96 @@ instance ToHttpApiData WalletId where
...
+                ]
+
+instance ToJSON DerivationSchemeVersion where
+    toJSON (DerivationSchemeVersion RandomDerivationScheme)       = String "random"
+    toJSON (DerivationSchemeVersion SequentialDerivationScheme) = String "sequential"
+
+instance FromJSON DerivationSchemeVersion where
+    parseJSON (String "random")       = pure (DerivationSchemeVersion RandomDerivationScheme)
+    parseJSON (String "sequential") = pure (DerivationSchemeVersion SequentialDerivationScheme)
+    parseJSON x = typeMismatch "Not a valid DerivationSchemeVersion" x
```
KtorZ: Any ways to get those instances generically ?

akegalj: :+1: yes - we can remove the suffix. This was just first thing on my mind

akegalj: resolved with https://github.com/input-output-hk/cardano-wallet/pull/324/commits/97aa53d02feb3e8049ca19d63273b1a87c58e0d4

although fallbacking to using orphan instance. I can get rid of this and define instance in its own module - but I don't think its necessary

src/Cardano/Wallet/API/V1/Types.hs
akegalj: fixed with https://github.com/input-output-hk/cardano-wallet/pull/324/commits/bb4c767012a087afaa90cce479f75b92f498a628 and https://github.com/input-output-hk/cardano-wallet/pull/324/commits/40e2a73f2ad8ca564e321464f03879075f38eccd

PR #327, opened 2019-02-08
Map Kernel import address errors to API errors

KtorZ:
<p align="right">#258</p>

# Overview

- [ ] I have map import-address kernel errors to API errors 

# Comments

```
 $ curl -vkX POST "https://localhost:8090/api/v1/wallets/123/addresses" \
    --cert state-integration/tls/edge/client.pem   \
    -H "Content-Type: application/json" \ 
    -d '[]'

< HTTP/1.1 404 Reference to an unexisting wallet was given.

{"status":"error","diagnostic":{},"message":"WalletNotFound"}
```

### DIFF
--- a/src/Cardano/Wallet/API/V1/ReifyWalletError.hs
+++ b/src/Cardano/Wallet/API/V1/ReifyWalletError.hs
@@ -34,6 +34,7 @@ translateWalletLayerErrors :: SomeException -> Maybe V1.WalletError
 translateWalletLayerErrors ex = do
        (    try' @CreateAddressError createAddressError
         <|> try' @ValidateAddressError validateAddressError
+        <|> try' @ImportAddressError importAddressError
 
         <|> try' @CreateAccountError createAccountError
         <|> try' @GetAccountError getAccountError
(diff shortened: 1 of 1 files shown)


src/Cardano/Wallet/API/V1/ReifyWalletError.hs
```diff
@@ -60,7 +61,12 @@ translateWalletLayerErrors ex = do
   where try' :: forall e. Exception e => (e -> V1.WalletError) -> Maybe V1.WalletError
         try' f = f <$> fromException @e ex
 
-
+importAddressError :: ImportAddressError -> V1.WalletError
+importAddressError e = case e of
+    (ImportAddressError (Kernel.ImportAddressKeystoreNotFound _)) ->
+        V1.WalletNotFound
+    (ImportAddressAddressDecodingFailed _) ->
```
piotr-iohk, line 68: Does this mean that when address is invalid (e.g. too short, too long etc.) the `WalletNotFound` will be returned?

KtorZ: Well, there's a bit of historical context here but indeed. The internal business logic is more expressive about errors than the outer API is capable of. And for backward-compatibility reasons, we haven't been able to extend the API. And that's why we have to do this mapping from Kernel errors to API errors. 

In the end, I map here to `WalletNotFound` for lack of a better one, and, for consistency with other errors mapping already in place. This happens indeed when whatever you submit as a wallet id is note a base58 encoded valid root id. Like I did in my example `/api/v1/wallets/123/addresses`. This returns a `WalletNotFound` (and `123` isn't a valid wallet id in a first place).

piotr-iohk: ok, thanks. :+1:

piotr-iohk (approved):
lgtm :+1:

piotr-iohk:
@KtorZ, this one needs to be ported to `cardano-sl`, no?

PR #301, opened 2019-02-04
[WIP] Checking occassional CI utxo failure

paweljakubas:
<p align="right">#</p>

# Overview

This is for the sake of finding CI failure and understand it.
**DO NOT MERGE**
**DO NOT REVIEW**

# Comments

PR #329, opened 2019-02-09
Turn on 220 tests and add new one

paweljakubas:
<p align="right">#220</p>

# Overview

This is the second PR tackling #220. In the first one the error outlawing transactions with zero amount was introduced. In this one, the previously commented out tests are corrected and new one is added

**I checked 5 times within this PR, and also > 20 times using two more PRs : #303 and #301 that the occasional CI is not recurring.**

# Comments

- [x] I have added one more demanding test for checking UTXO
- [x] I have corrected  `WALLETS_LIST_03 - One can filter wallets by balance` scenario
- [x] I removed unnecessary duplicity in two tests for the sake of shortening run time by 15 minutes, just in order to avoid travis time limit (42 minutes) otherwise

### DIFF
--- a/test/integration/Test/Integration/Scenario/Wallets.hs
+++ b/test/integration/Test/Integration/Scenario/Wallets.hs
@@ -168,7 +168,7 @@ spec = do
             resp <- unsafeRequest ("GET", fromString endpoint) $ Nothing
             verify (resp :: Either ClientError [Wallet]) expectations
 
-    describe "WALLETS_LIST_03 - One can filter wallets by balance" $ do
+    scenario "WALLETS_LIST_03 - One can filter wallets by balance" $ do
 
         let matrix =
                 [ ( "api/v1/wallets?balance=EQ%5B3%5D"
@@ -188,7 +188,7 @@ spec = do
                   , [ expectSuccess
                     , expectListSizeEqual 2
                     , expectListItemFieldEqual 0 amount 3
-                    , expectListItemFieldEqual 1 amount 0
+                    , expectListItemFieldEqual 1 amount 1
                     ]
                   )
                 , ( "api/v1/wallets?balance=GT%5B6%5D"
@@ -210,11 +210,11 @@ spec = do
                     , expectListSizeEqual 4
                     , expectListItemFieldEqual 0 amount 6
                     , expectListItemFieldEqual 1 amount 6
-                    , expectListItemFieldEqual 2 amount 6
-                    , expectListItemFieldEqual 3 amount 0
+                    , expectListItemFieldEqual 2 amount 3
+                    , expectListItemFieldEqual 3 amount 1
                     ]
                   )
-                , ( "api/v1/wallets?balance=RANGE%5B6%3,6D&sort_by=balance"
+                , ( "api/v1/wallets?balance=RANGE%5B3,6%5D&sort_by=balance"
                   , [ expectSuccess
                     , expectListSizeEqual 3
                     , expectListItemFieldEqual 0 amount 6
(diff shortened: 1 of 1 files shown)


test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -168,7 +168,7 @@ spec = do
             resp <- unsafeRequest ("GET", fromString endpoint) $ Nothing
             verify (resp :: Either ClientError [Wallet]) expectations
 
-    describe "WALLETS_LIST_03 - One can filter wallets by balance" $ do
+    scenario "WALLETS_LIST_03 - One can filter wallets by balance" $ do
```
piotr-iohk, line 171: What is the reason for this change? Is it due to performance?

The benefit of having it in a `descibre -> scenario` fashion here is that when a particular scenario fails you know which one it is. If you have bunch of different cases in a single scenario (like here, after the change) and one of them fails you need to debug every one of them to see which one is actually failing.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -414,7 +413,7 @@ spec = do
             , expectListItemFieldEqual 1 walletId ( Client.WalletId $ walletIds !! 0 )
             ]
 
-    describe "WALLETS_LIST_04 - One can sort results only by 'balance' and 'created_at'" $ do
+    scenario "WALLETS_LIST_04 - One can sort results only by 'balance' and 'created_at'" $ do
```
piotr-iohk, line 416: Same here.

test/integration/Test/Integration/Scenario/Wallets.hs
paweljakubas: huge time saving . Previously, every GET the setup was created from scratch. Every coin from initalCoins has to realized by `postTransaction` and it takes minutes to set-up it. And I think it is enough to create one setup and invoke all GETs. If we fail, it will tell that there is something wrong in the scenario - then locally it is easy to spot it using prints...

paweljakubas: otherwise, we are going to crunch 42 minute travis limit

test/integration/Test/Integration/Scenario/Wallets.hs
paweljakubas: the same rationale as described above. If we invoke GETs we do not need to spend several minutes per GET to separately start from scratch

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ok, fair enough :+1:

piotr-iohk (approved):
lgtm :+1:

akegalj (approved):

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -808,8 +806,6 @@ spec = do
             ]
 
     scenario "WALLETS_UTXO_04 - UTxO statistics reflect wallet's activity" $ do
-        pendingWith "disabled until we figure out why it is randomly failing.  See #220"
```
KtorZ, line 811: @paweljakubas what was the cause / fix in the end?

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -820,6 +816,16 @@ spec = do
...
             ]
 
+
+    scenario "WALLETS_UTXO_05 - UTxO statistics reflect wallet's activity" $ do
+        fixture <- setup $ defaultSetup
+            & initialCoins .~ [13, 43, 66, 101, 1339]
+        response <- request $ Client.getUtxoStatistics
+            $- (fixture ^. wallet . walletId)
+        verify response
+            [ expectWalletUTxO [13, 43, 66, 101, 1339]
```
KtorZ, line 826: @paweljakubas What's the added value with the scenario compare to the above already existing?

test/integration/Test/Integration/Scenario/Wallets.hs
paweljakubas: There were several things :  
1. There were tests with 0 amount initial coins that are now in Byron forbidden
2. Then I was able to detect this problem only with 3-element-list initialCoins
3. After some rebase including commit between 5-9 Feb I was not able to detect this malfunction (and I checked it more than 20 times, maybe even 30 times in three PRs)
4. Then I added correction to tests as above and also added change that reused calling initalCoins (in order not to exceed travis limit of 42 minutes)

test/integration/Test/Integration/Scenario/Wallets.hs
paweljakubas: As commented above, 
```
Then I was able to detect this problem only with 3-element-list initialCoins  
```
I decided to add something that could probably increase the chance of spotting the malfunction - but I have not encountered it afterwards

KtorZ: Hmmm... So you mean that, in the end, the UTxO statistics issue just ... vanished :thinking: ?

paweljakubas: It was not UTXO statistics issue, but 3-element-size list initalCoins (which call postTransaction inside) malfunction (in a sense, that in the third postTransaction call it sometimes happened that there are no utxo inputs available)... - this is where the error appeared. And at that time, Utxo statistics test was the only one whch used 3-element list in initialCoins.......  Yes, after rebase (probably 7-8 Feb) I was not able to reproduce it anymore. Previously one out of 3-5 trials I was able to catch it, then I tried **25-35 times** (see this and #303 and #301 to see details) and was not able to spot it. Then, I also followed another 10-20 CI runs and have not seen it.  Only after that number of observations I decided that the problem is not affecting anymore

PR #332, opened 2019-02-11
Fill placeholders by missing handlers for EOS-wallets.

denisshevchenko:
<p align="right">#</p>

# Overview

- [x] I have added missing handlers for EOS-wallets.

# Comments

### DIFF
--- a/src/Cardano/Wallet/Client.hs
+++ b/src/Cardano/Wallet/Client.hs
@@ -165,6 +165,19 @@ data WalletClient m
     , getEosWallet
         :: WalletId
         -> Resp m EosWallet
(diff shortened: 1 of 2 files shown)


uroboros (approved):
LGTM

akegalj (approved):

KtorZ:
Note that this was left out on purpose :) 
If we don't use it, let's avoid the burden of maintaining it. I understand from the Slack that defining this allowed @piotr-iohk to move forward writing some extra integration tests. Then, may we link this PR to the corresponding ticket to understand the rational behind this change :pray:

PR #334, opened 2019-02-12
EOSWALLETS tests

piotr-iohk:
<p align="right">#</p>

# Overview

- [x] I have added tests for EOSWALLETS_CREATE capability -> https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk
- [x] I have added few tests covering read, update, delete and list capability
- [x] I intend to add more tests for EOSWALLETS here

# Comments

### DIFF
--- a/test/integration/Test/Integration/Scenario/EosWallets.hs
+++ b/test/integration/Test/Integration/Scenario/EosWallets.hs
@@ -472,7 +472,6 @@ spec = do
     describe "EOSWALLETS_UPDATE_03 - addressPoolGap cannot be outside [10..100]" $ do
         forM_ ([-1, 0, 9, 101]) $ \poolGap -> scenario ("addressPoolGap = " ++ show (poolGap :: Int)) $ do
             fixture  <- setup $ defaultSetup
-                & rawAddressPoolGap .~ 12
             eowallet <- successfulRequest $ Client.postEosWallet $- NewEosWallet
                 (NE.fromList $ take 3 $ fixture ^. externallyOwnedAccounts)
                 noAddressPoolGap
@@ -522,9 +521,8 @@ spec = do
 
         forM_ matrix $ \(title, assurLevel, expectations) -> scenario ("assuranceLevel = " ++ title) $ do
             fixture  <- setup $ defaultSetup
-                & rawAddressPoolGap .~ 12
             eowallet <- successfulRequest $ Client.postEosWallet $- NewEosWallet
-                (NE.fromList $ take 3 $ fixture ^. externallyOwnedAccounts)
+                (NE.fromList $ take 5 $ fixture ^. externallyOwnedAccounts)
                 noAddressPoolGap
                 defaultAssuranceLevel
                 defaultWalletName
--- a/test/integration/Test/Integration/Scenario/EosWallets.hs
+++ b/test/integration/Test/Integration/Scenario/EosWallets.hs
@@ -10,6 +10,7 @@ import qualified Data.List.NonEmpty as NE
 
 import           Cardano.Wallet.Client.Http (ClientError, Wallet)
 import qualified Cardano.Wallet.Client.Http as Client
+import           Test.Hspec (describe)
 import           Test.Integration.Framework.DSL
 
 
@@ -22,7 +23,6 @@ spec = do
             noAddressPoolGap
             defaultAssuranceLevel
             defaultWalletName
-
         response <- request $ Client.getEosWallet $- eowallet ^. walletId
 
         verify response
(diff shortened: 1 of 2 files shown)


test/integration/Test/Integration/Scenario/EosWallets.hs
```diff
@@ -50,14 +50,88 @@ spec = do
...
+                            }
+                            ],
+                "addressPoolGap": #{poolGap},
+                "assuranceLevel": "normal",
+                "name": "My EOS Wallet"
+                }|]
+            verify (response :: Either ClientError EosWallet)
+                [ expectJSONError $ case poolGap of
+                                        -1 -> "Error in $.addressPoolGap: Word8 is either floating or will cause over or underflow: -1.0"
+                                        _  -> "Error in $.addressPoolGap: Address pool gap should be in range [10..100]"
```
piotr-iohk, line 73: I like this error message `Address pool gap should be in range [10..100]`. I was expecting it also when providing `-1` as `addressPoolGap`. Instead we have `Word8 is either floating or will cause over or underflow: -1.0`. Maybe we'd like to change it for consistency?

piotr-iohk:
@denisshevchenko - I'll ask here as it is related to creating EOS wallets. In the API doc -> https://input-output-hk.github.io/cardano-wallet/#tag/Externally-Owned-Wallets%2Fpaths%2F~1api~1v1~1wallets~1externally-owned%2Fpost it is said `Creates a new or restores an existing Wallet`. Is there an option to `restore` or the mention about the `restore` should be actually removed from the doc?

denisshevchenko:
> Is there an option to `restore`

@piotr-iohk Not yet.

denisshevchenko (approved):

test/integration/Test/Integration/Scenario/EosWallets.hs
akegalj: > Error in $.addressPoolGap: Word8 is either floating or will cause over or underflow: -1.0

msg is interesting one - I didn't expect to see something like this. For example when I try to manually create address pool gap with a negative number:
```haskell
> AddressPoolGap (-1)
```
I get
```haskell
    Literal -1 is out of the Word8 range 0..255
```
which makes sense as Word8 is defined in specified range.

Edit:
Aha - msg comes from `Aeson` and decoding a "-1" string into a `Word8`:

```haskell
> import Data.Aeson
> eitherDecode "-1" :: Either String Word8
Left "Error in $: Word8 is either floating or will cause over or underflow: -1.0"
```

so everything is as expected here. So you will try to push your json to the api. First thing that api will do is it will try to decode this json into expected Haskell data type. This will immediatelly fail in this json decoding part as Word8 won't be able to decode "-1" and `Aeson` library will rise the exception. If the exception wasn't rased by underlying aeson lib, we would catch it and rise "Address pool gap should be in range [10..100]". 

If we would like to rise only single exception for `AddressPoolGap` one could define custom `FromJSON` instance that works only for `[10..100]` (ie, `[minBound..maxBound]`) and not to use custom/generic aeson instance https://github.com/input-output-hk/cardano-wallet/blob/15dd91b05112c16be506129e5c00e88114f8bd16/src/Cardano/Wallet/Kernel/AddressPoolGap.hs#L33 . This would do the trick in our case:
```haskell
instance FromJSON AddressPoolGap where
    parseJSON = parseJSON >=> \gap -> case mkAddressPoolGap gap of
        Left err -> fail $ toString $ sformat build $ bprint
        ("Address pool gap should be in range ["%int%".."%int%"].")
        (minBound @AddressPoolGap)
        (maxBound @AddressPoolGap)
        Right x  -> pure x
```

or something along these lines. Anyway if you think this would be nice to have please open a new ticket and we will discuss it with rest of the team. For now I don't think this is necessary nor urgent to have

test/integration/Test/Integration/Scenario/EosWallets.hs
```diff
@@ -48,16 +49,521 @@ spec = do
...
+                        {
+                            "index": 2710723942
+                        }
+                        ],
+            "addressPoolGap": 10,
+            "assuranceLevel": "strict",
+            "name": "My EOS Wallet"
+            }|]
+        verify (response :: Either ClientError EosWallet)
+            [ expectJSONError "Error in $.accounts[0]: When parsing the record accountPublicKeyWithIx of type Cardano.Wallet.API.V1.Types.AccountPublicKeyWithIx the key publicKey was not present."
```
akegalj: hm - these types of error checks will gratelly depend on aeson version (msg might be modified in aeson) and on naming of the module where the function is located (if someone changes name of the module or name of the function itself msg will change)

I guess its ok as a temp solution. What do you think @denisshevchenko @uroboros ?

test/integration/Test/Integration/Scenario/EosWallets.hs
```diff
@@ -48,16 +49,521 @@ spec = do
...
+            & assuranceLevel .~ NormalAssurance
+
+        let endpoint = "api/v1/wallets/externally-owned/" <> fromWalletId (fixture ^. wallet . walletId)
+        updResp <- unsafeRequest ("PUT", endpoint) $ Just $ [json|{
+                "assuranceLevel": "strict",
+                "addressPoolGap": 12,
+                "name": "My EosWallet2"
+                }|]
+        verify (updResp :: Either ClientError EosWallet)
+            [ expectWalletError (UnknownError "UpdateEosWalletError")
```
akegalj, line 412: I don't think this is right error - right @denisshevchenko @uroboros ?

I believe we are missing case such as `<|> try' @UpdateEosWalletError updateEosWalletError` at https://github.com/input-output-hk/cardano-wallet/blob/develop/src/Cardano/Wallet/API/V1/ReifyWalletError.hs#L47

test/integration/Test/Integration/Scenario/EosWallets.hs
denisshevchenko: Agree, phrase in `expectJSONError` is too long and specific.

piotr-iohk: As for the error messages `Address pool gap should be in range [10..100]` and `Word8 is either floating or will cause over or underflow` I'd be for having one consistent message and obviously the first one is more user friendly and informative. I'll create a ticket. This is actually observed for `addressPoolGap` and `index` fields for certain values outside their ranges...

>depend on ... naming of the module where the function is located (if someone changes name of the module or name of the function itself msg will change)

makes sense, I'll shorten the messages. :+1:

piotr-iohk:
@akegalj I have created a ticket #342 and marked test `EOSWALLETS_LIST_01, EOSWALLETS_LIST_02 - cannot get FO wallets with EOS endpoint` with `pendingWith`. From my side it is ready to be merged.

test/integration/Test/Integration/Scenario/EosWallets.hs
akegalj: fixed in my local branch - will push after we merge #337 (which fixes some failures) and test the patch

PR #335, opened 2019-02-12
Prefilter FO-wallets and EO-wallets, in list-handlers.

denisshevchenko:
<p align="right">#336</p>

# Overview

- [x] I have added prefiltering for FO-walltes and EOS-wallets, in `list`-handlers.

# Comments

### DIFF
--- a/src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
+++ b/src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
@@ -312,11 +312,16 @@ getWallets :: MonadIO m
            -> Kernel.DB
            -> m (IxSet V1.Wallet)
 getWallets wallet db =
-    fmap IxSet.fromList $ forM (IxSet.toList allRoots) $ \root -> do
+    fmap IxSet.fromList $ forM foRoots $ \root -> do
         let rootId = root ^. HD.hdRootId
         updateSyncState wallet rootId (toWallet db root)
   where
-    allRoots = db ^. dbHdWallets . HD.hdWalletsRoots
+    -- From 'acid-state' point of view there's no difference between
+    -- FO-wallet and EO-wallet, because it's just an `HdRoot`. So we
+    -- have to prefilter 'HdRoot's which are correspond to FO-wallets only.
+    foRoots = flip filter allRoots $ \root ->
+        isLeft $ Kernel.addressPoolGapByRootId (root ^. HD.hdRootId) db
+    allRoots = IxSet.toList $ db ^. dbHdWallets . HD.hdWalletsRoots
 
 getEosWallets
     :: MonadIO m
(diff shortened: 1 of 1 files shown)


uroboros (approved):
looks isRight to me ;)

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -312,11 +312,16 @@ getWallets :: MonadIO m
...
-    fmap IxSet.fromList $ forM (IxSet.toList allRoots) $ \root -> do
+    fmap IxSet.fromList $ forM foRoots $ \root -> do
         let rootId = root ^. HD.hdRootId
         updateSyncState wallet rootId (toWallet db root)
   where
-    allRoots = db ^. dbHdWallets . HD.hdWalletsRoots
+    -- From 'acid-state' point of view there's no difference between
+    -- FO-wallet and EO-wallet, because it's just an `HdRoot`. So we
+    -- have to prefilter 'HdRoot's which are correspond to FO-wallets only.
+    foRoots = flip filter allRoots $ \root ->
```
akegalj, line 322: This solves the problem - but I wonder would it be wise to make a distinction between FO and EO wallets in our db?

akegalj (approved):

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
denisshevchenko: In this case we have to change `HdRoot` type, and (for simplicity) we decided to keep this type in its current state. At least, for _now_.

uroboros: There is an implicit distinction in the DB but only at the Account level. This is why we can only make the distinction on read (when we have the accounts available). We could make it clearer with a function in Kernel.Read that partitions a list of roots into ([FO],[EO]) <--- @denisshevchenko what do you think?

akegalj: I would go that route for now 
> function in Kernel.Read that partitions a list of roots into ([FO],[EO])

as this way we move this detail up the stack so there is less room for mistake and it can be reused in both places (maybe more in future).

denisshevchenko: > We could make it clearer with a function in Kernel.Read that partitions a list of roots into ([FO],[EO]) <--- @denisshevchenko what do you think?

Good idea! But probably we have to implement _two_ different functions for EO-list and FO-list..

uroboros: the way I see it is, say for getEOs ...

(_, eos) <- partitionWallets allRoots
for eos ...

uroboros: (but 2 separate functions in top of that would also add value I think)

KtorZ:
I believe this ended up in the wrong abstraction layer. We duplicated logic a bit everywhere when it comes to identifying F.O. wallets vs E.O. wallets. Instead, we could / should piggy-back on the DB layer to provide us with the right wallets when we ask for them. Outside of the Wallet Layer. 

Furthermore, there's one question we keep asking about wallets and it's "is this wallet fully-owned or not?". We must have a clear AND UNIQUE way of giving an answer to that question. It is not okay to have bits of duplicated logic which answers exactly this question :+1:

Issue #336, opened 2019-02-12
Make EO endpoints to manipulate EO wallets and FO to manipulate FO wallets only


piotr-iohk:
# Context

While testing #331 we have observed that one can manipulate EOS wallets with FO endpoints and FO wallets with EOS endpoints. In particular:

 - You are able to list EOS wallets with FO enpoint `GET /api/v1/wallets`.
 - You can successfully delete EOS wallet with FO endpoint and vice versa. (EOS with `DELETE /api/v1/wallets/{walletId}` and FO with `DELETE /api/v1/wallets/externally-owned/{walletId}`)
 - You can list EOS wallet with FO endpoint `GET /api/v1/wallets/{walletId}` (here you cannot do otherwise, and trying to get FO wallet with EOS enpoint results in `WalletNotFound` - which is expected I suppose)
 - You can update EOS wallets with FO endpoint and vice-versa (here, after updating one or another with its counterpart endpoint - an error is returned by the API, still wallets are updated)

There were integration tests added for that under #331, they are currently failing:

```
  EOSWALLETS_DELETE_02 - cannot delete FO wallet with EOS endpoint FAILED [1]
  EOSWALLETS_DELETE_02 - cannot delete EOS wallet with FO endpoint FAILED [2]
  EOSWALLETS_DETAILS_01 - cannot get EOS wallet with FO endpoint FAILED [3]
  EOSWALLETS_LIST_01 - cannot get EOS wallets with FO endpoint FAILED [4]
  EOSWALLETS_LIST_01 - cannot get FO wallets with EOS endpoint FAILED [5]
  EOSWALLETS_UPDATE_01 - cannot update FO wallets with EOS endpoint FAILED [6]
  EOSWALLETS_UPDATE_01 - cannot update EOS wallets with FO endpoint FAILED [7]

```

# Decision

Manipulation with EO wallets must be restricted only to EO endpoints.
Manipulation with FO wallets must be restricted only to FO endpoints.

# Acceptance Criterias

See ^.
Integration tests mentioned here should pass.

---

# Development Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #335        | `develop` |
| #337        | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| ?        | -        |

---

# Retrospective

denisshevchenko:
[...]

PR #337, opened 2019-02-12
Check for Read, Delete and Update operations, for all wallets.

denisshevchenko:
<p align="right">#336</p>

# Overview

- [x] I have added checks for `Read`, `Delete` and `Update` operations, for all wallets.

# Comments

### DIFF
--- a/src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
+++ b/src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
@@ -33,7 +33,6 @@ import qualified Cardano.Mnemonic as Mnemonic
 import qualified Cardano.Wallet.API.V1.Types as V1
 import qualified Cardano.Wallet.Kernel.Accounts as Kernel
 import           Cardano.Wallet.Kernel.Addresses (newHdAddress)
-import           Cardano.Wallet.Kernel.DB.AcidState (dbHdWallets)
 import qualified Cardano.Wallet.Kernel.DB.HdRootId as HD
 import qualified Cardano.Wallet.Kernel.DB.HdWallet as HD
 import           Cardano.Wallet.Kernel.DB.InDb (fromDb)
@@ -44,7 +43,9 @@ import           Cardano.Wallet.Kernel.Internal (walletKeystore, walletMeta,
                      walletProtocolMagic, (<.>), _wriProgress)
 import qualified Cardano.Wallet.Kernel.Internal as Kernel
 import qualified Cardano.Wallet.Kernel.Keystore as Keystore
-import           Cardano.Wallet.Kernel.Read (addressPoolGapByRootId)
+import           Cardano.Wallet.Kernel.Read (WalletOwnership (..),
+                     addressPoolGapByRootId, eosAccountsByRootId,
+                     getWalletOwnership, getWalletsByOwnership)
 import qualified Cardano.Wallet.Kernel.Read as Kernel
 import           Cardano.Wallet.Kernel.Restore (blundToResolvedBlock,
                      restoreWallet)
@@ -188,10 +189,15 @@ updateWallet :: MonadIO m
              -> m (Either UpdateWalletError V1.Wallet)
 updateWallet wallet wId (V1.WalletUpdate v1Level v1Name) = runExceptT $ do
     rootId <- withExceptT UpdateWalletWalletIdDecodingFailed $ fromRootId wId
-    v1wal <- fmap (uncurry toWallet) $
-               withExceptT UpdateWalletError $ ExceptT $ liftIO $
-                 Kernel.updateHdWallet wallet rootId newLevel newName
-    updateSyncState wallet rootId v1wal
+    db <- liftIO $ Kernel.getWalletSnapshot wallet
+    case getWalletOwnership rootId db of
+        WalletExternallyOwned ->
+            exceptT $ Left $ UpdateWalletError $ HD.UnknownHdRoot rootId
+        WalletFullyOwned -> do
+            v1wal <- fmap (uncurry toWallet) $
+                       withExceptT UpdateWalletError $ ExceptT $ liftIO $
+                         Kernel.updateHdWallet wallet rootId newLevel newName
+            updateSyncState wallet rootId v1wal
   where
     newLevel = fromAssuranceLevel v1Level
     newName  = HD.WalletName v1Name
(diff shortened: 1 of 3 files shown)


src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -188,10 +188,17 @@ updateWallet :: MonadIO m
              -> m (Either UpdateWalletError V1.Wallet)
 updateWallet wallet wId (V1.WalletUpdate v1Level v1Name) = runExceptT $ do
     rootId <- withExceptT UpdateWalletWalletIdDecodingFailed $ fromRootId wId
-    v1wal <- fmap (uncurry toWallet) $
-               withExceptT UpdateWalletError $ ExceptT $ liftIO $
-                 Kernel.updateHdWallet wallet rootId newLevel newName
-    updateSyncState wallet rootId v1wal
+    db <- liftIO $ Kernel.getWalletSnapshot wallet
+    case addressPoolGapByRootId rootId db of
```
uroboros: We are using `addressPoolGapByRootId` for asking "is EO wallet". I think it's worth making it more explicit with a wrapper `isEOWallet :: DB -> HdRootId -> Bool`

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -204,22 +211,29 @@ updateEosWallet
...
-        else do
-            result <- forM accounts $ \hdAcc -> do
-                let accId = hdAcc ^. HD.hdAccountBase . HD.hdAccountBaseId
-                res <- liftIO $ Kernel.updateAccountGap accId newGap wallet
-                return $ either (Left . UpdateEosWalletAccountError) Right res
-            exceptT $ case lefts result of
-                []    -> Right $ toEosWallet db newHdRoot newGap
-                (e:_) -> Left e
+    db' <- liftIO $ Kernel.getWalletSnapshot wallet
+    case addressPoolGapByRootId rootId db' of
```
uroboros: if you use `eosAccountsByRootId` here instead of `addressPoolGapByRootId`, this function will simplify quite a bit

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
uroboros: (we might as well add an isFOWallet too?)

uroboros (changes requested):

rvl (commented):
1. I think we could reduce the repetition and make it clearer.

2. I agree with @uroboros, but rather than `isEOWallet` and `isFOWallet` we could have something like:

```
data WalletOwnership
    = WalletFullyOwned
    | WalletExternallyOwned
    deriving (Show, Eq)

getWalletOwnership :: HD.HdRootId -> Kernel.PassiveWallet -> IO WalletOwnership
getWalletOwnership rootId = fmap ownership . Kernel.getWalletSnapshot
    where
        ownership db = case addressPoolGapByRootId rootId db of
                           Right _ -> WalletExternallyOwned
                           Left _ -> WalletFullyOwned
```

3. Is `HD.UnknownHdRoot` the right exception, or should there be another one added for the case where the action is not applicable?

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
denisshevchenko: Agree. I'll do it.

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -204,22 +210,27 @@ updateEosWallet
...
-        then exceptT $ Left $ UpdateEosWalletErrorNoAccounts wId
-        else do
-            result <- forM accounts $ \hdAcc -> do
-                let accId = hdAcc ^. HD.hdAccountBase . HD.hdAccountBaseId
-                res <- liftIO $ Kernel.updateAccountGap accId newGap wallet
-                return $ either (Left . UpdateEosWalletAccountError) Right res
-            exceptT $ case lefts result of
-                []    -> Right $ toEosWallet db newHdRoot newGap
-                (e:_) -> Left e
+    db' <- liftIO $ Kernel.getWalletSnapshot wallet
```
uroboros, line 213: As it stands, we're repeating logic that is captured by `eosAccountsByRootId`.
If you use that function instead of `getWalletOwnership` you'll get
* accountIds, no need to read them again
* checks for `GetEosWalletErrorNoAccounts` and `GetEosWalletErrorWrongAccounts` are built in

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
```diff
@@ -188,10 +189,15 @@ updateWallet :: MonadIO m
              -> m (Either UpdateWalletError V1.Wallet)
 updateWallet wallet wId (V1.WalletUpdate v1Level v1Name) = runExceptT $ do
     rootId <- withExceptT UpdateWalletWalletIdDecodingFailed $ fromRootId wId
-    v1wal <- fmap (uncurry toWallet) $
-               withExceptT UpdateWalletError $ ExceptT $ liftIO $
-                 Kernel.updateHdWallet wallet rootId newLevel newName
-    updateSyncState wallet rootId v1wal
+    db <- liftIO $ Kernel.getWalletSnapshot wallet
+    case getWalletOwnership rootId db of
```
uroboros, line 193: A (debatable) suggestion...

We can simplify these calls to `getWalletOwnership` with a helper function, e.g. 

`ensureOwnership :: WalletOwnership -> (HdRootId -> e) -> ...`

and use it like this:

_ <- ensureOwnership WalletFullyOwned (UpdateWalletError . HD.UnknownHdRoot)
-- from here on we can assume correct ownership...

uroboros (changes requested):
mostly LGTM - just one issue with `updateEosWallet`

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
akegalj: I like the proposal but underlying code would not change much. We would still have to do pattern on `case` expression - only some bits of the code would be rearanged slightly.

But its up to you - I am fine with both solutions

src/Cardano/Wallet/WalletLayer/Kernel/Wallets.hs
akegalj: @denisshevchenko what are your thoughts on this?

denisshevchenko: Yes, good idea. I'll use that function.

denisshevchenko: Done.

uroboros (approved):
LGTM!

akegalj (approved):

PR #338, opened 2019-02-13
Add nix shell file for stack build

rvl:
# Overview

When building with `stack --nix` it will fetch all system dependencies (e.g. rocksdb). If stack/nix integration is not enabled, then this change has no effect.

# Comments

Tested with `stack --nix build`.


angerman (approved):
LGTM

PR #340, opened 2019-02-14
[BIP44-239] Extend newTx to EO wallets

uroboros:

<p align="right">#</p>

# Overview

- [x] I have extended NewTx (used in NewPending/NewForeign) to apply EO prefiltering when adding a transaction to an EO Account

# Comments

These changes add a new capability (new Tx for EO wallets) and does not test it yet. 
Existing tests for "newTx for HdRnd wallets" are still satisfied so we can be sure that this does not break any existing functionality.

Since both @piotr-iohk and I want to test this new functionality next  (on API and Kernel level respectively) I suggest we review/merge this without tests to allow concurrent work between us. 

NOTE: Adding tests for this on kernel level requires generalising the NewPayment Tests to EO wallets. These tests are also used by Tx and TxMeta tests, which also need to be generalised to EO wallets.

### DIFF
--- a/src/Cardano/Wallet/Kernel/Pending.hs
+++ b/src/Cardano/Wallet/Kernel/Pending.hs
@@ -10,7 +10,6 @@ module Cardano.Wallet.Kernel.Pending (
 import           Universum hiding (State)
 
 import qualified Data.List.NonEmpty as NE
-import qualified Data.Map.Strict as Map
 
 import           Control.Concurrent.MVar (modifyMVar_)
 
@@ -18,20 +17,21 @@ import           Data.Acid.Advanced (update')
 
 import           Pos.Chain.Txp (Tx (..), TxAux (..), TxOut (..))
 import           Pos.Core (Coin (..))
-import           Pos.Crypto (EncryptedSecretKey)
 
-import           Cardano.Wallet.Kernel.DB.AcidState (CancelPending (..),
-                     NewForeign (..), NewForeignError (..), NewPending (..),
-                     NewPendingError (..))
-import           Cardano.Wallet.Kernel.DB.HdRootId (HdRootId)
+import           Cardano.Wallet.Kernel.DB.AcidState (CancelPending (..), DB,
+                     IsNewTxError (..), NewForeign (..), NewForeignError (..),
+                     NewPending (..), NewPendingError (..))
 import           Cardano.Wallet.Kernel.DB.HdWallet
 import           Cardano.Wallet.Kernel.DB.InDb
+import           Cardano.Wallet.Kernel.DB.Read (lookupHdAccountId)
 import qualified Cardano.Wallet.Kernel.DB.Spec.Pending as Pending
 import           Cardano.Wallet.Kernel.DB.TxMeta (TxMeta, putTxMeta)
 import           Cardano.Wallet.Kernel.Internal
-import           Cardano.Wallet.Kernel.Read (getFOWallets, getWalletSnapshot)
+import           Cardano.Wallet.Kernel.Read (getEosPools, getFOWallets,
+                     getWalletSnapshot, mkEosAddress)
 import           Cardano.Wallet.Kernel.Submission (Cancelled, addPending)
 import           Cardano.Wallet.Kernel.Util.Core
+import           Cardano.Wallet.WalletLayer.Kernel.Conv (exceptT, withExceptT)
 
 {-------------------------------------------------------------------------------
   Submit pending transactions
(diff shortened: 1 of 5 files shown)


src/Cardano/Wallet/Kernel/Read.hs
```diff
@@ -195,6 +196,13 @@ addressPoolGapByRootId rootId db
             Left $ GetEosWalletErrorWrongAccounts (sformat build rootId)
         Just res -> snd <$> res
 
+mkEosAddress
+    :: ProtocolMagic
+    -> PublicKey
+    -> Address
+mkEosAddress pm
```
uroboros, line 203: @akegalj this is an oddity that came in with ApplyBlock for EO wallets - what do you think I should do here? Is this ok?

akegalj: So we have moved this function from Cardano.Wallet.Kernel.Wallets to Cardano.Wallet.Kernel.Read . My personal opinion about these changes is not that strong - but `mkEosAddress` looks like a helper function that can reside in both (even in its own Kernel.Address or Kernel.Util module)

I believe it is ok for now

akegalj (commented):
I have tried to go through `newTx` details and didn't find anything suspicious - but will give it another go when I am more awake (as I am new to these bits)

denisshevchenko:
LGTM, however I'm not deeply in context..

denisshevchenko (approved):

src/Cardano/Wallet/Kernel/Read.hs
KtorZ: IMO, this function shouldn't exist in a first place :confused:  
It conveys a wrong idea that there's such thing as an "EOSAddress" whereas, the address scheme has nothing to do with the fact that the wallet is externally owned. Furthermore, we don't even manipulate the underlying scheme at this level, so `Address` are just plain Cardano addresses.  

The helper is really just a composition of two functions and the extra indirection is superfluous.

src/Cardano/Wallet/Kernel/Pending.hs
```diff
@@ -73,54 +73,87 @@ newForeign w accountId tx meta = do
...
+            -> HdAccountBase
+            -> m (Either e ([HdAddress], [Coin]))
+        prefilterOurs db HdAccountBaseFO{} = do
+            foRoots <- liftIO $ getFOWallets walletPassive db
+            let ours = allOurs foRoots
+            return . Right $ (fst <$> ours, coinForThisRoot ours)
+        prefilterOurs db HdAccountBaseEO{} = do
+            eoAccounts <- liftIO $ getEosPools db (mkEosAddress $ walletPassive ^. walletProtocolMagic)
+            let ours = allOurs eoAccounts
+            return . Right $ (fst <$> ours, coinForThisRoot ours)
```
KtorZ, line 156: This is really about getting the state for `IsOurs`. Only the first statement is different, and it'd be nice if the function would reflect on that as well.

src/Cardano/Wallet/Kernel/Pending.hs
```diff
@@ -73,54 +73,87 @@ newForeign w accountId tx meta = do
...
-            return (Right txMeta)
+newTx ActiveWallet{..} accountId tx partialMeta upd = runExceptT $ do
+    db <- liftIO $ getWalletSnapshot walletPassive
+
+    account <- withExceptT newTxUnknownAccountErr $ exceptT $
+                lookupHdAccountId db accountId
+
+    (allAddrs, coinsForRoot) <- ExceptT $ liftIO $
+                                  prefilterOurs db (account ^. hdAccountBase)
+    _ <- liftIO $ upd allAddrs
```
KtorZ, line 103: Why are we now discarding the error potentially coming from `upd` :thinking: ?

src/Cardano/Wallet/Kernel/Pending.hs
```diff
@@ -73,54 +73,87 @@ newForeign w accountId tx meta = do
...
+-- 2) Create and persist transaction metadata, which depends on the transaction
+-- output coins in this wallet (the root is determined by the given account).
 --
--- NOTE: we select "our" output addresses from the transaction and pass it along to the data layer
-newTx :: forall e. ActiveWallet
+-- 3) Finally, the transaction is submitted to the submission layer.
+--
+-- Will fail if the HdAccountId does not exist or if some inputs of the
+-- new transaction are not available for spending.
+newTx :: IsNewTxError e
```
KtorZ, line 88: This feels wrong, and I now understand why the error below is now being discarded. `e` was here unconstrained to signify that errors could only occurs through the update function, which types is left open for the caller.

Instead, the burden of fetching the state for `IsOurs` should be left to the caller; so instead here, having `IsOurs s => ActiveWallet -> (HdAccountId, s)` would make more sense and remove quite some logic from this function which in the end, doesn't care much if the wallet is external or fully owned; or about the derivation scheme that is being used (hence the abstraction).

Having said that, I know realize something... Why are we even touching this function? We can't create transaction for externally owned wallet through this path; transactions have to be created through the external transactions endpoints. We'll most likely need some rework here when we support the sequential scheme for fully owned wallet but that's completely out-of-scope for E.O. wallets.

KtorZ (changes requested):
A few things that bother me with this PR, especially the fact that, this is out-of-scope for E.O. wallets; we shouldn't need to change any of this at this stage.

src/Cardano/Wallet/Kernel/Read.hs
uroboros: 👍

src/Cardano/Wallet/Kernel/Pending.hs
uroboros: 👍

src/Cardano/Wallet/Kernel/Pending.hs
uroboros: yes, I should return that case statement

src/Cardano/Wallet/Kernel/Pending.hs
uroboros: @KtorZ yes this is required for EO transactions - keep in mind that it does tx submission and updates tx metadata, etc. You'll see that the EO code interacts with this code when `submitSignedTx` hands over to `newPending`. Hence these changes here, which say that based on the account type to which we are adding a tx, we apply the appropriate prefiltering.

PR #344, opened 2019-02-15
WALLET_DETAILS integration tests

piotr-iohk:
<p align="right">#</p>

# Overview

- [x] I have added tests for WALLET_DETAILS capability (https://docs.google.com/spreadsheets/d/1ztF2PnnrMnkAxrhdq3LUbP1KazxcxuMrbiObTs1CuTk)

# Comments

### DIFF
--- a/test/integration/Test/Integration/Framework/DSL.hs
+++ b/test/integration/Test/Integration/Framework/DSL.hs
@@ -86,7 +86,6 @@ module Test.Integration.Framework.DSL
     , assuranceLevel
     , backupPhrase
     , createdAt
-    , eosCreatedAt
     , externallyOwnedAccounts
     , failures
     , fromWalletId
@@ -122,6 +121,7 @@ import           Data.Default (Default (..))
 import qualified Data.Foldable as F
 import           Data.Generics.Internal.VL.Lens (lens)
 import           Data.Generics.Product.Fields (field)
+import           Data.Generics.Product.Positions (HasPosition, position)
 import           Data.Generics.Product.Typed (HasType, typed)
 import           Data.List ((!!))
 import qualified Data.List.NonEmpty as NonEmpty
@@ -393,11 +393,8 @@ assuranceLevel = typed
 backupPhrase :: HasType BackupPhrase s => Lens' s BackupPhrase
 backupPhrase = typed
 
-createdAt :: Lens' Wallet WalletTimestamp
-createdAt = field @"walCreatedAt"
-
-eosCreatedAt :: Lens' EosWallet WalletTimestamp
-eosCreatedAt = field @"eoswalCreatedAt"
+createdAt :: HasPosition 6 s s WalletTimestamp WalletTimestamp => Lens' s WalletTimestamp
+createdAt = position @6
 
 externallyOwnedAccounts :: HasType [AccountPublicKeyWithIx] s => Lens' s [AccountPublicKeyWithIx]
 externallyOwnedAccounts = typed
--- a/test/integration/Test/Integration/Framework/DSL.hs
+++ b/test/integration/Test/Integration/Framework/DSL.hs
@@ -86,6 +86,7 @@ module Test.Integration.Framework.DSL
     , assuranceLevel
     , backupPhrase
     , createdAt
+    , eosCreatedAt
     , externallyOwnedAccounts
     , failures
     , fromWalletId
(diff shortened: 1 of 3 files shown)


test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -26,6 +30,77 @@ spec = do
...
+            [ expectTxStatusEventually [InNewestBlocks, Persisted]
+            ]
+        getSourceAfter <- request $ Client.getWallet $- fixtureSource ^. wallet . walletId
+        verify getSourceAfter
+            [ expectFieldEqual amount (10000000 - 10 - fee)
+            ]
+        -- getDestAfter <- request $ Client.getWallet $- fixtureDest ^. wallet . walletId
+        -- verify getDestAfter
+        --     [ expectFieldEqual amount 10
+        --     ]
```
piotr-iohk: It seems I cannot get the desired amount on the destination wallet, even if on the `Client.postTransaction` request before I am waiting for the transaction status to be `expectTxStatusEventually [Persisted]`... Is there any other way I can make sure the amount from the transaction reached the destination? In particular I'd like to use endpoint `Client.getWallet` to verify the amount is there.

test/integration/Test/Integration/Framework/DSL.hs
```diff
@@ -86,6 +86,7 @@ module Test.Integration.Framework.DSL
     , assuranceLevel
     , backupPhrase
     , createdAt
+    , eosCreatedAt
```
KtorZ: This defy a bit the purpose of using lenses here :sweat_smile:  ... I believe we can define a `Lens s WalletTimestamp` quite easily, where `s` can be either a `Wallet` or an `EosWallet`

KtorZ: Note, this probably conflicts with the `spendingPasswordLastUpdate` because both timestamps have a similar type. Alright, we'll need to give it some thoughts.

test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -26,6 +30,77 @@ spec = do
...
+            (defaultDistribution 10 fixtureDest)
+            defaultGroupingPolicy
+            (Just $ defaultSpendingPassword)
+        verify respPayment
+            [ expectTxStatusEventually [InNewestBlocks, Persisted]
+            ]
+        getSourceAfter <- request $ Client.getWallet $- fixtureSource ^. wallet . walletId
+        verify getSourceAfter
+            [ expectFieldEqual amount (10000000 - 10 - fee)
+            ]
```
KtorZ, line 102: Fee are only constant here because we don't have any activity on the local cluster beside this very scenario. In practice, this isn't guaranteed at all, and that's why it's called "estimate fee". They may likely diverge from the actual ones especially on wallets with many UTxOs where inputs are selected randomly. You may end up with slightly different fees in practice.

Ideally, we would verify that the balance is less than before by at least, the amount of the transaction + the minimal fee. But that's pretty much all.

test/integration/Test/Integration/Scenario/Wallets.hs
KtorZ: So. The addresses present in the fixture aren't addresses of the underlying wallet. They are just a bunch of random addresses one can use to make payment too, but they aren't tracked by the wallet described in the fixture at all. Hence you not getting the wallet to be updated with the amount; funds are actually sent to... another random wallet.

If you want to get send to a particular address of that fixture wallet, use `customDistribution` and provide an account for that destination wallet instead :+1:

KtorZ (changes requested):

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: >Fee are only constant here because we don't have any activity on the local cluster beside this very scenario.

Given the circumstances you mention ^, do we want to change the scenario to what you propose or leave it as is?

If we'd like to change it to:
>we would verify that the balance is less than before by at least, the amount of the transaction + the minimal fee. But that's pretty much all.

then how do we calculate the minimal fee?

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: ah, ok :+1:

test/integration/Test/Integration/Framework/DSL.hs
piotr-iohk: If we are after having a lens with the name `createdAt` that works for both `Wallet` and `EosWallet` this is something I came up with:

```haskell
createdAt :: HasPosition 6 s s WalletTimestamp WalletTimestamp => Lens' s WalletTimestamp
createdAt = position @6
```
as it turns out that `walCreatedAt` and `eoswalCreatedAt` are both on the 6th position in their respective datatypes.

We could have that istead of:

```haskell
createdAt :: Lens' Wallet WalletTimestamp
createdAt = field @"walCreatedAt"

eosCreatedAt :: Lens' EosWallet WalletTimestamp
eosCreatedAt = field @"eoswalCreatedAt" 
```

I imagine that this is probably not the nicest approach but... well, do let know :)

piotr-iohk:
@KtorZ I have pushed the changes requested. I had two questions about:
 - `createdAt`_implementation_ using `HasPosition` (pushed this)
 - calculating minimal fee if we want to verify balance of the wallet after outgoing transaction differently

However If the tests are good as they are i think it might be merged...

KtorZ (approved):

Issue #330, opened 2019-02-09
Legacy-Free Networking Layer Prototype


KtorZ:
# Context

As part of #10 and #88 , we've done quite some research on _how_ to deal with the current haskell core node to retrieve blocks via the existing diffusion layer. We'd like however to be completely free of the current Haskell implementation in a very near future. This means being able to receive block headers and blocks from the network, without relying on any of the existing implementation nor, shared memory access. 

# Decision

Let's come up with a small (throw-away) prototype that will demonstrate _how_ we can achieve connecting to an existing trusted node to receive block headers and blocks. Doing so, we do not want to re-use any of the existing implementation, but rather show what's the minimal amount of work needed to _rebuild_ this particular aspect of the current networking layer.

There's some inspiration to get from the current BListener, WalletMode and DiffusionLayer, though we are aiming at something extremely simpler. As a mere consumer connecting to a single peer, our needs from the network are also much smaller than normal nodes. 

The goal of this task is to implement a simple prototype which connects to a single node, a print block headers broadcast by the node to the console (print their content in a human-readable way; so after having decoded the CBOR serialized bytes).

# Acceptance Criterias

- [ ] The prototype _must_ not re-use code from an existing implementation (neither `cardano-sl`, nor `cardano-wallet`)
- [ ] The prototype _must_ be simple, readable and commented with insights 
- [ ] The prototype _must_ receive block headers from a trusted node and print them to the console
- [ ] The prototype _must_ decode retrieved data from the node before printing them out
- [ ] The prototype _must_ not handle rollbacks 
- [ ] The prototype _may_ also retrieve corresponding blocks from the node and print them to the console
- [ ] The trusted node _may_ be part of a local cluster started with the cardano-sl's cluster tool
- [ ] This _shouldn't_ take more than a week of time. The ticket will be aborted should it take more.

---

# Development Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #347        | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| ?        | -        |

---

# Retrospective

PR #347, opened 2019-02-18
Minimal separate process downloading blocks from Byron node

paweljakubas:
<p align="right">#330</p>

# Overview

- [x] I made the prototype to be separate process that uses minimum network node primitives. No diffusion layer used. Using tcp would be more error prone and substatially less simple
- [x] The prototype is made to receive block headers from a trusted node and downloads corresponding block. The block is printed to console
- [x] The prototype talks with local cluster (started by integration tests)
- [x] All important steps are commented in the code. Byron specifics are located in separated file

# Comments

### DIFF
--- a/cardano-wallet.cabal
+++ b/cardano-wallet.cabal
@@ -622,3 +622,43 @@ test-suite integration
       Test.Integration.Scenario.Transactions
       Test.Integration.Scenario.Wallets
       Test.Integration.Scenario.Node
+
+
+executable decoupled-wallet
+  default-language:    Haskell2010
+
+  default-extensions:  DeriveGeneric
+                       DeriveDataTypeable
+                       GeneralizedNewtypeDeriving
+                       OverloadedStrings
+                       MonadFailDesugaring
+                       TemplateHaskell
+                       LambdaCase
+                       TypeApplications
+                       OverloadedStrings
+                       NoImplicitPrelude
+
+  ghc-options:         -threaded -Wall
+
+  build-depends:       base
+                     , async
+                     , binary
+                     , bytestring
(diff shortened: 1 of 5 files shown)


cardano-wallet.cabal
```diff
@@ -622,3 +622,43 @@ test-suite integration
...
+  build-depends:       base
+                     , async
+                     , binary
+                     , bytestring
+                     , cardano-sl-binary
+                     , cardano-sl-chain
+                     , cardano-sl-core
+                     , cardano-sl-infra
+                     , cardano-sl-networking
+                     , cardano-sl-util
```
KtorZ, line 652: I am having a look at the PR now and will push some refactorings changes on the branch. 
Note that, you've been cheating my friend :stuck_out_tongue: 

The requirements said:

```
* The prototype _must_ not re-use code from an existing implementation (neither `cardano-sl`, nor `cardano-wallet`)
```

And... I see quite a lot of `cardano-sl-***` dependencies here :upside_down_face: ...

paweljakubas: yes, I used **cardano-sl-networking** node and converse primitives. They use quite sophisticated way of communication there...Other dependencies are just infra (BiP, bipPacking, PeerData, VerInfo - required to talk with the node), of course chain (Block, BlockHeader, etc for definitions), core (for NewestFirst) .

So I faced cardano-dependences vs simplicity trade-off. For me not using the abovementioned would mean very complicated, not easy to understand and error-prone code. cardano-sl-networking use not so easy, I would say cryptic, bidirectional communication protocol...

Nevertheless, I agree - in this version there are still dependencies. But it is simple on the other way

PR #349, opened 2019-02-19
Revert recent NodeStateAdaptor changes (#87)

Anviking:
<p align="right">#</p>

# Overview

- [x] ~Reverted~Reset [`NodeStateAdaptor.hs`](https://github.com/input-output-hk/cardano-wallet/commits/develop/src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs) to 1da9ca509d571fd18ef6e103795e7161dd697f26
- [x] Fixed errors (where we create it) until it compiled
- [x] Re-Added `retrying` for node api connection check

# Comments
This is not a complete revert of #87. There were several good things in there, like CLI changes, actually starting the node server, and cleanup.

- [ ] I should write a ticket if we're doing this

### DIFF
--- a/src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
+++ b/src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
@@ -4,9 +4,6 @@
 {-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE RankNTypes                 #-}
 
-{-# OPTIONS_GHC -fno-warn-orphans #-}
--- For SecurityParameter from the Node API
-
 module Cardano.Wallet.Kernel.NodeStateAdaptor (
     WithNodeState     -- opaque
   , NodeStateAdaptor  -- opaque
@@ -27,6 +24,9 @@ module Cardano.Wallet.Kernel.NodeStateAdaptor (
   , getFeePolicy
   , getSlotCount
   , getCoreConfig
+  , getSlotStart
+  , getNextEpochSlotDuration
+  , getNodeSyncProgress
   , curSoftwareVersion
   , compileInfo
   , getNtpDrift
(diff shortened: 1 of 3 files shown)


Anviking:
(I’ll sanity check this tomorrow)

src/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -102,7 +104,7 @@ apiServer
     env <- ask
     let diffusion' = Kernel.fromDiffusion (lower env) diffusion
     logInfo "Testing node client connection"
-    eresp <- liftIO . NodeStateAdaptor.retrying . runExceptT $ NodeClient.getNodeSettings nodeClient
+    eresp <- liftIO . retrying . runExceptT $ NodeClient.getNodeSettings nodeClient
```
KtorZ, line 107: What's the need for this now :thinking: ?

Anviking: The [`node-settings`](https://github.com/input-output-hk/cardano-wallet/blob/171f15e4f4de0146f6d1acca70e3b5337cb396d0/src/Cardano/Wallet/API/V1/Handlers/Settings.hs#L18-L28) and `node-info` endpoints where we redirect to the node

src/Cardano/Wallet/Kernel/NodeStateAdaptor.hs
```diff
@@ -259,6 +257,27 @@ data NodeStateAdaptor m = Adaptor {
...
+      -- in this context we need something very simple: a tuple containing the
+      -- "global blockchain height" and the "node blockchain height". The
+      -- former is the maximum between the biggest height we observed from an
+      -- unsolicited block we received and the current local tip:
+      --
+      -- global_height = max (last_known_header, node_local_tip)
+      --
+      -- The latter is simply the node local tip, i.e. "how far we went into
+      -- chasing the global blockchain height during syncing".
+    , getNodeSyncProgress :: LockContext -> m (Maybe BlockCount, BlockCount)
```
Anviking, line 279: We could have `getNodeSyncProgress` remain removed, I guess, but ¯\_(ツ)_/¯

src/Cardano/Wallet/Server/Plugins.hs
KtorZ: I thought that was the goal of this PR to remove these calls :thinking: ?

Anviking: Hm… I was more concerned about the internal uses of NodeStateAdaptor fields. Now these use `withNodeState` again.

KtorZ: Okay. Fine by me.

KtorZ (approved):

Issue #315, opened 2019-02-06
Improve logging


denisshevchenko:
# Context

Our logging situation is terrible and not helpful. A lot of `Buildable`-instances are not as useful as we'd want. So we will need to fix these instances for pretty much every type we use to make sure that we get the most information we can out of our logging.That also probably mean getting rid of this `BuildableSafe` we did a while ago and just, not output password and private keys in the `Buildable`-instance at all.

# Decision

1. Review all `Buildable`-instances and check what outputs we get from the logs.
2. (Probably) get rid of `BuildableSafe`-instances by removing them with `Buildable`-instances.
3. Add a few actual log lines in the wallet code, apart from the API calls.

# Acceptance Criterias

- [ ] All `Buildable`-instances _must_ provide all information we need in logs.
- [ ] Information which is printed in logs _must_ be well-formatted.

---

# Development Plan

Proposal - open a separate PR for each layer:

1. `Buildable`-instances in `API`-layer.
2. `Buildable`-instances in `Wallet`-layer.
3. `Buildable`-instances in `Kernel`-layer.
4. `Buildable`-instances in `DB`-layer.

- [ ] Make sure all `Buildable`-instances in `cardano-wallet` produce useful information in logs.
  - [ ] Re-write existing `Buildable`-instances using utils from `Cardano.Wallet.Util` module.
- [ ] Replace `BuildableSafeGen`-instances by corresponding `Buildable`-instances and make sure we don't print out passwords and private keys. `BuildableSafeGen`-instances must be removed from these modules:
  - [ ] `Cardano.Wallet.API.Request`
  - [ ] `Cardano.Wallet.API.Request.Filter`
  - [ ] `Cardano.Wallet.API.Request.Sort`
  - [ ] `Cardano.Wallet.API.V1.Types`
  - [ ] `Cardano.Wallet.Types.UtxoStatistics`
- [ ] Add additional log messages we need.

# PR

| Number   | Base      |
| ---      | ---       |
| #316       | `develop` |
| #345       | `develop` |
| #348       | `develop` |
| #352       | `develop` |
| #358       | `develop` |
| #361       | `develop` |
| #363       | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| ?        | -        |

---

# Retrospective

akegalj:
[...]

PR #352, opened 2019-02-20
API V1 types: EOS-wallet related types logging improving.

denisshevchenko:
<p align="right">#315</p>

# Overview

- [x] I have improved logging for EOS-wallets related types.

# Comments

### DIFF
--- a/src/Cardano/Wallet/API/V1/Types.hs
+++ b/src/Cardano/Wallet/API/V1/Types.hs
@@ -179,7 +179,8 @@ import           Cardano.Wallet.API.V1.Generic (jsendErrorGenericParseJSON,
 import           Cardano.Wallet.API.V1.Swagger.Example (Example, example)
 import           Cardano.Wallet.Kernel.AddressPoolGap (AddressPoolGap)
 import           Cardano.Wallet.Types.UtxoStatistics
-import           Cardano.Wallet.Util (mkJsonKey, showApiUtcTime)
+import           Cardano.Wallet.Util (buildIndent, buildList, mkJsonKey,
+                     showApiUtcTime)
 
 import           Cardano.Mnemonic (Mnemonic)
 import qualified Pos.Binary.Class as Bi
@@ -283,7 +284,8 @@ newtype WalletCoin = WalletCoin { unWalletCoin :: Core.Coin }
 
 deriveSafeBuildable ''WalletCoin
 instance BuildableSafeGen WalletCoin where
-    buildSafeGen _ (WalletCoin c) = bprint build c
+    buildSafeGen sl (WalletCoin c) =
+        bprint (plainOrSecureF sl build (fconst "<hidden amount>")) c
 
 instance ToJSON WalletCoin where
     toJSON (WalletCoin c) = toJSON . Core.unsafeGetCoin $ c
@@ -615,15 +617,15 @@ instance Arbitrary UpdateEosWallet where
 
 deriveSafeBuildable ''UpdateEosWallet
 instance BuildableSafeGen UpdateEosWallet where
-    buildSafeGen sl UpdateEosWallet{..} = bprint ("{"
-        %" assuranceLevel="%buildSafe sl
-        %" name="%buildSafe sl
-        %" addressPoolGap="%build
-        %" }")
-        ueowalAssuranceLevel
+    buildSafeGen _ UpdateEosWallet{..} = bprint
+        ( "Update for EOS wallet"
+        % "\n  name: " % build
+        % "\n  " % build
+        % "\n  " % build
+        )
         ueowalName
+        ueowalAssuranceLevel
         ueowalAddressPoolGap
-
 
 -- | A type modelling the update of an existing wallet.
 data WalletUpdate = WalletUpdate {
@@ -886,11 +888,11 @@ instance Arbitrary Wallet where
 
 deriveSafeBuildable ''Wallet
 instance BuildableSafeGen Wallet where
-  buildSafeGen _ Wallet{..} = bprint
+  buildSafeGen sl Wallet{..} = bprint
     ( "Wallet"
     % "\n  " % build
     % "\n  name: " % build
-    % "\n  balance: " % build
+    % "\n  balance: " % buildSafe sl
     % "\n  created at: " % build
     % "\n  " % build
     % "\n  has spending password: " % build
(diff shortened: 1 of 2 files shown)


denisshevchenko:
Examples.

`EosWallet`:

```
EOS (externally-owned sequential) wallet
  wallet id: J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg
  name: My EOS-wallet
  balance: 1104330362855056 coin(s)
  address pool gap: 84
  assurance level: strict
  created at: 1191686829598181
```

List of `EosWallet`s:

```
- EOS (externally-owned sequential) wallet
    wallet id: J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg
    name: My EOS-wallet
    balance: 1104330362855056 coin(s)
    address pool gap: 84
    assurance level: strict
    created at: 1191686829598181
- EOS (externally-owned sequential) wallet
    wallet id: J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg
    name: My EOS-wallet
    balance: 36520912267882166 coin(s)
    address pool gap: 71
    assurance level: normal
    created at: 433876397698368
```

---

`NewEosWallet`:

```
Request for new EOS wallet
  name: My EOS-wallet
  address pool gap is undefined (default value 20 will be used)
  assurance level: normal
  accounts:
  - account public key: pub:b7732ec4, account ix: 3965755267
  - account public key: pub:1762b3cb, account ix: 3195482834
  - account public key: pub:b30d4f07, account ix: 3964313652
  - account public key: pub:30329f70, account ix: 2289830138
  - account public key: pub:b951bb70, account ix: 3797983371
  - account public key: pub:7cd01c19, account ix: 4212473129
  - account public key: pub:f5e60727, account ix: 3249079988
```

or:

```
Request for new EOS wallet
  name: My EOS-wallet
  address pool gap: 52
  assurance level: strict
  accounts:
  - account public key: pub:055c7ccb, account ix: 3192540224
  - account public key: pub:2623aacf, account ix: 3257216011
  - account public key: pub:eb787acf, account ix: 3476420452
```

---

`UpdateEosWallet`:

```
Update for EOS wallet
  name: My EosWallet
  assurance level: normal
  address pool gap: 44
```

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -943,19 +944,24 @@ instance Arbitrary EosWallet where
...
-  buildSafeGen sl EosWallet{..} = bprint ("{"
-    %" id="%buildSafe sl
-    %" name="%buildSafe sl
-    %" addressPoolGap="%build
-    %" balance="%buildSafe sl
-    %" }")
+  buildSafeGen _ EosWallet{..} = bprint
+    ( "EOS (externally-owned sequential) wallet"
+    % "\n  " % build
+    % "\n  name: " % build
```
akegalj, line 951: don't we leak info this way? I guess this would be good for development purpose but we don't want to leak id, name, balance in production logs.

I guess that's the whole idea behind `buildSafe` - to give us ability to guard and hide those details in production setting

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1217,16 +1220,17 @@ instance ToSchema NewEosWallet where
...
-        %" assuranceLevel="%buildSafe sl
-        %" name="%buildSafe sl
-        %" }")
-        neweoswalAccounts
+    buildSafeGen _ NewEosWallet{..} = bprint
+        ( "Request for new EOS wallet"
+        % "\n  name: " % build
+        % "\n  " % build
+        % "\n  " % build
+        % "\n  accounts:\n" % buildIndent 2 (buildList build)
```
akegalj, line 1229: wallet name should be safe and not printed out in production - this leaks info

src/Cardano/Wallet/Kernel/AddressPoolGap.hs
```diff
@@ -60,7 +60,15 @@ instance Arbitrary AddressPoolGap where
 
 instance Buildable AddressPoolGap where
     build (AddressPoolGap gap) =
-        bprint ("Address pool gap "%int) gap
+        bprint ("address pool gap: "%int) gap
+
+-- | There's default instance for 'Buildable (Maybe a)', but it produces
+-- just empty string for 'Nothing', it's not very informative.
+instance {-# OVERLAPPING #-} Buildable (Maybe AddressPoolGap) where
```
akegalj, line 67: We usually don't like overlapping instances but I think this is ok use case

akegalj (changes requested):
There are some privacy issues I believe with this PR as we are dropping safe/private logging

src/Cardano/Wallet/API/V1/Types.hs
denisshevchenko: > but we don't want to leak id, name, balance in production logs.

What's the problem with wallet id and wallet name in production logs??

src/Cardano/Wallet/API/V1/Types.hs
denisshevchenko: Why? It's just an arbitrary human-readable string, for example "My favorite cat is Jack". :-)

src/Cardano/Wallet/API/V1/Types.hs
denisshevchenko: @akegalj Ok, now balance is hidden via `buildSafe`.

src/Cardano/Wallet/API/V1/Types.hs
KtorZ: Agreed with @denisshevchenko on that. The name is just a metadata and is actually truly helpful to have in logs since that's what user use when they refer to their wallets.

I agree about hiding the balance though. That's is not something that is easily available from the blockchain since, in order to keep track of a wallet balance, one would require the private key (or at least, the public key). There's also a very minor "privacy" issue with the showing the complete account key. Note however that, those concerns are rather delusional since the database is anyway freely available unencrypted. So, all those pieces of information are already there on disk, and can be read at any time by any party having access to the files. 

Let's see what Charles Morgan and David think about this. I've already started a conversation with them on topic in order to know where to draw the line of privacy.

denisshevchenko: > I agree about hiding the balance though.

It is hidden already, via `buildSafe`.

akegalj: > Why? It's just an arbitrary human-readable string, for example "My favorite cat is Jack".

heh - it might be "stash for drugs" as well. If we are going to show this I believe we should have a note somewhere in our privacy letter that user is ok about this and that it will be listed in logs - so that user is aware of it and doesn't sue us or smth

src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -886,11 +888,11 @@ instance Arbitrary Wallet where
 
 deriveSafeBuildable ''Wallet
 instance BuildableSafeGen Wallet where
-  buildSafeGen _ Wallet{..} = bprint
+  buildSafeGen sl Wallet{..} = bprint
     ( "Wallet"
     % "\n  " % build
     % "\n  name: " % build
-    % "\n  balance: " % build
+    % "\n  balance: " % buildSafe sl
```
akegalj, line 895: although it turns out in the last discussion we had with guys that it is ok to show balance as well as daedalus is dumping wallet balances to the logs

src/Cardano/Wallet/API/V1/Types.hs
akegalj: We concluded in the slack discussion that it is ok to show name in logs as daedalus is already doing this and we are not pulling logs automatically - but user has to manually upload them to us.

akegalj (approved):
After discussion with rest of the team conclusion is that it is ok to log these details as Daedalus is already logging name/balance . Even then, we don't upload any logs automatically but user has to do it manually

PR #356, opened 2019-02-21
Wallets update delete utxo integration tests

piotr-iohk:
<p align="right">#</p>

# Overview

- [x] I have added integration tests for remaining wallet capabilities `WALLETS_UPDATE`, `WALLETS_DELETE`, `WALLETS_UTXO`
- [x] I have updated `manual/scripts/wallet.sh` adding ability to update and get utxo of the wallet using scripts (while testing it on the manually started node)

# Comments

### DIFF
--- a/test/integration/Test/Integration/Scenario/Wallets.hs
+++ b/test/integration/Test/Integration/Scenario/Wallets.hs
@@ -19,8 +19,6 @@ spec :: Scenarios Context
 spec = do
     scenario "WALLETS_DELETE_01 - deleted wallet is not available" $ do
         fixture <- setup $ defaultSetup
-            & walletName .~ "漢ę ó ł ąś ł żźćń字"
-            & mnemonicWords .~ testBackupPhrase
 
         successfulRequest $ Client.deleteWallet
             $- (fixture ^. wallet . walletId)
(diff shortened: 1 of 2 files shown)


test/integration/Test/Integration/Scenario/Wallets.hs
```diff
@@ -925,19 +1055,17 @@ spec = do
...
         verify response
             [ expectWalletUTxO [14, 42, 1337]
             ]
 
-
-    scenario "WALLETS_UTXO_05 - UTxO statistics reflect wallet's activity" $ do
+    scenario "WALLETS_UTXO_04 - UTxO statistics reflect wallet's activity" $ do
         fixture <- setup $ defaultSetup
             & initialCoins .~ [13, 43, 66, 101, 1339]
         response <- request $ Client.getUtxoStatistics
```
piotr-iohk, line 1071: If anyone has any idea for more utxo scenarios that could be added, please let me know :+1:

paweljakubas: maybe it would be nice to add a test for utxo that checks the real output from this endpoint. So for **initialCoins .~ [13, 43, 66, 101, 1339]** the user is really expecting : 
```
"data": {
"boundType": "Log10",
"allStakes": 1562,
"histogram": {
"10": 0,
"100": 3,
"1000": 1,
"10000": 1,
"100000": 0,
"1000000": 0,
"10000000": 0,
"100000000": 0,
"1000000000": 0,
"10000000000000000": 0,
"100000000000000": 0,
"1000000000000": 0,
"10000000000": 0,
"100000000000": 0,
"45000000000000000": 0,
"10000000000000": 0,
"1000000000000000": 0
}
}
```

paweljakubas (approved):

test/integration/Test/Integration/Scenario/Wallets.hs
piotr-iohk: good point! I think currently our framework does not allow for this (I mean asserting for the raw response), but it would be definitely nice thing to have.

Issue #357, opened 2019-02-21
Move EOS-related stuff to HdRoot


denisshevchenko:
# Context

Currently, EOS-related stuff (accounts public keys and address pool gap) are storing in `HdAccount`s (to be more precise, in `HdAccountBase` which is a part of `HdAccount`). But this solution comes with a few problems:

- Trying to keep consistency with the address pool gaps between accounts.
- Having to review most of our structure to work with `HdAccountBase`, changing in the end a lot of code that is very little concerned by the E.O. wallets.
- Having extra checks to make sure that when accessing accounts of a wallets, we only deal with the correct sort of account (i.e. with correct branch of `HdAccountBase`).

So, we want to modify the `HdRoot` to hold a list of account public keys in a keystore fashion, as well as the address pool gap. In the end, this allows us to have very similar implementation between E.O. and F.O. wallet where they both go looking for public keys in a "keystore" from a list of keys.

# Decision

Make a list of account public keys and address pool gap a part of `HdRoot`.

# Acceptance Criterias

1. List of account public keys _must_ be a part of `HdRoot`.
2. Address pool gap _must_ be a part of `HdRoot`.

---

# Development Plan

- [ ] Move `AddressPoolGap` and `[Core.PublicKey]` (probably `NonEmpty` instead of `[]`) to `HdRoot`.
- [ ] Return `HdAccountId` to `HdAccount`.
- [ ] Remove `HdAccountBase`.
- [ ] Change all corresponding lenses, in DB-layer.
- [ ] Change (and probably remove) all related functions. For example, update `AddressPoolGap` value not in corresponding accounts, but in `HdRoot` itself.

# PR

| Number   | Base      |
| ---      | ---       |
| #?       | `develop` |

# QA 

| Criteria | Coverage |
| ---      | ---      |
| ?        | -        |

---

# Retrospective

PR #360, opened 2019-02-22
clean-up access to encrypted secret keys and address pools construction

KtorZ:
<p align="right">#357</p>

# Overview

- This is in preparation of some upcoming restructuring where we get rid of the HdAccountBase and replace it with an HdRootBase. This will ultimately remove a lot of errors linked to the possible
inconsistency we could get across all accounts of a root.

- This particular restructuring move some responsibility to the callers of getESK and getPools and
purify a bit those primitives functions. This way, we leave to the caller to decide what to do
when something goes wrong. In this particular case, we fail hard with an invariant violation

# Comments

### DIFF
--- a/src/Cardano/Wallet/Kernel/BListener.hs
+++ b/src/Cardano/Wallet/Kernel/BListener.hs
@@ -34,7 +34,8 @@ import           Pos.Crypto (EncryptedSecretKey)
 import           Pos.DB.Block (getBlund)
 import           Pos.Util.Log (Severity (..))
 
-import           Cardano.Wallet.Kernel.AddressPool (AddressPool)
+import           Cardano.Wallet.Kernel.AddressPool (AddressPool,
+                     ErrAddressPoolInvalid (..))
 import           Cardano.Wallet.Kernel.DB.AcidState (ApplyBlock (..),
                      ObservableRollbackUseInTestsOnly (..), SwitchToFork (..),
                      SwitchToForkInternalError (..))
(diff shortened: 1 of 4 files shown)


src/Cardano/Wallet/Kernel/Read.hs
```diff
@@ -60,96 +54,69 @@ getWalletSnapshot pw = query' (pw ^. wallets) Snapshot
...
+            HdAccountBaseFO{} ->
+                Nothing
+            (HdAccountBaseEO accId pubKey gap) ->
+                Just (gap, Map.singleton accId pubKey)
+
+    addressesByAccountId'
+        :: DB
+        -> HdAccountId
+        -> [(Address, Word32)]
+    addressesByAccountId' _db accId = addressesByAccountId db accId
```
denisshevchenko, line 102: If the first argument is not used - why do we need it? Maybe just remove `_db`?

src/Cardano/Wallet/Kernel/Read.hs
```diff
@@ -60,96 +54,69 @@ getWalletSnapshot pw = query' (pw ^. wallets) Snapshot
...
-
-    toAccountAddressPools' :: HdRoot -> [(HdAccountId, AddressPool Address)]
-    toAccountAddressPools' root
-        = toAccountAddressPools root (eosAccountsByRootId (root ^. hdRootId) db)
-
-    toAccountAddressPools
+    -- NOTE:
+    -- We arbitrarily take the maximum AddressPoolGap here. This is rather
+    -- incorrect but will go away with the next PR as part of #357!
+    externallyOwnedRoots
```
denisshevchenko, line 87: Maybe we have to unify our naming? I mean that in some places we use a full form `externallyOwned`, but in other places we use a short one, `EO`...

src/Cardano/Wallet/Kernel/BListener.hs
```diff
@@ -78,9 +79,20 @@ prefilterContext
...
-    hdFOs <- getFOWallets pw db
-    hdEOs <- getEosPools db (mkEosAddress $ pw ^. walletProtocolMagic)
+    hdFOs <- Map.traverseWithKey invariantFO =<< getEncryptedSecretKeys pw db
+    hdEOs <- Map.traverseWithKey invariantEO $ getAddressPools (pw ^. walletProtocolMagic) db
     return (foreigns, hdFOs, hdEOs)
+  where
+    invariantFO :: HdRootId -> Maybe a -> IO a
+    invariantFO rootId = flip maybe return $ fail $ toString $
+        "Cardano.Wallet.Kernel.BListener: invariant violation: encrypted secret key \
+        \hasn't been found for the given root id: " <> sformat build rootId
```
denisshevchenko, line 89: Maybe just match explicitly?

```
    invariantFO :: HdRootId -> Maybe a -> IO a
    invariantFO rootId (Just k) = return k
    invariantFO rootId Nothing = fail $ toString $
        "Cardano.Wallet.Kernel.BListener: invariant violation: encrypted secret key \
        \hasn't been found for the given root id: " <> sformat build rootId
```

And the same for `invariantEO`..

denisshevchenko (approved):
LGTM. I've asked some little questions though..

